Issue #31, opened 2023-06-05
As a Developer, I need to implement catch-up process as a Rust program, to be able to use it as a base for a Rust decenteralized snark worker


akoptelov:
Definition of Done: Rust program (redux-based) that performs catch-up process in a Mina network, preparing Rust ledger implementation.

akoptelov:
Long-range fork implemented in openmina SM.

There is a way to capture blocks participating in fork resolution. Tests should be added using them.

Using network split, we can have integration tests for fork resolution.

Issue #5, opened 2022-11-15
As a developer, I need to add consensus logic to the webnode


binier:
[implementation in mina-rs](https://github.com/ChainSafe/mina-rs/tree/main/consensus) should come in handy.

- [x] Implement best tip selection during short range fork.
- [x] Implement best tip selection during long range fork.

PR #34, opened 2023-06-06
Long range fork resolution

akoptelov:
I've added long range fork handling. Also I've introduced an intermediate state for fork range detection.

I'm going to collect block pairs for different forks (both long and short ranges) and add them along with tests.

Re: #5 
Fix: #31

### DIFF
--- a/snarker/src/consensus/consensus_reducer.rs
+++ b/snarker/src/consensus/consensus_reducer.rs
@@ -186,3 +186,79 @@ impl ConsensusState {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use mina_p2p_messages::v2::{MinaStateProtocolStateValueStableV2, StateHash};
+
+    use super::long_range_fork_take;
+
+    fn long_range_fork_test(
+        tip: &str,
+        cnd: &str,
+        tip_hash: &str,
+        cnd_hash: &str,
+        expect_take: bool,
+    ) {
+        let tip = serde_json::from_str::<MinaStateProtocolStateValueStableV2>(tip).unwrap();
+        let cnd = serde_json::from_str::<MinaStateProtocolStateValueStableV2>(cnd).unwrap();
+        let tip_hash = tip_hash.parse::<StateHash>().unwrap();
+        let cnd_hash = cnd_hash.parse::<StateHash>().unwrap();
+
+        let (take, _) = long_range_fork_take(
+            &tip.body.consensus_state,
+            &cnd.body.consensus_state,
+            &tip_hash,
+            &cnd_hash,
+        );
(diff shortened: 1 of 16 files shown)


akoptelov:
@binier I've migrated long forks to new version, please review.

snarker/src/consensus/consensus_reducer.rs
```diff
@@ -87,6 +117,40 @@ impl ConsensusState {
...
+                    long_range_fork_take(tip_cs, candidate_cs, &tip_hash, candidate_hash);
+
+                let Some(candidate_state) = self.blocks.get_mut(candidate_hash) else {
+                    return;
+                };
+                candidate_state.status = ConsensusBlockStatus::LongRangeForkResolve {
+                    time: meta.time(),
+                    compared_with: tip_hash,
+                    decision: if take {
+                        ConsensusLongRangeForkDecision::Take(why)
```
binier, line 146: For consistency, can we use `Use` and `Ignore` instead of `Take` and `Keep`?

snarker/src/consensus/consensus_state.rs
```diff
@@ -37,6 +37,29 @@ impl ConsensusShortRangeForkDecision {
...
+pub enum ConsensusLongRangeForkResolutionKind {
+    SubWindowDensity,
+    ChainLength,
+    Vrf,
+    StateHash,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum ConsensusLongRangeForkDecision {
+    Keep(ConsensusLongRangeForkResolutionKind),
```
binier, line 53: Can we rename this for consistency?
`Keep` -> `Ignore`
`Take` -> `Use`

binier (changes requested):

snarker/src/consensus/consensus_state.rs
akoptelov: Keep/take is a terminology from Mina implementation, like here:
https://github.com/MinaProtocol/mina/blob/472c82cded7362594be3681cba5d41862477d817/src/lib/consensus/proof_of_stake.ml#L2959

I thought that using it is a good idea.

akoptelov: I agree about consistency. We can choose either Mina terminology or Ignore/Use.

Also I would get rid of negated enum for the short range fork.

binier: Cool. Let's use mina terminology then for short range fork also.

akoptelov:
@binier I've synced up all the names between long and short forks, also extracted short range decision to a function and added some tests for short range forks.

binier (approved):
Looks good. Feel free to merge.

Issue #44, opened 2023-07-04
As a developer, I need to implement custom logging in node


binier:
`tracing` library is used for logging in Openmina.

Implement custom subscriber which will process those logs and print to terminal.
- Instead of current time, logger should use passed `time`.
- Have an ENV var which can customize whether we log details or not. If not we should only display `summary` part of the log.

akoptelov:
@binier Hi Zura, I can handle this one. Some comments/questions.

1. With relation to time, I guess you want to have a timestamp of an event, right? Shouldn't we instead have it as an additional log entry property? Otherwise we can end up in a situation where we have strange order of timestamps in logs.
2. I think verbosity should be based on entry kind, instead of printing the same log entry differently depending on requested level. As an example, consider a log entry on a huge message reception. If there is a single log entry with both summary and message content, it would be quite hard to analyze the log with its content included. On the other hand, if verbosity is decreased, then there would be no way of getting that content. Instead, we can have an entry containing a summary, with e.g. `debug` level, and another one with e.g. `trace` level containing the message itself. That way we can just use grep/awk to read high-level logs, and then focus on specific data.

binier:
@akoptelov 

> Otherwise we can end up in a situation where we have strange order of timestamps in logs.

Yes if we mix usage of monotonic time and system time like it's done now. But if we store these two as a static vars: https://github.com/openmina/redux-rs/blob/155f8208f3105174cbd32172e7580b164d772349/src/store.rs#L56-L57
We could use just monotonic time everywhere.

---

> I think verbosity should be based on entry kind, instead of printing the same log entry differently depending on requested level

I had in mind additional flag plus verbosity level.

> If there is a single log entry with both summary and message content, it would be quite hard to analyze the log with its content included

I'd say it's mostly useful if we will use some tool to process those logs. Basically what we had for webnode in FE.

> On the other hand, if verbosity is decreased, then there would be no way of getting that content

True, but same is the case with low verbosity logging as well. It's a tradeoff. In general summary should contain enough info to somewhat reason about the thing.
Still you might be right. I don't really have a preference for that. One thing to keep in mind though is that existing logs were written for webnode to be viewed in the frontend. So many of them include big data like full blocks as a log entry fields.

akoptelov:
> @akoptelov
> 
> > Otherwise we can end up in a situation where we have strange order of timestamps in logs.
> 
[...]

PR #52, opened 2023-07-07
Tracing console output with proper monotonic time

akoptelov:
- [x] Use standard tracing subscriber for console output
- [x] Specify monotonic time for trace events

Issue #44

### DIFF
--- a/cli/src/commands/mod.rs
+++ b/cli/src/commands/mod.rs
@@ -3,10 +3,17 @@ pub mod snarker;
 
 pub type CommandError = Box<dyn std::error::Error>;
 
-#[derive(Debug, structopt::StructOpt)]
-#[structopt(name = "openmina", about = "Openmina Cli")]
+#[derive(Debug, clap::Parser)]
+#[command(name = "openmina", about = "Openmina Cli")]
+pub struct OpenminaCli {
+    #[command(subcommand)]
+    pub command: Command,
+}
+
+#[derive(Debug, clap::Subcommand)]
 pub enum Command {
     Snark(snark::Snark),
+    /// Standalone snarker
     Snarker(snarker::Snarker),
 }
 
--- a/cli/src/commands/snark/mod.rs
+++ b/cli/src/commands/snark/mod.rs
@@ -1,16 +1,21 @@
 pub mod precalculate_block_verifier_index_and_srs;
 pub use precalculate_block_verifier_index_and_srs::PrecalculateBlockVerifierIndexAndSrs;
 
-#[derive(Debug, structopt::StructOpt)]
-#[structopt(name = "openmina", about = "Openmina Cli")]
-pub enum Snark {
+#[derive(Debug, clap::Args)]
+pub struct Snark {
+    #[command(subcommand)]
+    pub command: SnarkCommand,
+}
+
+#[derive(Debug, clap::Subcommand)]
+pub enum SnarkCommand {
     PrecalculateBlockVerifierIndexAndSrs(PrecalculateBlockVerifierIndexAndSrs),
 }
 
 impl Snark {
     pub fn run(self) -> Result<(), crate::CommandError> {
-        match self {
-            Self::PrecalculateBlockVerifierIndexAndSrs(v) => v.run(),
+        match self.command {
+            SnarkCommand::PrecalculateBlockVerifierIndexAndSrs(v) => v.run(),
         }
     }
 }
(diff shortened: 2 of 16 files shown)


akoptelov:
@binier I've also added submodules, now it is easier to just modify our dependencies right in the project tree.

cli/src/commands/snarker/mod.rs
```diff
@@ -34,23 +36,33 @@ mod http_server;
...
 }
 
 impl Snarker {
     pub fn run(self) -> Result<(), crate::CommandError> {
-        // stderrlog::new()
-        //     .verbosity(5)
-        //     // .module(module_path!()) // remove this line to see libp2p logs.
-        //     .init()
-        //     .unwrap();
+        unsafe { backtrace_on_stack_overflow::enable() };
```
binier, line 63: We don't need this anymore.

node/shared/src/block.rs
```diff
@@ -10,18 +10,36 @@ pub use mina_p2p_messages::v2::StateHash as BlockHash;
 pub type ArcBlock = Arc<Block>;
 pub type ArcBlockWithHash = BlockWithHash<Arc<Block>>;
 
-#[derive(Serialize, Deserialize, Debug, Clone)]
+#[derive(Serialize, Deserialize, Clone)]
```
binier, line 13: Tmp changes that were in original branch, We should undo changes in this file.

cli/src/commands/snarker/http_server.rs
```diff
@@ -69,7 +69,8 @@ pub async fn run(port: u16, rpc_sender: super::RpcSender) {
             let result: Option<RpcStateGetResponse> =
                 rpc_sender_clone.oneshot_request(RpcRequest::GetState).await;
 
-            with_json_reply(&result, StatusCode::OK)
+            let result = result.map(|v| v.transition_frontier.sync);
```
binier, line 72: Tmp change, needs to be removed.

p2p/src/channels/rpc/mod.rs
```diff
@@ -222,7 +222,7 @@ impl P2pRpcResponse {
 
         if matches!(result_kind, RpcResultKind::Err) {
             let err = mina_p2p_messages::rpc_kernel::Error::binprot_read(r)?;
-            let err = format!("{:?}", err);
+            let err = dbg!(format!("{:?}", err));
```
binier, line 225: Tmp change, needs to be removed.

snarker/src/ledger/ledger_service.rs
```diff
@@ -186,6 +186,12 @@ impl<T: LedgerService> TransitionFrontierService for T {
         block: ArcBlockWithHash,
         pred_block: ArcBlockWithHash,
     ) -> Result<(), String> {
+        eprintln!(
```
binier, line 189: Tmp change, needs to be removed.

binier (commented):
Looks good. We just need to revert extra changes caused by force push.

cli/src/commands/mod.rs
```diff
@@ -3,10 +3,17 @@ pub mod snarker;
 
 pub type CommandError = Box<dyn std::error::Error>;
 
-#[derive(Debug, structopt::StructOpt)]
-#[structopt(name = "openmina", about = "Openmina Cli")]
+#[derive(Debug, clap::Parser)]
+#[command(name = "openmina", about = "Openmina Cli")]
+pub struct OpenminaCli {
```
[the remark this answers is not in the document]
akoptelov: This is how they suggest using `clap`. I was able to get rid of of that top-level struct, having the enum to be a `clap::Parser`, but it does not work for inner subcommands. Another reason, is that there might be some common options, and they should be added to the `OpenminaCli` struct. I would keep it like this.

cli/src/commands/snark/mod.rs
```diff
@@ -1,16 +1,21 @@
 pub mod precalculate_block_verifier_index_and_srs;
 pub use precalculate_block_verifier_index_and_srs::PrecalculateBlockVerifierIndexAndSrs;
 
-#[derive(Debug, structopt::StructOpt)]
-#[structopt(name = "openmina", about = "Openmina Cli")]
-pub enum Snark {
+#[derive(Debug, clap::Args)]
+pub struct Snark {
```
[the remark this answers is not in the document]
akoptelov: See my prev comment.

cli/src/commands/mod.rs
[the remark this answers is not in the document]
binier: Yeah but from my experience, cli that has multiple layers on which args can be passed, is harder to work with. But ok, that works.

binier (approved):
Looks good, thanks.

Issue #47, opened 2023-07-04
As a developer, I need to implement program for creating snarks in a separate process and communication protocol with it


binier:

binier:
This runs ocaml snark worker process: https://github.com/openmina/mina/blob/4c57c37fcd2a5674adad7116ebde19195dd17310/src/lib/snark_worker/functor.ml#L228

Since we don't yet have proof generation logic in Rust, we will have to use that.

I think sexp inputs can be passed as a cli arguments to this standalone worker, I don't remember exactly. 

We might need to modify this a bit to accept binprot encoded inputs from the Rust process.

akoptelov:
- Started implementing statemachine-side data/code for handling external processes
- Started implementing RPC protocol with Mina snark worker executable

akoptelov:
Implementing assertions:
- [x] commitments are not greater than workers
- [x] ready snark work is increasing
- [ ] snark value assertions
- [ ] snark order assertions

https://ci.openmina.com/openmina/openmina/79/1/7

akoptelov:
- fixed cancel-related redux
- dockerized/k8s-zed openmina
- improved logging

PR #58, opened 2023-08-04
Poc/mina snark worker int poc

akoptelov:
DONE:
- [x] External snark worker process handling
- [x] Automatic commitment creation
- [x] Workers state RPC

TODO (for other PRs):
- [ ] Multiple snark workers
- [ ] Cancellation policy
- [ ] Improve snark workers reporting
- [ ] CI

Fix: #47


cli/src/commands/snarker/http_server.rs
```diff
@@ -191,12 +219,95 @@ pub async fn run(port: u16, rpc_sender: super::RpcSender) {
...
+                                        == Some("application/octet-stream") =>
+                                {
+                                    JsonOrBinary::binary(spec)
+                                }
+                                RpcSnarkerJobSpecResponse::Ok(spec) => JsonOrBinary::json(spec),
+                                _ => JsonOrBinary::error("error", StatusCode::BAD_REQUEST),
+                            },
+                        )
+
+                    // let res: Option<RpcSnarkerJobSpecResponse> = rpc_sender_clone
```
binier, line 258: We can remove these right?

node/shared/src/snark_job_id.rs
```diff
@@ -23,15 +23,15 @@ pub struct LedgerHashTransitionPasses {
 
 impl std::fmt::Display for LedgerHashTransition {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}->{}", &self.source, &self.target)
+        write!(f, "{}-{}", &self.source, &self.target)
```
binier, line 26: I used `->` to signify a ledger transition. Do we need to change that?

node/shared/src/snark_job_id.rs
```diff
@@ -53,7 +53,7 @@ impl FromStr for LedgerHashTransitionPasses {
     type Err = ();
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let (first_pass, second_pass) = s.split_once(":").ok_or(())?;
+        let (first_pass, second_pass) = s.split_once("_").ok_or(())?;
```
binier, line 56: With dash and underscore it might be harder to read.

snarker/src/external_snark_worker/external_snark_worker_effects.rs
```diff
@@ -0,0 +1,92 @@
...
+        ExternalSnarkWorkerAction::SubmitWork(action) => {
+            let job_id = &action.job_id;
+            let config = &store.state().config;
+            let public_key = config.public_key.clone();
+            let fee = config.fee.clone();
+            let Some(job) = store.state().snark_pool.get(job_id) else {
+                return;
+            };
+            let protocol_state_body = |block_hash: StateBodyHash| {
+                store
```
binier, line 48: We should try to find protocol state here as well: https://github.com/openmina/openmina/blob/7a23546e3fadb33de4d0c49c13668539c8442568/snarker/src/transition_frontier/transition_frontier_state.rs#L16

Job might refer to block older than our root. In such case protocol state will be there.

snarker/src/external_snark_worker/external_snark_worker_actions.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ExternalSnarkWorkerWorkResultAction {
+    pub result: super::SnarkWorkResult,
+}
+
+impl EnablingCondition<State> for ExternalSnarkWorkerWorkResultAction {
+    fn is_enabled(&self, #[allow(unused_variables)] state: &State) -> bool {
+        true
```
binier, line 94: Some of these should be enabled only if we are in working state right?

binier (changes requested):

node/shared/src/snark_job_id.rs
akoptelov: This is much more friendly to URL, when you need to copy-paste it to e.g. `curl`

node/shared/src/snark_job_id.rs
akoptelov: The same as above

snarker/src/external_snark_worker/external_snark_worker_effects.rs
akoptelov: I've replaced the closure with a fn in the transition frontier. BTW, we should have a map from state body hashes, to avoid recalculating them each time...

binier (approved):

Issue #59, opened 2023-08-04
As a developer, I need to implement handling for incoming rpcs from libp2p peers


binier:
Rpc handling in state machine was implemented in commit: https://github.com/openmina/openmina/commit/210cad536b4a8d8c4f57888bd4ed877f37be4b60

Service part is implemented for WebRTC, but it's not implemented for LibP2P.

Branch that should be used as a base: `feat/standalone_snark_worker`.

In libp2p service, rpc behaviour needs to be extend so that we can handle incoming rpcs: https://github.com/openmina/openmina/tree/7a23546e3fadb33de4d0c49c13668539c8442568/p2p/src/service_impl/libp2p/rpc

Events from rpc behaviour is handled [here](https://github.com/openmina/openmina/blob/7a23546e3fadb33de4d0c49c13668539c8442568/p2p/src/service_impl/libp2p/mod.rs#L348) which basically simply forwards the event to state machine. I don't think we need any changes there.

We simply need to recieve rpc message in state machine: https://github.com/openmina/openmina/blob/7a23546e3fadb33de4d0c49c13668539c8442568/p2p/src/channels/rpc/mod.rs#L38

Then response message will be sent to libp2p service and from here we need to add code so that we send that response to libp2p peer: https://github.com/openmina/openmina/blob/7a23546e3fadb33de4d0c49c13668539c8442568/p2p/src/service_impl/libp2p/mod.rs#L232

PR #60, opened 2023-08-06
feat(standalone_snark_worker_new_libp2p_rpc_behaviour): implement responding on incoming queries

vlad9486:
#59 

Use my crate here. We can move it in your repository if you prefer.


p2p/src/service_impl/libp2p/mod.rs
```diff
@@ -368,6 +501,183 @@ impl Libp2pService {
         }
     }
 
+    fn handle_event_rpc<E: From<P2pEvent>>(
+        swarm: &mut Swarm<Behaviour<E>>,
+        peer_id: PeerId,
+        event: RpcBehaviourEvent,
+    ) {
+        let sender = swarm.behaviour_mut().event_source_sender.clone();
+        let sender = |event: P2pEvent| {
```
binier, line 510: Maybe `send` would be a better name here?

p2p/src/service_impl/libp2p/mod.rs
```diff
@@ -368,6 +501,183 @@ impl Libp2pService {
...
+                    ch_sender(P2pEvent::Channel(P2pChannelEvent::Received(
+                        peer_id.into(),
+                        Ok(ChannelMsg::Rpc(msg)),
+                    )))
+                };
+
+                fn parse_q<M: RpcMethod>(bytes: Vec<u8>) -> M::Query {
+                    let mut bytes = bytes.as_slice();
+                    <QueryPayload<M::Query> as BinProtRead>::binprot_read(&mut bytes)
+                        .unwrap()
```
binier, line 545: We can't have panics when parsing received messages.

p2p/src/service_impl/libp2p/mod.rs
```diff
@@ -368,6 +501,183 @@ impl Libp2pService {
...
+                        .0
+                }
+
+                fn parse_r<M: RpcMethod>(bytes: Vec<u8>) -> Result<M::Response, RpcError> {
+                    let mut bytes = bytes.as_slice();
+                    <ResponsePayload<M::Response> as BinProtRead>::binprot_read(&mut bytes)
+                        .unwrap()
```
binier, line 552: We can't have panics when parsing received messages.

binier (changes requested):
@vlad9486  Looks Great! Thanks. Just go over requested changes and let's merge this.

binier (changes requested):
@vlad9486 Last change. Could you move `rpc-behaviour` crate inside `p2p` crate and rename to `libp2p-rpc-behaviour`? Otherwise on top level having that dir will be confusing...

Ideally we should extend bootstrap test to make sure this works, as in we are able to bootstrap Ocaml node and another Rust node. But let's do that in another pr.

binier (approved):

PR #133, opened 2023-11-16
Implement root snarked ledger synchronization

### LINKED ISSUE #91: As a developer, I need to implement logic for maintaining root snarked ledger, so that rust nodes can bootstrap other nodes
We synchronize root snarked ledger initially during bootstrap,
but as root block moves, once snarked ledger changes, we need to update our snarked ledger mask.

Once we are bootstrapped, we need to always have snarked ledger mask stored here:
https://github.com/openmina/openmina/blob/ef53e52c4da959d9830d3aa55d96e14ced719d25/node/src/ledger/ledger_service.rs#L70

We need to add logic [here](https://github.com/openmina/openmina/blob/ef53e52c4da959d9830d3aa55d96e14ced719d25/node/src/transition_frontier/transition_frontier_effects.rs#L150) so that when catchup is finished, we update our root snarked ledger if we have to.
There we should compare old chain (`state.transition_frontier.chain`) with new chain (`state.transition_frontier.sync.chain`) and if root snarked ledger is different, we need to update the mask in the service.

No need for garbage collection as it's already done here: https://github.com/openmina/openmina/blob/ef53e52c4da959d9830d3aa55d96e14ced719d25/node/src/ledger/ledger_service.rs#L335

tizoc:
Fixes #91

### DIFF
--- a/node/src/ledger/ledger_service.rs
+++ b/node/src/ledger/ledger_service.rs
@@ -9,16 +9,22 @@ use ledger::{
         currency::{Amount, Fee, Slot},
         scan_state::{
             AvailableJobMessage, ConstraintConstants, JobValueBase, JobValueMerge,
-            JobValueWithIndex,
+            JobValueWithIndex, Pass,
         },
-        transaction_logic::{local_state::LocalState, protocol_state::protocol_state_view},
+        transaction_logic::{
+            local_state::LocalState,
+            protocol_state::{protocol_state_view, ProtocolStateView},
+            transaction_partially_applied::TransactionPartiallyApplied,
+            Transaction,
+        },
     },
+    sparse_ledger::SparseLedger,
     staged_ledger::{
         diff::Diff,
         staged_ledger::{SkipVerification, StagedLedger},
     },
     verifier::Verifier,
-    AccountIndex, BaseLedger, Mask, TreeVersion,
+    AccountIndex, BaseLedger, Mask, TreeVersion, UnregisterBehavior,
 };
 use mina_hasher::Fp;
 use mina_p2p_messages::v2::{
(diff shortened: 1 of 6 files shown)


node/src/ledger/ledger_service.rs
```diff
@@ -324,6 +334,123 @@ impl<T: LedgerService> TransitionFrontierService for T {
...
+        let root_snarked_ledger = ctx
+            .snarked_ledgers
+            .remove(old_root.snarked_ledger_hash())
+            .ok_or_else(|| format!(
+                "push_snarked_ledger: could not find old root snarked ledger: {} (in_staged: {}, in_sync_staged: {})",
+                old_root.snarked_ledger_hash().to_string(),
+                ctx.staged_ledgers.keys().any(|h| h == old_root.snarked_ledger_hash()),
+                ctx.sync.staged_ledgers.keys().any(|h| h == old_root.snarked_ledger_hash())
+            ))?;
+        let mut mt = root_snarked_ledger.make_child();
```
binier, line 357: Do we create this tmp mask so that eventually in case of error, we discard it and reinsert the old snarked ledger?

binier (approved):
Only one question. Otherwise approved for merge.

node/src/ledger/ledger_service.rs
tizoc: @binier yes, you want to avoid mutating the original ledger until you are sure there will be no failures. Reinsertion would not be needed because it was not removed (in the same way the new one is not inserted until it is ready).

These failures should not happen in practice tho unless the implementation is buggy (but the way to figure out if it is buggy or not is to have these checks). All the transactions that are re-applied here have been applied before when they came in a block.

Btw note that right now the node cannot recover from this, in the caller there is a `panic!` if this `push_snarked_ledger` call fails with an error.

PR #150, opened 2023-12-07
Improve p2p testing (doc/impl)

akoptelov:
In order to understand the quality of our p2p layer we need a good structure of p2p tests, so it is obvious what aspects of p2p are covered by tests, and what each test actually tests.

### DIFF
--- a/node/src/p2p/listen/p2p_listen_actions.rs
+++ b/node/src/p2p/listen/p2p_listen_actions.rs
@@ -4,25 +4,25 @@ use redux::EnablingCondition;
 use crate::State;
 
 impl EnablingCondition<State> for P2pListenNewAction {
-    fn is_enabled(&self, #[allow(unused_variables)] state: &State) -> bool {
+    fn is_enabled(&self, state: &State) -> bool {
         self.is_enabled(&state.p2p)
     }
 }
 
 impl EnablingCondition<State> for P2pListenExpiredAction {
-    fn is_enabled(&self, #[allow(unused_variables)] state: &State) -> bool {
+    fn is_enabled(&self, state: &State) -> bool {
         self.is_enabled(&state.p2p)
     }
 }
 
 impl EnablingCondition<State> for P2pListenErrorAction {
-    fn is_enabled(&self, #[allow(unused_variables)] state: &State) -> bool {
+    fn is_enabled(&self, state: &State) -> bool {
         self.is_enabled(&state.p2p)
     }
 }
 
 impl EnablingCondition<State> for P2pListenClosedAction {
-    fn is_enabled(&self, #[allow(unused_variables)] state: &State) -> bool {
+    fn is_enabled(&self, state: &State) -> bool {
         self.is_enabled(&state.p2p)
     }
 }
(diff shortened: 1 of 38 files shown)


node/src/p2p/listen/p2p_listen_actions.rs
```diff
@@ -0,0 +1,28 @@
+use p2p::listen::{P2pListenNewAction, P2pListenExpiredAction, P2pListenErrorAction, P2pListenClosedAction};
+use redux::EnablingCondition;
+
+use crate::State;
+
+impl EnablingCondition<State> for P2pListenNewAction {
+    fn is_enabled(&self, #[allow(unused_variables)] state: &State) -> bool {
```
tizoc, line 7: is this annotation needed? `state` is referenced bellow

node/testing/src/node/config.rs
```diff
@@ -9,6 +9,14 @@ pub enum NodeTestingConfig {
     Rust(RustNodeTestingConfig),
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
+pub enum TestPeerId {
+    #[default]
+    Derived,
```
binier, line 15: Can you add commit for `Derived` for a warning that it might cause issues if scenario interacts with an external program, E.g. replayer, ocaml node, which aren't under the control of the framework. In case of multiple parallel runs, peer ids will be same for the nodes in those runs and external program might disconnect one because of that.

node/testing/src/scenarios/driver.rs
```diff
@@ -0,0 +1,363 @@
...
+use super::ClusterRunner;
+
+pub fn match_addr_with_port_and_peer_id(
+    port: u16,
+    peer_id: PeerId,
+) -> impl Fn(&P2pConnectionOutgoingInitOpts) -> bool {
+    move |conn_opt| match conn_opt {
+        P2pConnectionOutgoingInitOpts::LibP2P(libp2p_opts) => {
+            &libp2p_opts.peer_id == &peer_id && libp2p_opts.port == port
+        }
```
binier, line 24: Why not compare webrtc variant as well?

node/testing/src/scenarios/driver.rs
```diff
@@ -0,0 +1,363 @@
...
+pub fn identify_event(peer_id: PeerId) -> impl Fn(ClusterNodeId, &Event, &State) -> bool {
+    move |_, event, _| {
+        matches!(
+            event,
+            Event::P2p(P2pEvent::Libp2pIdentify(peer, _)) if peer == &peer_id
+        )
+    }
+}
+
+pub struct Driver<'cluster> {
```
binier, line 104: Do we need more abstraction on top of `ClusterRunner`? Roles between `Driver` and `ClusterRunner` seems to be the same. Could we move this logic to `ClusterRunner` instead? If it's the name, we can rename it to `Driver` if that's better.

node/testing/src/node/config.rs
akoptelov: @binier You mean, a comment, not commit?

node/testing/src/scenarios/driver.rs
akoptelov: Well, driver is a bunch of utilities on top of the cluster runner. Until it is more clear what would be separation, let's keep it like this.

tizoc (approved):
Looks good to me, but @binier should be able to give much better feedback than me.

Issue #124, opened 2023-11-02
As a developer, I need to implement block producing logic (without proof generation) for testing purposes


binier:
Goal is to prepare state machine without implementing:
1. Logic to fetch/prepare data from epoch ledger. Blocked by: #103 
1. Vrf evaluator. "interface"/placeholder state machine will be implemented for it, without implementing an actual logic in the scope of this task.

binier:
Implemented most of the state machine for block producer.

Implemented most of the block construction logic. Missing construction of: 
- `min_window_density`  in `consensus_state`
- `sub_window_densities`  in `consensus_state`

binier:
Wrapped up epoch ledgers syncing and issues with maintaining root snarked ledger.

Now getting back to working on block producer and tests for it.

PR #151, opened 2023-12-08
Block Producer

binier:
closes: #124


node/src/block_producer/vrf_evaluator/block_producer_vrf_evaluator_effects.rs
```diff
@@ -0,0 +1,131 @@
...
+                // slot is in the current epoch
+                if next_slot <= current_epoch_end {
+                    let vrf_input: VrfEvaluatorInput = VrfEvaluatorInput::new(
+                        current_epoch_data.seed.clone(),
+                        current_epoch_data.delegator_table.clone(),
+                        next_slot,
+                        current_epoch_data.total_currency,
+                        current_epoch_data.ledger.clone(),
+                    );
+                    store.dispatch(BlockProducerVrfEvaluatorEvaluateVrfAction { vrf_input });
```
binier, line 73: Enabling condition of `BlockProducerVrfEvaluatorEvaluateVrfAction` should contain all the above checks.

node/src/block_producer/vrf_evaluator/block_producer_vrf_evaluator_effects.rs
```diff
@@ -0,0 +1,131 @@
...
+                &vrf_evaluator_state.current_epoch,
+                &vrf_evaluator_state.current_epoch_data,
+                &vrf_evaluator_state.next_epoch_data,
+            ) {
+                let current_epoch_end = current_epoch * SLOTS_PER_EPOCH + SLOTS_PER_EPOCH - 1;
+                let next_epoch_end = (current_epoch + 1) * SLOTS_PER_EPOCH + SLOTS_PER_EPOCH - 1;
+
+                // slot is in the current epoch
+                if next_slot <= current_epoch_end {
+                    let vrf_input: VrfEvaluatorInput = VrfEvaluatorInput::new(
```
binier, line 66: Can we wrap all the inputs except slot in `Arc` so we avoid bunch of extra cloning? This clone can become expensive with bunch of delegators.

node/src/block_producer/vrf_evaluator/block_producer_vrf_evaluator_effects.rs
binier: One way to do that is to extract all this logic into function which takes state and returns next `VrfEvaluatorInput`. Then in the enabling condition you can verify that passed inputs match the ones returned from that function. For data in `Arc` we can compare just the pointers, otherwise comparing delegate table might be a bit expensive.

binier (commented):
@adonagy reviewed vrf related parts. Please take a look at them. Gonna review and cleanup block producer logic next.

adonagy:
@binier Addressed the review in #201 , please continue the review of the VRF part there :)

PR #154, opened 2023-12-12
Openmina Frontend

directcuteo:
Openmina Frontend Angular Application for Rust based node


tizoc:
@directcuteo do you care about the commit history? @sebastiencs had some script (or the steps required) to keep the original commit history when moving code from another repository

PR #178, opened 2023-12-26
Replace libp2p with state machine implementation

vlad9486:


emberian:
I can help review when desired although most of my focus is on preparing for the release right now

PR #184, opened 2024-01-11
Connect vrf and block producer

adonagy:


binier (approved):
Looks good (apart from compilation error)

adonagy:
@binier Fixed the compilation error and also added the correct slot calculation.

PR #193, opened 2024-01-24
Refactor: Inline action group definitions into the parent enum type 

tizoc:
Example of getting rid of "leaf" types used for single action definitions and replacing them with a single action type for the whole group by inlining all fields in the enum tags.

Updated modules so far:
- [x] p2p_channels_rpc
- [x] p2p_channels_snark
- [x] p2p_channels_snark_job_commitment
- [x] p2p_channels_best_tip
- [x] external_snark_worker
- [x] snark_work_verify
- [x] snark_block_verify
- [x] consensus
- [x] snark_pool_candidate
- [x] snark_pool

### DIFF
--- a/node/build.rs
+++ b/node/build.rs
@@ -48,15 +48,52 @@ fn is_same_file<P: AsRef<Path>>(file1: P, file2: P) -> Result<bool, std::io::Err
     Ok(b1 == b2)
 }
 
+#[derive(Debug, Clone)]
+enum EnumElement {
+    InlineFields {
+        tag: String,
+        action_kind: String,
+        has_fields: bool,
+    },
+    Nested((String, String)),
+}
+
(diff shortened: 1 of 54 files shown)


sebastiencs (approved):
+1 for those changes. It remove some indirections and make the code easier to follow and understand

tizoc:
I am updating other modules in the meantime, but I will wait a bit before merging in case somebody wants to comment on it.

binier:
@tizoc @sebastiencs Maybe we should merge this (at least snark pool part), since @sebastiencs will be working on transaction pool using snark pool as a template.

tizoc:
ok, I will give you one more pass to ensure that nothing has changed its meaning with the changes and will merge the current progress.

PR #195, opened 2024-01-24
Fix for services in libp2p replacement

akoptelov:
@vlad9486 Instead of having new stubs for P2p-related services when internal libp2p is used, it is better to have an alternative implementation for existing ones, with conditional implementations. These services are used by webrtc logic as well, and as far as I concerned, we want to keep it working, either with external or internal libp2p. Also methods like `outgoing_init` should be defined for any p2p implementation.

Please take a look, probably I can make it even more elegant, like to factor out external libp2p service (discovery related methods).


akoptelov:
Another thing that should be done here, is to move all libp2p sensitive events under `#[cfg(...)]` attributes

PR #198, opened 2024-01-24
Fix block application, when the block is empty

sebastiencs:
Before this commit, applying an empty block was incrementing `ParallelScan::curr_job_seq_no`


sebastiencs:
Do not merge yet, I need to check if OCaml increment this counter or not, when applying an empty block

sebastiencs:
Closing, the counter is correctly incremented. Our issue is related to forks

PR #199, opened 2024-01-25
Save `StagedLedger` and block to a file, when the block application fail

sebastiencs:
When a staged ledger hash mismatch occurs, dump the staged ledger + block to a file so we can easily reproduce the bug in Rust and OCaml: https://github.com/openmina/openmina/blob/8e68037aafddd43842a54c8439baeafee4c6e1eb/ledger/src/staged_ledger/staged_ledger.rs#L5959

### DIFF
--- a/node/src/ledger/ledger_service.rs
+++ b/node/src/ledger/ledger_service.rs
@@ -29,7 +29,7 @@ use ledger::{
 };
 use mina_hasher::Fp;
 use mina_p2p_messages::v2::{
-    DataHashLibStateHashStableV1, LedgerHash, MinaBaseAccountBinableArgStableV2,
+    self, DataHashLibStateHashStableV1, LedgerHash, MinaBaseAccountBinableArgStableV2,
     MinaBaseLedgerHash0StableV1, MinaBaseSokMessageStableV1, MinaBaseStagedLedgerHashStableV1,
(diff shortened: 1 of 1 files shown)


sebastiencs:
@binier Would you prefer having this logic in the state machine ?

node/src/ledger/ledger_service.rs
```diff
@@ -837,6 +855,54 @@ impl<T: LedgerService> RpcLedgerService for T {
...
+    let apply_context = ApplyContext {
+        accounts: staged_ledger
+            .ledger()
+            .to_list()
+            .iter()
+            .map(v2::MinaBaseAccountBinableArgStableV2::from)
+            .collect::<Vec<_>>(),
+        scan_state: staged_ledger.scan_state().into(),
+        pending_coinbase: staged_ledger.pending_coinbase_collection().into(),
+        pred_block: (&*pred_block.block).clone(),
```
binier, line 893: Very minor thing but I think `&` is extra here and everywhere where clone is done right after, since clone takes re`&self`.

binier (approved):
Looks good.

node/src/ledger/ledger_service.rs
sebastiencs: @binier You're right. Thanks

PR #201, opened 2024-01-25
Vrf evaluator cleanup

adonagy:
Addressing the review in #151 

Cleaned up the vrf evaluator code accordingly. 

One last thing that is hanging is moving the logic out of `BlockProducerVrfEvaluatorEvaluationSuccessAction` effects to the enabling condition. I don't really grasp the point made in this [comment](https://github.com/openmina/openmina/pull/151#discussion_r1466140896). Moving the same logic to the enabling condition seems a bit redundant to me as the effect will only dispatch the action with values that makes sense (no input sent for 2 epoch in the future as we have no data for that yet).

I think it could be more elegant if we made the vrf evaluator aware what epoch it's currently evaluating but that would require the overhaul of the evaluator state (later polishing).

### DIFF
--- a/node/native/src/block_producer/vrf_evaluator.rs
+++ b/node/native/src/block_producer/vrf_evaluator.rs
@@ -20,7 +20,7 @@ pub fn vrf_evaluator(
     while let Some(vrf_evaluator_input) = vrf_evaluation_receiver.blocking_recv() {
         let mut vrf_result = VrfEvaluationOutput::SlotLost(vrf_evaluator_input.global_slot);
 
-        for (index, account) in vrf_evaluator_input.delegatee_table.iter() {
+        for (index, account) in vrf_evaluator_input.delegator_table.iter() {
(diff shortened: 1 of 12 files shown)


binier (approved):
Looks good.

Regarding having logic in enabling condition.
Basically it's in order to have insurance. As we iterate on code, we might change where that action gets dispatched or how and we might fail to maintain that logic in the effects. If we have it in enabling condition, it will guard us from an invalid dispatch.
Also for action fuzzing. We don't use that and I'm not sure if we'll use it, but enabling conditions were very useful for filtering generated actions.

adonagy:
@binier Ah, that makes perfect sense.
IMO we can merge this and then when we get to refactoring the state for prod version, I will follow the approach to include that logic in the enabling condition.

binier:
@adonagy before refactoring, might be a good idea to have a good suite of tests.

PR #203, opened 2024-01-26
Cache verifier index

akoptelov:
In this commit I've implemented very basic caching of indices. If that would work fine, we should also include SRS hash to the file name (in case input for verifier index is changed). and probably also move it directly to ledger (unless we don't want to have OS ops in that crate)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3408,6 +3408,7 @@ dependencies = [
 name = "mina-tree"
 version = "0.2.0"
 dependencies = [
+ "anyhow",
  "ark-ec",
  "ark-ff",
  "ark-poly",
--- a/ledger/Cargo.toml
+++ b/ledger/Cargo.toml
@@ -69,6 +69,7 @@ blake2 = "0.10"
 crc32fast = "1"
 chrono = "0.4"
 serde_with = "3.6.1"
+anyhow = "1.0.75"
 
 [target.'cfg(target_family = "wasm")'.dependencies]
 getrandom = { version = "0.2", features = ["js"] }
--- a/ledger/src/proofs/verifier_index.rs
+++ b/ledger/src/proofs/verifier_index.rs
@@ -26,7 +26,10 @@ use crate::{
     VerificationKey,
 };
 
-use super::transaction::InnerCurve;
+use super::{
+    caching::{verifier_index_from_bytes, verifier_index_to_bytes},
+    transaction::InnerCurve,
+};
 use super::{
     transaction::endos,
     wrap::{Domain, Domains},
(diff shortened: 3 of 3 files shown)


akoptelov:
BTW, serialized indices are quite small:
```
$ ls -l ~/.cache/openmina/
total 8
-rw-rw-r-- 1 alexander alexander 2268 Jan 26 11:32 block_verifier_index.bin
-rw-rw-r-- 1 alexander alexander 2268 Jan 26 11:33 transaction_verifier_index.bin
```

sebastiencs:
> BTW, serialized indices are quite small:

@akoptelov They are small because `VerifierIndex` is serialized by skipping lots of data, and those data are slow to recompute:
https://github.com/o1-labs/proof-systems/blob/0863042ca383c6578564e166724e9dbc66da19bf/kimchi/src/verifier_index.rs#L134-L151


Instead of using `serde_cbor` to de/serialize the `VerifierIndex`, we should use the existing `verifier_index_{from,to}_bytes` : 
https://github.com/openmina/openmina/blob/8e68037aafddd43842a54c8439baeafee4c6e1eb/ledger/src/proofs/caching.rs#L352
https://github.com/openmina/openmina/blob/8e68037aafddd43842a54c8439baeafee4c6e1eb/ledger/src/proofs/caching.rs#L362
Those 2 methods do not skip any data

node/testing/src/cluster/mod.rs
```diff
@@ -46,10 +51,93 @@ use crate::{
...
+}
+
 lazy_static::lazy_static! {
     static ref VERIFIER_SRS: Arc<VerifierSRS> = get_srs().into();
-    static ref BLOCK_VERIFIER_INDEX: Arc<VerifierIndex> = get_verifier_index(VerifierKind::Blockchain).into();
-    static ref WORK_VERIFIER_INDEX: Arc<VerifierIndex> = get_verifier_index(VerifierKind::Transaction).into();
+    static ref BLOCK_VERIFIER_INDEX: Arc<VerifierIndex> = {
+        const NAME: &str = "block_verifier_index.bin";
+        read_index(NAME).map(|v| {
+            info!(system_time(); "Block verifier index is loaded from {NAME}");
```
sebastiencs, line 114: Small detail, maybe we should print the full path here, and below ?

akoptelov:
@sebastiencs now they are much bigger:
```
$ ls -l ~/.cache/openmina/*
-rw-rw-r-- 1 alexander alexander 5622552 Jan 26 18:44 /home/alexander/.cache/openmina/block_verifier_index.bin
-rw-rw-r-- 1 alexander alexander 5622552 Jan 26 18:46 /home/alexander/.cache/openmina/transaction_verifier_index.bin
```

binier:
@akoptelov For webnode we do have that implementation, so it might be useful. It includes hash verification.

Dumping key + srs: https://github.com/openmina/openmina/blob/webnode/cli/src/commands/snark/precalculate_block_verifier_index_and_srs.rs

Loading: https://github.com/openmina/openmina/blob/6a2dcb43af04fea43b348c05f161e919326ccc7a/node/wasm/src/lib.rs#L261

akoptelov:
@sebastiencs @binier What should happen to SRS (and verifier index) to change? Hard fork? I'm thinking would it be enough to just have 1 file per index kind. That should be enough for a common user, but for developers this might be essential to be able to work with different versions...

sebastiencs:
@akoptelov The Verifier index will change whenever the circuits change.
`rampup8` already contain changes to circuits, I don't know if more changes are planned.

Note that the verifier index is created/derived from a `Prover`.
To create a `Prover` we need its list of gates, I have pushed them all to this repository:
https://github.com/openmina/circuit-blobs/tree/main/rampup4

- All `*_gates.json` are the list of gates, for each circuit
- `*_internal_vars.bin` and `*_rows_rev.bin` are other data we need for witness generation: [link](https://github.com/MinaProtocol/mina/blob/28ad2335c3970b856849137196c7693c18edee78/src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml#L707-L710)
- All others `*.bin` are snark works (`ExternalSnarkWorkerRequest`) used for testings
- `.txt` files are witnesses vectors generated from OCaml, they are used for testings too

So in the future, we will need to use the `*_gates.json` files, to create the provers, and create the verifier index from them.
`block_verifier_index.bin` and `transaction_verifier_index.bin` probably won't be necessary anymore.
Though the advantage of having them is that we can create the verifier index quickly, without having to create their provers (which is quite slow).

> for developers this might be essential to be able to work with different versions

Probably not, we will work only with the original circuits

akoptelov:
OK, than it seems enough for now to have just these two cached files, each containing also a hashsums, for both srs and the index itself.

akoptelov:
@sebastiencs I've moved the caching to the `ledger` crate. Are there any tests that these indices are involved in, so I can check that the loading works fine?

sebastiencs:
@akoptelov There is no such test for now

PR #204, opened 2024-01-26
chored(build): improve dependencies

akoptelov:
- reverse core -> ledger dep
- remove ledger dep from p2p

@sebastiencs I had to comment out ledger address printing in `p2p`. Can you suggest a better way for displaying that `addr`? @binier do we need that info at all?


binier:
@akoptelov Yes we do need that. It's used for event patterns which are used by testing framework for matching events.

Can we move `LedgerAddress` to `openmina_core`?

akoptelov:
> @akoptelov Yes we do need that. It's used for event patterns which are used by testing framework for matching events.
> 
> Can we move `LedgerAddress` to `openmina_core`?

Well, that was what I meant saying "can't find an easy solution" )

Then we should do that, I'll check.

akoptelov:
> @akoptelov Yes we do need that. It's used for event patterns which are used by testing framework for matching events.
> 
> Can we move `LedgerAddress` to `openmina_core`?

Moving `Address` to `openmina_core` is quite a big task. It pulls a lot of dependencies there, so I wouldn't do that. Instead, I've just replicated the needed behaviour in place.

p2p/src/channels/rpc/mod.rs
```diff
@@ -108,15 +109,30 @@ impl Default for P2pRpcRequest {
...
+        .copied()
+        .map(|byte| {
+            (0..8)
+                .into_iter()
+                .map(|b| byte & (1 << (7 - b)) != 0)
+                .map(|b| if b { '1' } else { '0' })
+                .collect::<String>()
+        })
+        .collect()
+}
```
akoptelov, line 124: @sebastiencs here I've replicated `Into<Address>::to_string()` behaviour from `ledger` crate. Can you double-check the `Address::to_string()` implementation? There's something fishy about it: we're getting bit by bit, but only `length` times, while when constructing an address we use that `length` to count bytes, not bits...

https://github.com/openmina/openmina/blob/build%2Fgroom-deps/core/src/ledger/address/raw.rs#L482

sebastiencs: > when constructing an address we use that length to count bytes

Do you mean when we construct an `Address` from a string ? ([here ?](https://github.com/openmina/openmina/blob/8e68037aafddd43842a54c8439baeafee4c6e1eb/ledger/src/address/raw.rs#L20))

That's because the string representation uses 1 byte for each bit:
- `"01011"`: those 5 bytes represent 5 bits.
- `0b01011` Here the bits/native format are just 5 bits

An address of depth 5 needs 5 bits, or 5 bytes in string format

p2p/src/channels/rpc/mod.rs
```diff
@@ -108,15 +109,30 @@ impl Default for P2pRpcRequest {
...
             Self::BestTipWithProof => Ok(()),
             Self::LedgerQuery(ledger_hash, query) => {
-                let addr_to_str = |addr| {
-                    let addr = ledger::Address::from(addr);
-                    format!("depth: {}, addr: {}", addr.length(), addr.to_string())
+                let addr_to_str = |addr: &MerkleAddressBinableArgStableV1| {
+                    let depth = &addr.0 .0;
+                    let bits = to_binary(&addr.1[..(*depth as usize)]);
+                    format!("depth: {depth}, addr: {bits}")
                 };
```
sebastiencs, line 136: Here `depth` is the number of bits in `addr.1`. 
```
&addr.1[..(*depth as usize)]
```
This is not correct, because `depth` is not the number of bytes.

That `addr_to_str` should be:
```
fn addr_to_str(addr: &MerkleAddressBinableArgStableV1) -> String {
    let MerkleAddressBinableArgStableV1(depth, bits) = addr;
    let depth = depth.as_u64() as usize;

    let addr: String = (0..depth)
        .map(|index| {
            let byte_index = index / 8;
            let bit_index = index % 8;
            bits[byte_index] & (1 << (7 - bit_index)) != 0
        })
        .map(|b| if b { '1' } else { '0' })
        .collect();

    format!("depth: {depth}, addr: {addr}")
}
```

sebastiencs (changes requested):

p2p/src/channels/rpc/mod.rs
akoptelov: Ah, right, got confused with `depth` and `pos` in the `From` impl. Will fix that.

sebastiencs: @akoptelov Ah yes, that `pos` should use a different name,

sebastiencs (approved):

p2p/src/channels/rpc/mod.rs
akoptelov: No, I was looking at `to_string` and `from()` impls. I've been mislead by `lenght` used in two different contexts in the later one:
```rust
...
        let common_length = NBYTES.min(pos.len());
...
            length: depth.as_u64() as _,
```

The functionality is fine.

binier (approved):

PR #218, opened 2024-02-07
fix(messages): generate new wire types

akoptelov:
The summary of Mina p2p types change:
- New bounded types introduced for String and Array. Actually, there is one global limit for the String (100M bytes) and two limited arrays (capped with 16 and 4000 items)
- Some Strings wrapped in types are used directly (e.g. `token_symbol` and `zkapp_uri`)
- Consensus state field `curr_global_slot` is renamed to `curr_global_slot_since_hard_fork`
- In the ProtocolVersion, `major`/`minor` renamed to `transaction`/`network`
- ProtocolVersion's version is changed from 1 to 2 (shouldn't affect bin_prot).
- For ZkApps, `Permissions.set_verification_key` field is extended with a UInt (specifying protocol version, if I'm not mistaken)
- A new protocol constant is added, `grace_period_slot`
- Account precondition representation is simplified: before there were Full, Nonce and Accept, now it is the previous content of the Full variant.

Also now all integers in the generated code are unsigned.


akoptelov:
Update:
- `ArrayN16` is used in most cases for bounded arrays (types generator cannot distinguish between `ArrayN16` and `ArrayN4000` for now), we need to review if that array type is correct.
- New `List` type is used for OCaml `list` type, that is backed by `LinkedList`
- Most of the type conversions in ledger/node were updated to use iterators to be collection-agnostic.

ledger/src/scan_state/conv.rs
```diff
@@ -1554,9 +1532,10 @@ impl From<&SignedCommand> for MinaBaseSignedCommandStableV2 {
 
 impl From<&MinaBaseZkappCommandTStableV1WireStableV1> for zkapp_command::ZkAppCommand {
     fn from(cmd: &MinaBaseZkappCommandTStableV1WireStableV1) -> Self {
+        // TODO: is to_wire needed? if yes, implement it for mut iterator
         Self {
             fee_payer: (&cmd.fee_payer).into(),
-            account_updates: (&cmd.account_updates).into(),
+            account_updates: (&cmd.account_updates).iter().collect(),
```
akoptelov, line 1538: @sebastiencs will we ever need of_wire? currently it is noop.

ledger/src/scan_state/conv.rs
```diff
@@ -1566,7 +1545,7 @@ impl From<&zkapp_command::ZkAppCommand> for MinaBaseZkappCommandTStableV1WireSta
     fn from(cmd: &zkapp_command::ZkAppCommand) -> Self {
         Self {
             fee_payer: (&cmd.fee_payer).into(),
-            account_updates: (&cmd.account_updates).into(),
+            account_updates: (&cmd.account_updates).iter().map(Into::into).collect(),
```
akoptelov, line 1548: @sebastiencs will we ever need to_wire? currently it is noop.

PR #222, opened 2024-02-11
Fix compilation on nightly toolchain

sebastiencs:
Fix compilation on `rustc 1.78.0-nightly (6cc484351 2024-02-10)`
Ran the following commands, to update some dependencies:
```
cargo update -p curve25519-dalek@4.1.1
cargo update -p rand_core@0.5.1
cargo update -p ahash@0.7.7
cargo update -p ahash@0.8.6
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -143,9 +143,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.7.7"
+version = "0.7.8"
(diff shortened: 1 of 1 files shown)


tizoc (approved):
lgtm, but I also think we should switch to a stable version for now since nightly doesn't seem to be required for anything https://github.com/openmina/openmina/pull/221

sebastiencs:
@tizoc Yes this is only to be able to compile locally, using nightly

PR #227, opened 2024-02-15
fix(berkeley-rc1): fix RPCs

akoptelov:
- bump version for answer_sync_ledger_query
- sync up with the new Error type

@tizoc Following up our discussion in Slack. Now I assume that the only option is to have a sexp representation for errors. Can you confirm that? Or should we extend that at some point?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3250,6 +3250,7 @@ dependencies = [
  "mina-poseidon",
  "mina-signer",
  "o1-utils",
+ "rsexp",
  "serde",
  "serde_json",
  "sha2 0.10.8",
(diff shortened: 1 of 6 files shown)


tizoc:
That is correct, for the bin_prot representation they always convert the errors to their Sexp representation, which is then parsed back (but kept as `Info.Sexp`):

https://github.com/MinaProtocol/mina/blob/1551e2faaa246c01636908aabe5f7981715a10f4/src/lib/bounded_types/bounded_types.ml#L148-L156

and `Error.t_of_sexp` will just wrap the sexp in `Message.Sexp` which is the same as `Info.Sexp`

tizoc (approved):
lgtm, and I see rsexp claims to be compatible with OCaml's sexplib so I would expect for it to behave the same way when parsing.

PR #254, opened 2024-03-01
Move time-related enabling conditions to p2p

### LINKED ISSUE #253: Enabling conditions are split across codebase (e.g. p2p/node for timeout)
Currently for triggering timeouts enabling conditions are used that access current time via `state.time()`.

That means that such enabling conditions only can be defined for the global state, and it is impossible to do that in `p2p` crate for p2p actions.

For example, for testing local to p2p module we need to have that timeouts available there.

akoptelov:
closes #253 

@binier ~Note that now time passed to enabling condition is slightly differ from what is in the `state.time()`, but I think this is more consistent, as `state.time()` might jump back, and time that comes from redux is monotonous.~ See discussion below.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -33,14 +33,14 @@ pub trait EnablingCondition<State> {
     /// Enabling condition for the Action.
     ///
     /// Checks if the given action is enabled for a given state.
-    fn is_enabled(&self, state: &State) -> bool {
+    fn is_enabled(&self, state: &State, time: Timestamp) -> bool {
         ...
     }
 }
 ```
 
-`is_enabled(state)` must return `false`, if action doesn't make sense given
-the current state.
+`is_enabled(state, time)` must return `false`, if action doesn't make sense given
+the current state and, optionally, time.
 
 For example message action from peer that isn't connected or we don't know
 about in the state, must not be enabled.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4971,7 +4971,7 @@ dependencies = [
 [[package]]
 name = "redux"
 version = "0.1.0"
-source = "git+https://github.com/openmina/redux-rs.git?branch=feat/global-time#056887aa168c018d5da7d34eb44ad37c76e1508c"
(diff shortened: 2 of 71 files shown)


binier:
`state.time()` can't jump back. Why do you think so? In fact it should always be equal to the time that you are passing now to the enabling condition, from what I checked briefly.

Because `state.time()` also gets the last dispatched action time seen by the reducer and it will match `last_action_id` in the `Store`. https://github.com/openmina/redux-rs/blob/0a85a1efb6c817f9714010b801efc7cd882cb947/src/store.rs#L175

akoptelov:
There are some issues still, probably I screwed up enabling conditions when merged them.

akoptelov:
@binier Zura, how about that? Having additional time-aware method proved to be error-prone, so I've just added the `time` parameter to `is_enabled`.

tizoc:
Is this required because the substates don't have access to the top-level state time?

What about adding a new method to the trait that is defined as:

```rust
fn is_enabled_with_time(&self, state: &State, _time: Timestamp) -> bool {
    self.is_enabled(state)
}
```

which is what redux would call, but it allows for the current enabling conditions that do not require this workaround to keep their current definitions?

akoptelov:
> Is this required because the substates don't have access to the top-level state time?
Well, partially. On the other hand, keeping time in state is kind of a workaround.


> What about adding a new `is_enabled_with_time` method (which redux will call instead of `is_enabled`) to the trait that is defined as:

That was my first idea, but with current impl we need to be super-careful implementing `EnablingCondition` for the global state. By default it calls `is_enabled`, and if it is `is_enabled_with_time` that is used for the substate, then it will be a mistake. With only one method we get rid of such errors.

akoptelov:
@binier If you're fine with this, I'd like to merge it. @binier also this one: https://github.com/openmina/redux-rs/pull/1

tizoc:
@akoptelov right now I cannot come up with a better alternative that solves the issue you mentioned above, so +1 from me

PR #272, opened 2024-03-11
VRF Evaluator refactor + tests

adonagy:
@binier Here is the production version refactor. I'm still cleaning up some tests, but please review the state machine part


node/src/block_producer/vrf_evaluator/block_producer_vrf_evaluator_state.rs
```diff
@@ -1,72 +1,73 @@
...
-    // TODO(adonagy): move to block producer state probably
-    pub current_epoch: Option<u32>,
-    pub current_best_tip_slot: u32,
     pub latest_evaluated_slot: u32,
-    pub last_possible_evaluation_slot: u32,
     pub genesis_timestamp: redux::Timestamp,
+    last_evaluated_epoch: Option<u32>,
+    last_block_heights_in_epoch: BTreeMap<u32, u32>,
+    pending_evaluation: Option<PendingEvaluation>,
+    epoch_context: EpochContext,
```
binier, line 27: Still there are impossible states that could be avoided by a different design.

E.g. `pending_evaluation` could be `Some(_)` when the `status` isn't `BlockProducerVrfEvaluatorStatus::SlotEvaluationPending`.
Same for other fields.

I know it's a hustle to include those fields in enum variants and carry them across variants which does include that field, but it makes impossible states not even representable. Also in general it does lead to more restricted/simpler state machine.

node/src/block_producer/vrf_evaluator/block_producer_vrf_evaluator_state.rs
```diff
@@ -101,51 +382,555 @@ impl EpochData {
...
+    /// The service returned the evaluation succesfully
+    EpochEvaluationSuccess {
+        time: redux::Timestamp,
+        epoch_number: u32,
+    },
+    /// Waiting for the next possible epoch evaluation
+    WaitingForNextEvaluation {
+        time: redux::Timestamp,
+        current_epoch_number: u32,
+        current_best_tip_height: u32,
```
binier, line 472: In general, I wouldn't duplicate data for best tip or anything that is already stored in state machine. Only times when we do need to duplicate the data, is for example in this case if we need to react to best tip change in a way that we need to know what data from old best tip was and what is the data from new best tip.

In this case, if we need to know those previous values from the best tip in order to decide whether we need to keep "waiting for next evaluation", or to move on. If old values are irrelevant and we can decide just based on the new best tip, then this duplication should be removed. I haven't checked which is the case here.

binier (approved):
@adonagy  I only checked and commented on the design. It's definitely an improvement. I did make some suggestions to keep in mind going forward. We could discuss it more if you'd like.

Overall looks good!

PR #294, opened 2024-03-22
Change minimum Rust toolchain to 1.77

sebastiencs:
Use Rust 1.77, to be able to use the now stable [array::each_ref](https://doc.rust-lang.org/stable/std/primitive.array.html#method.each_ref) instead of `array::from_fn`.

`each_ref` guarantee that the source and destination have the same length.
That will protect us from length mismatch in future refactoring, or when binprot types are updated


tizoc (approved):
LGTM. When this gets merged to `develop` please notify everybody about the new minimum required version.

dkuehr:
Looks much better than the previous approach.
Btw, did you notice any speedup? (I suspect not much since in many cases the array elements are the same size as a reference)

dkuehr (approved):

sebastiencs:
> Btw, did you notice any speedup? (I suspect not much since in many cases the array elements are the same size as a reference)

I didn't see any changes on performances. The arrays we're working with are probably not big enough

PR #301, opened 2024-03-26
Fix berkeleynet bootstrap for new libp2p

akoptelov:
- mio errors are propagated to peer management system, so reconnection works now
- better integration of RPCs with new libp2p
- new-style actions for new libp2p
- fix for p2p state JSON serialization


tizoc:
@akoptelov all the "refactor" commits are the beginning are just switching to new style actions right? or is there something else to review there?

akoptelov:
> @akoptelov all the "refactor" commits are the beginning are just switching to new style actions right? or is there something else to review there?

Yes, they are. Basically I'd like @vlad9486 to take a look at them. There is just a single place where I changed the control flow to better match the new style.

PR #302, opened 2024-03-26
Fix clippy errors and warnings

adonagy:


ledger/src/proofs/transaction.rs
```diff
@@ -559,50 +559,15 @@ impl PlonkVerificationKeyEvals<Fp> {
...
+            psm: cvar(psm),
+            complete_add: cvar(complete_add),
+            mul: cvar(mul),
+            emul: cvar(emul),
+            endomul_scalar: cvar(endomul_scalar),
         }
     }
 
-    /// For debugging
-    fn to_string(&self) -> String {
```
sebastiencs, line 572: This has been moved into a `impl std::fmt::Display`, could you just rename `to_string` to `to_dbg_string` instead ?
And rename `from_string` just below to `from_dbg_string`

tizoc:
Replaced by #434

PR #309, opened 2024-03-28
fix(bootstrap): TF sync + RPC fixes

akoptelov:
- Add best_tip to TF chain
- Fix pending RPC handling for peer disconnection
- Better RPC errors handling


tizoc (approved):
LGTM (at least the parts for which I have enough context to understand)

PR #311, opened 2024-03-29
fix(libp2p): remove panic from RPC query

akoptelov:
+ refactor RPC query/response effects. Re: #300


akoptelov:
@tizoc this is mostly refactoring, but anyway let's include into the release as well.

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
```diff
@@ -73,8 +76,7 @@ impl P2pNetworkRpcState {
             }
             P2pNetworkRpcAction::OutgoingQuery { query, .. } => {
                 self.last_id = query.id;
-                // TODO: remove when query is done
```
tizoc, line 76: @akoptelov do you know what this TODO was about?

tizoc (approved):
LGTM

PR #322, opened 2024-04-03
Create a separate service to manage a LedgerCtx in separate thread.

Sventimir:
This PR adds `ledger_manager` module, which defines `LedgerManager`, a service which launches a thread keeping track of a `LedgetCtx` and handling requests querying its state and making changes. All operations are performed asyncronously in a thread separate from the caller's, but syncronous wrappers for these calls are provided to avoid dramatic changes to ledger service API yet.

This new service is created along with `NodeService` instance and the latter has a handle on the former, so it can redirect incoming requests to the new service. This way we avoid big changes to `NodeService`'s API at this point. Once this is merged, we can consider updating call sites to asynchronous workflow on case-by-case basis.

### DIFF
--- a/node/src/ledger/ledger_manager.rs
+++ b/node/src/ledger/ledger_manager.rs
@@ -1,0 +1,144 @@
+use mina_p2p_messages::v2::{
+    LedgerHash, MinaBaseAccountBinableArgStableV2, MinaBaseSparseLedgerBaseStableV2,
+    MinaLedgerSyncLedgerAnswerStableV2, MinaLedgerSyncLedgerQueryStableV1,
+    MinaStateProtocolStateValueStableV2, NonZeroCurvePoint, StateHash,
+};
+use openmina_core::block::ArcBlockWithHash;
+use openmina_core::channels::mpsc;
+use openmina_core::snark::{Snark, SnarkJobId};
+use std::collections::{BTreeMap, BTreeSet};
+use std::sync::mpsc::{channel, RecvError, Sender};
+use std::sync::Arc;
+use std::thread;
+
+use super::ledger_service::LedgerCtx;
+use crate::block_producer::vrf_evaluator::{
+    BlockProducerVrfEvaluatorLedgerService, DelegatorTable,
+};
+use crate::block_producer::{
+    BlockProducerLedgerService, BlockProducerWonSlot, StagedLedgerDiffCreateOutput,
+};
+use crate::ledger::LedgerAddress;
+use crate::p2p::channels::rpc::StagedLedgerAuxAndPendingCoinbases;
+use crate::rpc::{RpcLedgerService, RpcScanStateSummaryScanStateJob};
+use crate::transition_frontier::sync::{
+    ledger::snarked::TransitionFrontierSyncLedgerSnarkedService,
+    ledger::staged::{
+        StagedLedgerAuxAndPendingCoinbasesValid, TransitionFrontierSyncLedgerStagedService,
+    },
+    TransitionFrontierRootSnarkedLedgerUpdates,
+};
+use crate::transition_frontier::{CommitResult, TransitionFrontierService};
+use ledger::Mask;
+use mina_signer::CompressedPubKey;
+use openmina_node_account::AccountPublicKey;
+
+/// The type enumerating different requests that can be made to the
+/// service. Each specific constructor has a specific response
+/// constructor associated with it. Unfortunately, this relationship
+/// can't be expressed in the Rust type system at the moment. For this
+/// reason this type is private while functions wrapping the whole call
+/// to the service are exposed as the service's methods.
+enum LedgerRequest {
+    AccountsSet {
+        snarked_ledger_hash: LedgerHash,
+        parent: LedgerAddress,
+        accounts: Vec<MinaBaseAccountBinableArgStableV2>,
+    }, // expected response: LedgerHash
+    BlockApply {
+        block: ArcBlockWithHash,
+        pred_block: ArcBlockWithHash,
+    }, // expected response: Success
+    ChildHashesGet {
+        snarked_ledger_hash: LedgerHash,
+        parent: LedgerAddress,
+    }, // expected response: ChildHashes
+    Commit {
+        ledgers_to_keep: BTreeSet<LedgerHash>,
+        root_snarked_ledger_updates: TransitionFrontierRootSnarkedLedgerUpdates,
+        needed_protocol_states: BTreeMap<StateHash, MinaStateProtocolStateValueStableV2>,
+        new_root: ArcBlockWithHash,
+        new_best_tip: ArcBlockWithHash,
+    }, // expected response: CommitResult
+    ComputeSnarkedLedgerHashes {
+        snarked_ledger_hash: LedgerHash,
+    }, // expected response: Success
+    CopySnarkedLedgerContentsForSync {
+        origin_snarked_ledger_hash: LedgerHash,
+        target_snarked_ledger_hash: LedgerHash,
+        overwrite: bool,
+    }, // expected response: SnarkedLedgerContentsCopied
+    GetProducerAndDelegates {
+        ledger_hash: LedgerHash,
+        producer: AccountPublicKey,
+    }, // expected response: ProducerAndDelegates
+    GetProducersWithDelegates {
+        ledger_hash: LedgerHash,
+        filter: fn(&CompressedPubKey) -> bool,
+    }, // expected response: ProducersWithDelegatesMap
+    GetMask {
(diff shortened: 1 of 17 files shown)


node/src/ledger/ledger_manager.rs
```diff
@@ -0,0 +1,515 @@
...
+use crate::ledger::LedgerAddress;
+use crate::p2p::channels::rpc::StagedLedgerAuxAndPendingCoinbases;
+use crate::rpc::RpcScanStateSummaryScanStateJob;
+use crate::transition_frontier::sync::{
+    ledger::staged::StagedLedgerAuxAndPendingCoinbasesValid,
+    TransitionFrontierRootSnarkedLedgerUpdates,
+};
+use crate::transition_frontier::CommitResult;
+
+/* The type enumerating different requests that can be made to the
```
tizoc, line 25: @Sventimir Just a quick comment, here you can use `///` to make this a [doc comment](https://doc.rust-lang.org/rust-by-example/meta/doc.html) on the enum.

Sventimir: That's good to know, thanks. Done.

node/src/ledger/ledger_manager.rs
```diff
@@ -0,0 +1,515 @@
...
+
+    fn call(&self, request: LedgerRequest) -> Result<LedgerResponse, String> {
+        let (responder, receiver) = oneshot::channel();
+        self.sender
+            .send(LedgerRequestWithChan { request, responder })
+            .unwrap();
+        result_join(receiver.blocking_recv())
+    }
+
+    pub fn accounts_set(
```
tizoc, line 262: Is there any advantage on wrapping the ledger address `Arc<Mutex<...>>` instead of just cloning it and passing it by-value?

node/src/ledger/ledger_manager.rs
```diff
@@ -0,0 +1,515 @@
...
+/// The type enumerating different requests that can be made to the
+/// service. Each specific constructor has a specific response
+/// constructor associated with it. Unfortunately, this relationship
+/// can't be expressed in the Rust type system at the moment. For this
+/// reason this type is private while functions wrapping the whole call
+/// to the service are exposed as the service's methods.
+enum LedgerRequest {
+    AccountsSet {
+        snarked_ledger_hash: LedgerHash,
+        parent: Arc<Mutex<LedgerAddress>>,
```
tizoc, line 34: Lets just use a LedgerAddress here, there is no advantage in wrapping it in `Arc` and `Mutex` right? it is a cheap value to copy, no need to share it.

node/src/ledger/ledger_manager.rs
```diff
@@ -0,0 +1,515 @@
...
+    }, // expected response: Success
+    ChildHashesGet {
+        snarked_ledger_hash: LedgerHash,
+        parent: LedgerAddress,
+    }, // expected response: ChildHashes
+    Commit {
+        ledgers_to_keep: BTreeSet<LedgerHash>,
+        root_snarked_ledger_updates: TransitionFrontierRootSnarkedLedgerUpdates,
+        needed_protocol_states: BTreeMap<StateHash, MinaStateProtocolStateValueStableV2>,
+        new_root: Arc<Mutex<ArcBlockWithHash>>,
```
tizoc, line 49: A general comment about all these `Arc<Mutex<...>>` values. Given that the original values are not already wrapped in `Arc<Mutex<...>>` you will have to clone their anyway, so I don't see any advantage in this extra wrapping here.

node/src/ledger/ledger_manager.rs
```diff
@@ -0,0 +1,515 @@
...
+        new_best_tip: Arc<Mutex<ArcBlockWithHash>>,
+    }, // expected response: CommitResult
+    ComputeSnarkedLedgerHashes {
+        snarked_ledger_hash: Arc<Mutex<LedgerHash>>,
```
tizoc, line 53: Same as with addresses I would just clone the hash and pass it by value here (you will have to do it anyway to be able to wrap it)

tizoc (changes requested):
@Sventimir I think in general it goes in the right direction, but I don't want to merge a partial implementation. Making use of the `LedgerManager` thread even if the API exposed to the service still blocks should be part of this PR so that we can test it.
You will find that once you start using it you may have to start making changes to the types (I left a few comments related to that), and that some assumptions don't hold, requiring further changes.

Lets wait for @binier's feedback too.

node/src/ledger/ledger_manager.rs
Sventimir: Not sure. I changed all the references in the old API to `Arc<Mutex<...>>`, because it was a simplest solution. Copying is sometimes costly and sometimes not possible at all. Admittedly I didn't checks these particular examples. I assumed that they were passed by reference previously for a reason and given that these `Arc`s aren't meant to live long (at most until the request is handled), I thought it the safest choice. I didn't consider the fact that wrapping in `Arc<Mutex<...>>` might require substantial changes to where these functions are called. I will review these refs and clone wherever possible.

Sventimir:
@tizoc , thanks. I removed the shared references from the API. I will continue in this branch to actually use the code then.

node/src/ledger/ledger_manager.rs
```diff
@@ -0,0 +1,627 @@
...
+        snarked_ledger_hash: &LedgerHash,
+    ) -> Result<(), String> {
+        self.call(LedgerRequest::ComputeSnarkedLedgerHashes {
+            snarked_ledger_hash: snarked_ledger_hash.clone(),
+        })
+        .and_then(|res| {
+            if let LedgerResponse::Success = res {
+                Ok(())
+            } else {
+                Err("LedgerManager::compute_snarked_ledger_hashes: unexpected response".to_string())
```
tizoc, line 359: lets include the response in all these error results (should never happen, but it if does it will be easier to figure out what is going on):

```suggestion
                Err(format!("LedgerManager::compute_snarked_ledger_hashes: unexpected response: {res:?}"))
```

node/src/ledger/ledger_manager.rs
```diff
@@ -0,0 +1,627 @@
...
+pub struct LedgerManager {
+    sender: mpsc::UnboundedSender<LedgerRequestWithChan>,
+    runtime: thread::JoinHandle<LedgerCtx>,
+}
+
+impl LedgerManager {
+    pub fn new(ledger_ctx: LedgerCtx) -> LedgerManager {
+        let (sender, receiver) = mpsc::unbounded_channel();
+        let runtime = thread::spawn(move || {
+            let mut ledger_ctx = ledger_ctx;
```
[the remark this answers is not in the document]
tizoc: Nitpick, but lets rename `runtime` to `join_handle`. Also, when I see a call to a `new` function I don't usually expect for a new thread to be created, so lets rename it to something like `LedgerManager::spawn` instead.

tizoc (changes requested):
@Sventimir left a few comments, just small changes, nothing big.

node/src/ledger/ledger_manager.rs
```diff
@@ -0,0 +1,627 @@
...
+        let (responder, receiver) = channel();
+        self.sender
+            .send(LedgerRequestWithChan { request, responder })
+            .unwrap();
+        result_join(receiver.recv())
+    }
+
+    pub async fn wait_for_stop(
+        self,
+    ) -> Result<LedgerCtx, Box<dyn std::any::Any + std::marker::Send + 'static>> {
```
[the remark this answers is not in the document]
Sventimir: Yes this function is indeed unused at the moment. And yes, the thread ends in case the handle gets dropped, which at the moment is when the whole `NodeService` is dropped. I added the function for completeness of the API and also in case someone needed to stop the manager explicitly at some point and retrieve the LedgerCtx. Admittedly, I cannot imagine a scenario in which that would be better than addding the necessary functionality to the manager, but then again I don't know the system that well yet. In any case, I can remove it if you prefer.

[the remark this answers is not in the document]
tizoc: @Sventimir leave it there for now. At some point we will have to join that thread because it may be in the middle of some I/O operations when the node gets shut down that we may want to wait for (specially if we implement a persisted version of the ledger, scan state, etc to lower memory usage)

node/src/ledger/ledger_manager.rs
Sventimir: Done.

node/src/ledger/ledger_manager.rs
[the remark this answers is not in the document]
Sventimir: Okay, leaving it there then.

tizoc (approved):

PR #342, opened 2024-04-09
fix(transition-frontier): Ensure that by the time the staged ledger reconstruct is complete, the target has not changed

### LINKED ISSUE #331: There is a race condition during async staging ledger reconstruction if the best tip changes.
This will result in the following error:

```
2024-04-05T10:33:13.042885Z  INFO node::logger::logger_effects: time: "1712313193042841120", node_id: "2acaa5JaGrgYBQgfb5dMCfgjdAGk1Hws53atKdwRtALaHMAJcAW", kind: "TransitionFrontierSyncLedgerRootSuccess", summary: "Transition frontier root ledger sync success"
thread 'main' panicked at /home/alexander/openmina/develop/node/src/transition_frontier/sync/transition_frontier_sync_effects.rs:225:62:
called `Result::unwrap()` on an `Err` value: "parent staged ledger missing: jwM1TDLp2qYXruuHwDwb2qvxBGLWk4sqtRfjoN8A964JgpKYUQr"
```

During the staged ledger reconstruction we may connect to new peers and update the best tip target as a result, which also may result in a root ledger change. So by the time the blocks begin to get applied the root has changed and the staged ledger needed by the first successor block will be missing.

We need to make sure that by the time the snarked ledger reconstruct steps finishes, the ledger hashes still match:

https://github.com/openmina/openmina/blob/a6e79936b4fdbdc30030348d0e1da852855fabc2/node/src/transition_frontier/sync/ledger/staged/transition_frontier_sync_ledger_staged_actions.rs#L195

tizoc:
(fixes #331)

### DIFF
--- a/node/src/transition_frontier/sync/ledger/staged/transition_frontier_sync_ledger_staged_actions.rs
+++ b/node/src/transition_frontier/sync/ledger/staged/transition_frontier_sync_ledger_staged_actions.rs
@@ -194,16 +194,21 @@ impl redux::EnablingCondition<crate::State> for TransitionFrontierSyncLedgerStag
                         TransitionFrontierSyncLedgerStagedState::ReconstructPending { .. }
                     )
                 }),
-            TransitionFrontierSyncLedgerStagedAction::ReconstructSuccess { .. } => state
+            TransitionFrontierSyncLedgerStagedAction::ReconstructSuccess { ledger_hash } => state
                 .transition_frontier
                 .sync
                 .ledger()
                 .and_then(|s| s.staged())
                 .map_or(false, |s| {
+                    // Assumption here is that if the hash doesn't match, it is because the reconstruct
+                    // is stale (best tip changed while reconstruction was happening). The staging
+                    // ledger reconstruction logic itself will already validate that the resulting
+                    // reconstructed ledger matches the expected hash.
+                    let expected_hash = &s.target().staged.staged_ledger_hash;
                     matches!(
                         s,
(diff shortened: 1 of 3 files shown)


node/src/transition_frontier/sync/ledger/staged/transition_frontier_sync_ledger_staged_actions.rs
```diff
@@ -194,16 +194,22 @@ impl redux::EnablingCondition<crate::State> for TransitionFrontierSyncLedgerStag
...
                     )
                 }),
-            TransitionFrontierSyncLedgerStagedAction::ReconstructSuccess { .. } => state
+            TransitionFrontierSyncLedgerStagedAction::ReconstructSuccess { ledger_hash } => state
                 .transition_frontier
                 .sync
                 .ledger()
                 .and_then(|s| s.staged())
                 .map_or(false, |s| {
+                    // NOTE(tizoc): Assumption here is that the reconstructed staged ledger will always match,
```
tizoc, line 203: @binier what do you think of this? best option I can think of is to pass the expected hash to the `staged_ledger_reconstruct` call so that the service can validate it (and trigger an error if it fails)

tizoc: Ignore the above, during staged ledger reconstruction (inside the ledger implementation logic) this is validated already, so it will result in an error, no need to add anything extra.

Issue #338, opened 2024-04-08
As a developer, I need to have extensive set of tests for p2p, so I can see regressing while improving/modifying code


akoptelov:
- [x] Fix existing p2p tests
- [ ] Make sure existing p2p tests work with webrtc version 
- [ ] Add discovery tests
- [x] Add rpc tests
- [x] Add identify tests
- [ ] Add pubsub tests
- [ ] TODO

akoptelov:
UPD.
Identify base test is ready, working with RPCs

akoptelov:
UPD. Working with Kademlia tests. Blocker now is #341

PR #351, opened 2024-04-11
Fix existing p2p tests

akoptelov:
re: #338


akoptelov:
@binier and @vlad9486 Please take a look if you want to.

PR #359, opened 2024-04-17
Make ledger reconstruction asynchronous

Sventimir:
Remove @tizoc's temporary solution for making ledger reconstruction asynchronous. Now that ledger operations are handled by a `LedgerManager` in separate thread, it is sufficient to request the ledger reconstruction without blocking and have an event notify the state machine that it's ready.
* Remove `staged_ledger_reconstruct_result_store` function, moving its functionality to `staged_ledger_reconstruct`
* Make calls to `staged_ledger_reconstruct` non-blocking
* Make the `LedgerManager` send out the appropriate event when ledger is reconstructed.


node/src/ledger/ledger_service.rs
```diff
@@ -430,79 +429,49 @@ impl LedgerCtx {
...
-                    mask,
-                    LocalState::empty(),
-                    parts.staged_ledger_hash.0.to_field(),
-                    (&parts.pending_coinbase).into(),
-                    |key| states.get(&key).cloned().unwrap(),
-                )
-            } else {
-                StagedLedger::create_exn(CONSTRAINT_CONSTANTS.clone(), mask)
-            };
+        let result = if let Some(parts) = parts {
```
tizoc, line 444: @Sventimir something that was useful about the previous workaround was that while the staged ledger was being reconstructed (the most expensive ledger operation) other requests could still be served.

Unless it gets complicated it would be good to recover that aspect.

An important detail is that in general, usually when ledgers are being mutated, it is because they are being constructed as a new `Mask` (so whoever is mutating a ledger will have exclusive access to it, because until the ledger construction is complete, nobody will need to read from it). And since `Mask`s can be sent across threads, spawning a new thread to handle the construction task should not be an issue (this is why it worked before).

It is not super important for this to be changed now, because @binier is already solving it in his refactoring (in general, not just for this call) by separating read/construct operations based on what I described in the previous paragraph.

Sventimir: I believe it should now work as you want.

binier:
Closing as similar change is included in this pr: https://github.com/openmina/openmina/pull/362

Issue #361, opened 2024-04-18
Issue connecting to node running on localhost


0xMimir:
When running 2 nodes on same computer, they don't connect one to another. First node has remote peers, while second has only first node as peer.

When following query is ran on graphql interface:
```graphql
query MyQuery {
  syncStatus
}
```

First node gets to `BOOTSTRAP` and later to `SYNCED`, while second node is stuck at `LISTENING`

Steps to reproduce:

In one terminal run:
```sh
export OPENMINA_DISCOVERY_FILTER_ADDR=false
cargo run --bin openmina -- node -s 5KgwoC4tGKNE4G3XP6JDYXKPSMyzjyEfKxdACsnqwTXBaSD75y4
```

In second terminal run:
```sh
export OPENMINA_DISCOVERY_FILTER_ADDR=false
cargo run --bin openmina -- node -P /ip4/127.0.0.1/tcp/8302/p2p/12D3KooWJDihKQfdBUgbRQpMUyUj7avdaJUqin1hJGGQKE8oGF72 --libp2p-port 4001 -p 4903
```

This same issue also happens when node is ran on another computer on local network.

binier:
@akoptelov @vlad9486 Could you take a look?

akoptelov:
@0xMimir thanks for posting this issue.

I can see following problems:
- Connection to the Rust node. The second node constantly disconnects from the first one with the reason `transition frontier RPC timeout`. Looks like we don't handle that properly.
- Peers disovery. With the current implementation, a peer can have only one IP address, while Kademlia provides multiple addresses. The `OPENMINA_DISCOVERY_FILTER_ADDR` is rather a workaround for this issue, as when connecting to a public network we want to filter out local addresses, while in testing we should use localhost ones. In other words, with that variable set to `true` the node will inaccessible addresses for peers outside of the local network. Also I can see some errors regarding Kademlia in the log. The test suite for Kademlia currently is under construction.

@binier FYI

PR #367, opened 2024-04-24
Bugfix for #361

0xMimir:
I have code to handle action being `SendResponse` and state being `Incoming(RequestIsReady)` in kademlia stream effects, this allows for node to respond to incoming requests.

@akoptelov, Could you take a look

### DIFF
--- a/p2p/src/network/kad/stream/p2p_network_kad_stream_effects.rs
+++ b/p2p/src/network/kad/stream/p2p_network_kad_stream_effects.rs
@@ -48,16 +48,17 @@ impl P2pNetworkKademliaStreamAction {
                     data: P2pNetworkKademliaRpcRequest::FindNode { key },
                 }),
             ) => {
-                store.dispatch(P2pNetworkKademliaAction::AnswerFindNodeRequest {
-                    addr,
-                    peer_id,
-                    stream_id,
-                    key: *key,
-                });
+                let key = *key;
                 store.dispatch(A::WaitOutgoing {
                     addr,
                     peer_id,
                     stream_id,
+                });
+                store.dispatch(P2pNetworkKademliaAction::AnswerFindNodeRequest {
+                    addr,
+                    peer_id,
+                    stream_id,
+                    key,
                 });
                 Ok(())
             }
--- a/p2p/tests/kademlia.rs
+++ b/p2p/tests/kademlia.rs
@@ -20,11 +20,15 @@ async fn kademlia() {
         .expect("should build cluster");
 
     let node1 = cluster
-        .add_rust_node(RustNodeConfig::default())
+        .add_rust_node(RustNodeConfig {
+            discovery: true,
+            ..Default::default()
+        })
         .expect("node1");
 
     let config = RustNodeConfig {
         initial_peers: vec![Listener::Rust(node1)],
+        discovery: true,
         ..Default::default()
     };
 
@@ -59,6 +63,7 @@ async fn kademlia() {
     let node3 = cluster
         .add_rust_node(RustNodeConfig {
             initial_peers: vec![Listener::Rust(node2)],
+            discovery: true,
             ..Default::default()
         })
         .expect("Error creating node 3");
--- a/p2p/tests/kademlia.rs
+++ b/p2p/tests/kademlia.rs
@@ -1,0 +1,171 @@
+use p2p::P2pNetworkKadBucket;
+use p2p_testing::{
+    cluster::{Cluster, ClusterBuilder, Listener},
+    futures::TryStreamExt,
+    predicates::kad_finished_bootstrap,
+    rust_node::{RustNodeConfig, RustNodeId},
+    stream::ClusterStreamExt,
+};
+use std::time::Duration;
+
+#[tokio::test]
+async fn kademlia() {
+    std::env::set_var("OPENMINA_DISCOVERY_FILTER_ADDR", "false");
+
+    let mut cluster = ClusterBuilder::new()
+        .ports(11000..11200)
+        .idle_duration(Duration::from_millis(100))
+        .start()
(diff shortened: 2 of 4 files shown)


p2p/src/network/kad/stream/p2p_network_kad_stream_effects.rs
```diff
@@ -177,6 +177,31 @@ impl P2pNetworkKademliaStreamAction {
...
+            (
+                A::SendResponse {
+                    addr,
+                    peer_id,
+                    stream_id,
+                    data,
+                },
+                D::Incoming(I::RequestIsReady {
+                    data: P2pNetworkKademliaRpcRequest::FindNode { key },
+                }),
```
akoptelov, line 189: The `P2pNetworkKademliaAction::AnswerFindNodeRequest` action dispatches before `WaitOutgoing`, so the stream ends up in the wrong state.

akoptelov (changes requested):
This fix resolves the issue only partially.

Please take a look at the existing effect code for `IncomingData` action, when state is `RequestIsReady`. It first dispatches `AnswerFindNodeRequest` (to query Kademlia routing table) and then `WaitOutgoing` so reducer changes the state. It looks like the order should be opposite -- we first should expect an answer from our Kademlia routing table, and then ask it. 

Also, regarding the amount of states, incoming kademlia stream should be properly handled. A server (this node in this case) should expect more queries, and close the stream only after the client closes its half.

https://github.com/libp2p/specs/tree/master/kad-dht#rpc-messages

p2p/src/network/kad/stream/p2p_network_kad_stream_effects.rs
0xMimir: Just replacing order between `WaitOutgoing` and `AnswerFindNodeRequest` fixes the issue, I completely overlooked this.

akoptelov: Well, this is not an easy task to understand all the logic from the first time...

akoptelov:
@0xMimir Don't you mind also adding a test on this? There's a brand new p2p framework, I can help you with that.

This is very basic example of a test: https://github.com/openmina/openmina/blob/develop/p2p/tests/basic.rs

Shortly I'll add one for `identify` protocol.

akoptelov:
> Shortly I'll add one for `identify` protocol.

This one:
https://github.com/openmina/openmina/blob/3bcae092f34ee862d16b1cb7c979495eb8438b24/p2p/tests/identify.rs

p2p/tests/kademlia-incoming.rs
```diff
@@ -0,0 +1,78 @@
...
+        .routing_table
+        .buckets
+        .first()
+        .expect("There must be at least 1 bucket");
+
+    assert_eq!(
+        bucket.len(),
+        2,
+        "There must be 2 items in bucket self and seed node"
+    );
```
akoptelov, line 47: Hmm... does it really work for you?
The thing with this testing cluster is that unless you poll its stream, nothing will happen, no connection, no kad messages exchange.

akoptelov (changes requested):
@0xMimir In order to run the cluster, you should poll its events by using some stream operations. Without it statemachines of the nodes inside it won't progress.

This might look tool low-level, but in this way you have the full control over the nodes, where to continue or stop the execution, and inject an action at the right time.

Please take a look at the `basic.rs` example, there's a code that waits for node to listen, then connects another node to it, and waits for it to be a ready peer, and so on.

p2p/tests/kademlia-incoming.rs
0xMimir: Test does work, and first node is in seconds node routing table, but I am not sure how, since `effects` from `P2pNetworkKadAction` never get called.

akoptelov: Ah, that explains things. Indeed, Kademlia table is populated with initial peers on creation. Well, that makes sense to test as well, but more complex case where routing is updated during Kademlia bootstrap is more important,

akoptelov:
@0xMimir 
Also, can you put all kademlia tests into a single file?

akoptelov (changes requested):
Looks good, but I want to ask you about the last change. I've changed the default node configuration for p2p tests so discovery is turned off. Please enable it explicitly when needed.

akoptelov:
@0xMimir you will need to rebase on top of the latest `develop`

p2p/tests/kademlia.rs
```diff
@@ -0,0 +1,171 @@
...
+    assert!(seed_peer.is_some(), "Node1 peer not found");
+}
+
+#[tokio::test]
+#[ignore]
+async fn kademlia_incoming() {
+    std::env::set_var("OPENMINA_DISCOVERY_FILTER_ADDR", "false");
+
+    let mut cluster = ClusterBuilder::new()
+        .ports(11000..11200)
```
akoptelov, line 100: Tests from a single test file are executed in different threads, and we should make sure they use disjoint set of ports, otherwise they will fail randomly.

I need to implement a shared storage for port ranges...

For now this is fine as the test is ignored.

akoptelov (approved):
LGTM, with a comment for a future.

akoptelov:
UPD. I see, there is no discovery enabled in the test yet...

0xMimir:
After rebase test started failing, but when testing if second node gets into `BOOTSTRAP` state and first node responds with kademlia, everything works, I am currently debugging why the test is failing

0xMimir:
> UPD. I see, there is no discovery enabled in the test yet...

I enabled it in last commit

0xMimir:
> LGTM

I don't have permission to merge pull requests, so can you do it.

PR #369, opened 2024-04-25
Load deamon.json configuration

Sventimir:
This PR creates the possibility to load genesis ledger and protocol constants from `daemon.json` file that is used by OCaml nodes. Not everything is currently read from that file; only the values that we might want to use at the moment. The format can be extended to encompass all possible options understood by OCaml nodes if required.


node/src/daemon_json/json_genesis.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    pub fn slots_per_sub_window(&self) -> UnsignedExtendedUInt32StableV1 {
+        UnsignedExtendedUInt32StableV1(Number(self.slots_per_sub_window))
+    }
+
+    pub fn grace_period_slots(&self) -> UnsignedExtendedUInt32StableV1 {
+        UnsignedExtendedUInt32StableV1(Number(self.grace_period_slots))
+    }
+
+    pub fn genesis_state_timestamp(&self) -> anyhow::Result<BlockTimeTimeStableV1> {
```
akoptelov, line 43: I wouldn't use `anyhow` based errors in library crates.

node/src/daemon_json/json_genesis.rs
```diff
@@ -0,0 +1,66 @@
...
+                BlockTimeTimeStableV1(UnsignedExtendedUInt64Int64ForVersionTagsStableV1(Number(
+                    dt.timestamp() as u64,
+                )))
+            })
+            .map_err(|e| anyhow::anyhow!(e))
+    }
+
+    pub fn to_genesis_constants_checked(
+        &self,
+    ) -> anyhow::Result<MinaBaseProtocolConstantsCheckedValueStableV1> {
```
akoptelov, line 55: See above.

akoptelov (changes requested):
I think there should be a special error type instead of `anyhow`-based one.

And also there are some warnings from clippy that might need fixing.

Sventimir:
> And what about specifying daemon.json from node's command line?
@akoptelov, I fear this won't work just by itself. For a custom network you either need to operate in seed node (which I believe Openmina cannot do at the moment) or specify a custom peer list (which I don't know if is possible). I can implement the option switch, but it won;t be very useful unless we can also specify a custom peer list.

akoptelov:
@Sventimir different set of seeds is the least of the issues:
```
cargo run --release -p cli node --peers \
  /ip4/34.48.73.58/tcp/10003/p2p/12D3KooWAdgYL6hv18M3iDBdaK1dRygPivSfAfBNDzie6YqydVbs \
  /ip4/35.245.82.250/tcp/10003/p2p/12D3KooWLjs54xHzVmMmGYb7W5RVibqbwD1co7M2ZMfPgPm7iAag \
  /ip4/35.245.106.190/tcp/10003/p2p/12D3KooWEiGVAFC7curXWXiGZyMWnZK9h8BKr88U8D5PKV3dXciv
```
Another thing is to specify a proper chain id. We don't have that option anymore, the idea is to derive it from constants/genesis, but we can revert that at least temporarily.

[0001-chain_id.patch.gz](https://github.com/openmina/openmina/files/15127784/0001-chain_id.patch.gz)

Sventimir:
@akoptelov, chain id can and indeed should be derrived from genesis parameters. Okay, I didn't realize we can pass a peer list. In that case I'll add the option to load a config at startup.

node/src/daemon_json/json_genesis.rs
Sventimir: Done in d3d4829f.

node/src/daemon_json/json_genesis.rs
Sventimir: Done in: d3d4829f.

Sventimir:
I added a (mandatory) option to specify a config file. Unfortuantely,. when supplied wityh the berkeley JSON config, the node crashes:
```
thread 'main' panicked at /home/sven/work/openmina/node/src/transition_frontier/sync/transition_frontier_sync_effects.rs:50:62:
called `Result::unwrap()` on an `Err` value: "Tried to copy from non-existing snarked ledger with hash: jwkqwgAC6MXgfiZmynHRqXV6PGbMbLwFCx56Y2rt5vwdumf6ofp"
```

tizoc:
The issue for that bug is here https://github.com/openmina/openmina/issues/331, if you launch the node again it may work this time.

In theory it should have been fixed by https://github.com/openmina/openmina/pull/342 but it seems thats not the case

tizoc:
@Sventimir actually, your seems to be different, the problem is that the genesis ledger is not there

node/src/transition_frontier/genesis/transition_frontier_genesis_config.rs
```diff
@@ -130,6 +135,44 @@ impl GenesisConfig {
...
+                    .ok_or(anyhow::anyhow!("No accounts in config"))?;
+                let accounts: Vec<Account> = account_list
+                    .iter()
+                    .map(daemon_json::Account::to_account)
+                    .collect::<Result<Vec<Account>, AccountConfigError>>()
+                    .map_err(anyhow::Error::from)?;
+                let (mut mask, total_currency) = Self::build_ledger_from_accounts(accounts);
+                let result = GenesisConfigLoaded {
+                    constants,
+                    ledger_hash: ledger_hash(&mut mask),
```
tizoc, line 170: @Sventimir which hash are you getting here when using the daemon.json file? issue could be that you are not inserting the winner account (check the ledger tool and how it generates it)

https://github.com/openmina/openmina/blob/develop/tools/ledger-tool/src/main.rs#L128-L138

Sventimir: I added the genesis winner account in https://github.com/openmina/openmina/pull/369/commits/629b87e8eaa99414634b19ab9311ba837b46b4cc, but it does not seem to help much. I still get the same error.

tizoc: @Sventimir try adding the genesis account first, then the rest

Sventimir: Did that, didn't help either. The ledger hash I get is different than expected: `jxA9yKKm77xdUbCBp59bdaTVRD9MqzX8HorKyg5sghFQeckUxnL`.

Sventimir:
Rebased and fixed problems with loading network config from file. The node does sync again.

tools/ledger-tool/src/main.rs
```diff
@@ -148,6 +148,7 @@ fn main() -> anyhow::Result<()> {
     }
 
     for account in &accounts {
+        println!("{account:?}");
```
tizoc, line 151: leftover print

tizoc (approved):
@Sventimir LGTM, please squash and cleanup the commits in any way you think makes sense and merge once you are happy with it

cli/src/commands/node/mod.rs
```diff
@@ -158,9 +166,17 @@ impl Node {
             )
         });
 
+        let config_file = self.config.map(File::open).transpose()?;
```
akoptelov, line 169: If incorrect file will be specified, something like `[ERROR] No such file or directory (os error 2)`, with no clue which file is not found. I would use `anyhow::Context` here, but for some reason we're using boxed Error in `cli`, so probably an `error` log entry is worth adding here.

cli/src/commands/node/mod.rs
```diff
@@ -158,9 +166,17 @@ impl Node {
...
+        let conf: Option<DaemonJson> = config_file.map(serde_json::from_reader).transpose()?;
```
akoptelov, line 170: The same as above.

akoptelov (changes requested):
LGTM, with two minor fixes to add.

cli/src/commands/node/mod.rs
Sventimir: Does this look better:
```
2024-05-09T09:19:20.509149Z ERROR openmina::commands::node: time: "1715246360507319345", kind: "ConfigFileError", summary: "failed to open config file", error: "Os { code: 2, kind: NotFound, message: \"No such file or directory\" }"
[ERROR] No such file or directory (os error 2)
```
?

cli/src/commands/node/mod.rs
Sventimir: Like this:
```
2024-05-09T09:22:07.440473Z ERROR openmina::commands::node: time: "1715246527440454455", kind: "ConfigFileError", summary: "failed to parse config file", error: "Error(\"expected value\", line: 1, column: 1)"
[ERROR] expected value at line 1 column 1
```
?

akoptelov (approved):

PR #390, opened 2024-05-09
fix(node): Fix zero value for the empty coinbase stack data hash

tizoc:
This was causing the genesis protocol state hash to be computed incorrectly.

### DIFF
--- a/mina-p2p-messages/src/v2/manual.rs
+++ b/mina-p2p-messages/src/v2/manual.rs
@@ -257,7 +257,10 @@ impl EpochSeed {
 
 impl CoinbaseStackData {
     pub fn zero() -> Self {
-        MinaBasePendingCoinbaseCoinbaseStackStableV1(BigInt::zero()).into()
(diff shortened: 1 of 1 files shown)


mina-p2p-messages/src/v2/manual.rs
```diff
@@ -257,7 +257,10 @@ impl EpochSeed {
 
 impl CoinbaseStackData {
     pub fn zero() -> Self {
```
sebastiencs, line 259: @tizoc That method should probably be renamed to `empty` ? To make it clear that it's value is not zero

tizoc: good point

tizoc: (and same for the callers which are currently named `zero` too)

tizoc: https://github.com/openmina/openmina/pull/391

PR #392, opened 2024-05-13
Implemented basic pubsub and tests

vlad9486:


node/testing/src/cluster/mod.rs
```diff
@@ -388,10 +389,11 @@ impl Cluster {
                 // TODO(binier): record instead of panicing.
                 match res {
                     InvariantResult::Violation(violation) => {
-                        panic!(
```
vlad9486, line 391: @binier I have to disable it, because it crashes.

binier: Just cherry pick this commit and try again. It shouldn't fail.

`6c1cb95941163908880c146823e94c1d1a45f539`

cli/src/commands/node/mod.rs
```diff
@@ -193,6 +194,9 @@ impl Node {
                 timeouts: P2pTimeouts::default(),
                 chain_id: CHAIN_ID.to_owned(),
                 peer_discovery: !self.no_peers_discovery,
+                initial_time: SystemTime::now()
```
binier, line 197: Why do we have duplicate state here? Maybe it can be avoided? Also use `redux::Timestamp` everywhere for time, never `SystemTime`.

binier: If it's used just for pubsub nonce, then please change the name.

node/testing/src/cluster/mod.rs
binier: Also if an invariant is failing, we can't just disable it. We need to investigate and fix either the invariant or the node. I understand you were working on p2p and invariant that was failing could have been unrelated, but instead of disabling it, please create an issue and ask on the channel for someone to fix it.

PR #408, opened 2024-05-16
Compute chain id from network configuration at startup.

Sventimir:
Up until now chain id was hard coded in source and so the network could not access different networks without modification and recompilation. With this change the chain id will now be computed at startup. Node still won't be reconfigurable to other networks without recopmpilation, because we can't compute some of the input and still had to hard-code them, but modifying them makes it easier to adapt the code for other networks (it boils down to changing some constants).

Solve: https://github.com/openmina/openmina/issues/397

### DIFF
--- a/cli/src/commands/node/mod.rs
+++ b/cli/src/commands/node/mod.rs
@@ -8,10 +8,10 @@ use std::time::Duration;
 use libp2p_identity::Keypair;
 use mina_p2p_messages::v2::{
     CurrencyFeeStableV1, NonZeroCurvePoint, NonZeroCurvePointUncompressedStableV1,
-    UnsignedExtendedUInt64Int64ForVersionTagsStableV1,
+    UnsignedExtendedUInt32StableV1, UnsignedExtendedUInt64Int64ForVersionTagsStableV1,
 };
 use node::transition_frontier::genesis::GenesisConfig;
-use openmina_core::ChainId;
+use openmina_core::{constants, ChainId};
 use rand::prelude::*;
 
 use redux::SystemTime;
@@ -38,10 +38,6 @@ use node::{
 
 use openmina_node_native::rpc::RpcService;
 use openmina_node_native::{http_server, tracing, NodeService, P2pTaskSpawner, RpcSender};
-
-// old:
-// 3c41383994b87449625df91769dff7b507825c064287d30fada9286f3f1cb15e
-const CHAIN_ID: ChainId = openmina_core::CHAIN_ID;
 
 /// Openmina node
 #[derive(Debug, clap::Args)]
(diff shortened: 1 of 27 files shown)


cli/src/commands/node/mod.rs
```diff
@@ -188,10 +185,20 @@ impl Node {
...
-            None => TransitionFrontierConfig::new(node::config::BERKELEY_CONFIG.clone()),
+        let genesis_config = match conf {
+            Some(c) => Arc::new(GenesisConfig::DaemonJson(c)),
+            None => node::config::BERKELEY_CONFIG.clone(),
         };
+        let transition_frontier = TransitionFrontierConfig::new(genesis_config.clone());
+        let protocol_constants = genesis_config.protocol_constants()?;
+        let chain_id = ChainId::compute(
+            constants::CONSTRAINT_SYSTEM_DIGESTS.as_slice(),
+            &constants::GENESIS_STATE_HASH,
```
akoptelov, line 196: Just to make sure... Is it constant in OCaml?

core/src/chain_id.rs
```diff
@@ -0,0 +1,185 @@
...
+impl AsRef<[u8]> for ChainId {
+    fn as_ref(&self) -> &[u8] {
+        &self.0
+    }
+}
+
+impl Display for ChainId {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        for b in self.0 {
+            write!(f, "{:02x}", b)?;
```
akoptelov, line 122: Why not `hex`? BTW, there are `derive_more` and `serde_as` that can derive such representations.

core/src/chain_id.rs
```diff
@@ -0,0 +1,185 @@
...
+        }
+        Ok(())
+    }
+}
+
+impl Debug for ChainId {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        write!(f, "ChainId({})", self)
```
akoptelov, line 130: I would suggest default Debug. Why omit it?

akoptelov (approved):
LGTM, with a question about genesis block hash and few suggestions how to reduce source code.

core/src/chain_id.rs
Sventimir: I got hex printed out in the logs with this representation.

core/src/chain_id.rs
Sventimir: I guess hex will be more readable than something the compiler might derive for a `u8` array,

cli/src/commands/node/mod.rs
Sventimir: It's a part of network configuration. As far as I could see, it's contrlled by a dune profile in OCaml.

core/src/chain_id.rs
```diff
@@ -0,0 +1,185 @@
...
+    };
+    use time::format_description::well_known::Rfc3339;
+
+    #[test]
+    fn test_berkeley_chain_id() {
+        // Compute the chain id for the Berkeley network and compare it the real one.
+        let genesis_state_timestamp = OffsetDateTime::parse("2024-02-02T14:01:01Z", &Rfc3339)
+            .unwrap()
+            .unix_timestamp_nanos();
+        let genesis_constants = MinaBaseProtocolConstantsCheckedValueStableV1 {
```
tizoc, line 154: lint check failure here because the variable is unused

core/src/chain_id.rs
akoptelov: I mean, why don't use `hex::encode()`?

Sventimir: Ah, the package `hex`. Yes, I can use it.

Sventimir:
This PR is encompassed by: https://github.com/openmina/openmina/pull/420.

PR #420, opened 2024-05-21
task(p2p): make p2p start on demand

### LINKED ISSUE #418: As a Developer, I need p2p to have delayed initialization, so I can calculate its prerequisites (chain_id) in statemachine
This is needed to make chain_id calculated instead of specifying on node startup.

See #397

akoptelov:
This PR adds possibility to make delayed initialization for p2p part. This is needed for calculating chain_id in the state machine.

fixes #418

### DIFF
--- a/core/src/chain_id.rs
+++ b/core/src/chain_id.rs
@@ -27,7 +27,7 @@ fn hash_genesis_constants(
 ) -> [u8; 32] {
     let mut hasher = Blake2b256::default();
     let genesis_timestamp = OffsetDateTime::from_unix_timestamp_nanos(
-        constants.genesis_state_timestamp.0 .0 .0 as i128,
+        (constants.genesis_state_timestamp.0 .0 .0 * 1000000) as i128,
     )
     .unwrap();
     let time_format =
--- a/core/src/chain_id.rs
+++ b/core/src/chain_id.rs
@@ -1,0 +1,150 @@
+use mina_p2p_messages::v2::{
+    MinaBaseProtocolConstantsCheckedValueStableV1, StateHash, UnsignedExtendedUInt32StableV1,
+};
+use multihash::{Blake2b256, Hasher};
+use time::macros::format_description;
+use time::OffsetDateTime;
+
+use std::fmt::{self, Debug, Display, Formatter};
+
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+#[derive(Clone, PartialEq, Eq)]
+pub struct ChainId([u8; 32]);
+
+fn md5_hash(data: u8) -> String {
+    let mut hasher = md5::Context::new();
+    hasher.consume(data.to_string().as_bytes());
+    let hash: Md5 = *hasher.compute();
+    hex::encode(&hash)
+}
+
+type Md5 = [u8; 16];
+
+fn hash_genesis_constants(
+    constants: &MinaBaseProtocolConstantsCheckedValueStableV1,
+    tx_pool_max_size: &UnsignedExtendedUInt32StableV1,
+) -> [u8; 32] {
+    let mut hasher = Blake2b256::default();
+    let genesis_timestamp = OffsetDateTime::from_unix_timestamp_nanos(
+        constants.genesis_state_timestamp.0 .0 .0 as i128,
+    )
+    .unwrap();
+    let time_format =
+        format_description!("[year]-[month]-[day] [hour]:[minute]:[second].[subsecond digits:6]Z");
+    hasher.update(constants.k.to_string().as_bytes());
+    hasher.update(constants.slots_per_epoch.to_string().as_bytes());
+    hasher.update(constants.slots_per_sub_window.to_string().as_bytes());
+    hasher.update(constants.delta.to_string().as_bytes());
+    hasher.update(tx_pool_max_size.to_string().as_bytes());
+    hasher.update(genesis_timestamp.format(&time_format).unwrap().as_bytes());
+    hasher.finalize().try_into().unwrap()
+}
+
+impl ChainId {
+    pub fn compute(
+        constraint_system_digests: &[Md5],
+        genesis_state_hash: &StateHash,
+        genesis_constants: &MinaBaseProtocolConstantsCheckedValueStableV1,
+        protocol_transaction_version: u8,
+        protocol_network_version: u8,
+        tx_max_pool_size: &UnsignedExtendedUInt32StableV1,
+    ) -> ChainId {
+        let mut hasher = Blake2b256::default();
+        let constraint_system_hash = constraint_system_digests
+            .iter()
+            .map(|md5| hex::encode(md5))
+            .reduce(|acc, el| acc + &el)
+            .unwrap_or_else(|| String::new());
+        let genesis_constants_hash = hash_genesis_constants(genesis_constants, tx_max_pool_size);
+            hasher.update(genesis_state_hash.to_string().as_bytes());
+            hasher.update(constraint_system_hash.to_string().as_bytes());
+            hasher.update(hex::encode(&genesis_constants_hash).as_bytes());
+            hasher.update(md5_hash(protocol_transaction_version).as_bytes());
+            hasher.update(md5_hash(protocol_network_version).as_bytes());
+        ChainId(hasher.finalize().try_into().unwrap())
+    }
+
+    pub fn to_hex(&self) -> String {
+        hex::encode(&self.0)
+    }
+
+    pub fn from_hex(s: &str) -> Result<ChainId, hex::FromHexError> {
+        let h = hex::decode(s)?;
+        let bs = h[..32]
+            .try_into()
+            .or(Err(hex::FromHexError::InvalidStringLength))?;
+        Ok(ChainId(bs))
+    }
+
+    pub fn from_bytes(bytes: &[u8]) -> ChainId {
+        let mut arr = [0u8; 32];
+        arr.copy_from_slice(&bytes[..32]);
+        ChainId(arr)
+    }
+}
+
+impl Serialize for ChainId {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        serializer.serialize_str(&self.to_hex())
+    }
+}
+
+impl<'de> Deserialize<'de> for ChainId {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let s = String::deserialize(deserializer)?;
+        ChainId::from_hex(&s).map_err(serde::de::Error::custom)
+    }
+}
+
+impl AsRef<[u8]> for ChainId {
+    fn as_ref(&self) -> &[u8] {
+        &self.0
+    }
+}
+
+impl Display for ChainId {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        write!(f, "{}", self.to_hex())?;
+        Ok(())
+    }
+}
+
+impl Debug for ChainId {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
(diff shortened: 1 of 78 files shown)


node/src/ledger/read/ledger_read_actions.rs
```diff
@@ -27,7 +27,9 @@ pub enum LedgerReadAction {
 impl redux::EnablingCondition<crate::State> for LedgerReadAction {
     fn is_enabled(&self, state: &crate::State, _time: redux::Timestamp) -> bool {
         match self {
-            LedgerReadAction::FindTodos => state.ledger.read.is_total_cost_under_limit(),
+            LedgerReadAction::FindTodos => {
+                state.p2p.ready().is_some() && state.ledger.read.is_total_cost_under_limit()
```
binier, line 31: Read requests arent just for p2p. Please revert this change.

node/src/state.rs
```diff
@@ -94,3 +96,124 @@ impl State {
...
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum P2pInitializationError {
+    #[error("p2p is already initialized")]
+    AlreadyInitialized,
+}
+
+#[macro_export]
+macro_rules! p2p_ready {
```
binier, line 114: Can't we have a function with the same logic? I'm worried about growing compilation times.

node/testing/src/scenarios/cluster_runner.rs
```diff
@@ -117,15 +118,15 @@ impl<'a> ClusterRunner<'a> {
         )
     }
 
-    pub fn get_chain_id(&self) -> Option<String> {
+    pub fn get_chain_id(&self) -> Option<ChainId> {
```
binier, line 121: Do we still need chain id inside cluster?

binier: Not sure why we had it in the first place. If in test we are interacting with the ocaml node, we will get the chain_id from there, otherwise chainid didnt matter.

node/testing/src/simulator/mod.rs
```diff
@@ -41,9 +42,9 @@ impl Simulator {
         {
             chain_id
         } else if let Some(node) = runner.ocaml_node(ClusterOcamlNodeId::new_unchecked(0)) {
-            node.chain_id_async().await.unwrap().into_bytes()
+            node.chain_id_async().await.unwrap()
         } else {
-            "<unknown_chain_id>".into()
```
binier, line 46: Since chain id calculation will be done inside the node, do we still need this or chain id configuration in rust node at all?

binier (changes requested):

node/src/action.rs
```diff
@@ -53,7 +55,7 @@ impl redux::EnablingCondition<crate::State> for CheckTimeoutsAction {}
 
 #[cfg(feature = "replay")]
 impl redux::EnablingCondition<crate::State> for Action {
-    fn is_enabled(&self, _: &crate::State, _time: redux::Timestamp) -> bool {
-        true
+    fn is_enabled(&self, state: &crate::State, _time: redux::Timestamp) -> bool {
+        !matches!(self, Action::InitializeP2p { .. }) || state.p2p.ready().is_none()
```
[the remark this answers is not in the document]
akoptelov: @binier `InitializeP2p` shouldn't be triggered twice. How can I express that?

node/src/state.rs
akoptelov: This is mostly for sorting out places where an action need p2p but it is not initialized. It turned into `panic` with proper info easily. Also we should bail out if there's no p2p (`return;`). As to compilation time, such macro rules is a drop in the ocean.

node/src/ledger/read/ledger_read_actions.rs
akoptelov: Hmm... ok, there was a place where p2p was kind of essential, i.e. we collected ready peers. But for uninitialized p2p it will be just empty list.

node/testing/src/simulator/mod.rs
akoptelov: Now that's gone.

core/src/chain_id.rs
```diff
@@ -0,0 +1,160 @@
...
+    tx_pool_max_size: &UnsignedExtendedUInt32StableV1,
+) -> [u8; 32] {
+    let mut hasher = Blake2b256::default();
+    let genesis_timestamp = OffsetDateTime::from_unix_timestamp_nanos(
+        (constants.genesis_state_timestamp.0 .0 .0 * 1000000) as i128,
+    )
+    .unwrap();
+    let time_format =
+        format_description!("[year]-[month]-[day] [hour]:[minute]:[second].[subsecond digits:6]Z");
+    let input = format!(
```
akoptelov, line 39: @Sventimir Here a new string is allocated, that contains concatenated bytes. Can you do this by just passing each part to the `hasher.update()`?

core/src/chain_id.rs
```diff
@@ -0,0 +1,160 @@
...
+        tx_max_pool_size: &UnsignedExtendedUInt32StableV1,
+    ) -> ChainId {
+        let mut hasher = Blake2b256::default();
+        let constraint_system_hash = constraint_system_digests
+            .iter()
+            .map(|md5| u8_slice_as_hex(md5))
+            .collect::<Vec<String>>()
+            .join("");
+        let genesis_constants_hash = hash_genesis_constants(genesis_constants, tx_max_pool_size);
+        let input = format!(
```
akoptelov, line 68: See my previous comment.

core/src/chain_id.rs
```diff
@@ -0,0 +1,160 @@
...
+use time::OffsetDateTime;
+
+use std::fmt::{self, Debug, Display, Formatter};
+
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+#[derive(Clone, PartialEq, Eq)]
+pub struct ChainId([u8; 32]);
+
+fn u8_slice_as_hex(slice: &[u8]) -> String {
```
akoptelov, line 15: Let's use `hex::encode` instead of it.

core/src/chain_id.rs
```diff
@@ -0,0 +1,160 @@
...
+        protocol_transaction_version: u8,
+        protocol_network_version: u8,
+        tx_max_pool_size: &UnsignedExtendedUInt32StableV1,
+    ) -> ChainId {
+        let mut hasher = Blake2b256::default();
+        let constraint_system_hash = constraint_system_digests
+            .iter()
+            .map(|md5| u8_slice_as_hex(md5))
+            .collect::<Vec<String>>()
+            .join("");
```
akoptelov, line 66: Probably something like `Iterator::reduce` will be more appropriate then `collect/join`, without allocating a vector.

core/src/chain_id.rs
```diff
@@ -0,0 +1,160 @@
...
+        let input = format!(
+            "{}{}{}{}{}",
+            genesis_state_hash,
+            constraint_system_hash,
+            u8_slice_as_hex(&genesis_constants_hash),
+            md5_hash(protocol_transaction_version),
+            md5_hash(protocol_network_version)
+        );
+        hasher.update(input.as_bytes());
+        ChainId(hasher.finalize().try_into().unwrap())
```
akoptelov, line 77: There should be a way to make hash type-safe for `[u8; 32]`.

core/src/chain_id.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    pub fn as_hex(&self) -> String {
```
akoptelov, line 80: As per Rust convention, methods prefixed with `as_` are computationally cheap way to reinterpret the value. Here, `to_hex` will be more appropriate.

core/src/chain_id.rs
```diff
@@ -0,0 +1,160 @@
...
+        hex::encode(&self.0)
+    }
+
+    pub fn from_hex(s: &str) -> Result<ChainId, hex::FromHexError> {
```
akoptelov, line 84: Let's remove it.

core/src/chain_id.rs
```diff
@@ -0,0 +1,160 @@
...
+        let h = hex::decode(s)?;
+        let bs = h[..32].try_into().or(Err(hex::FromHexError::InvalidStringLength))?;
+        Ok(ChainId(bs))
+    }
+
+    pub fn from_bytes(bytes: &[u8]) -> ChainId {
+        let mut arr = [0u8; 32];
+        arr.copy_from_slice(&bytes[..32]);
```
akoptelov, line 92: It will panic if input is shorter. Let's remove the whole method, or use sized array instead.

akoptelov (commented):
@Sventimir can you please take a look to my comments?

node/src/action.rs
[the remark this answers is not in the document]
tizoc: @akoptelov in this case you need to define a type for this action like in the TODO comment, so that that action is dispatched (and `is_enabled` for that type is checked) instead of `Action::InitializeP2p` directly.

Note that queueing `is_enabled` for `Action` is always called, never directly for the subtype (simplifes a lot of stuff, and having `Action::is_enabled` behave differently from the nested types is also confusing). But somehow during replay it should still return `true` I assume (this here reminded me that I still need to solve that)

core/src/chain_id.rs
tizoc: it will, but what reads better? I don't think these allocations here are something to worry about, this code will be called what, once during startup?

tizoc: oops, sorry, thought this was about the time format above, in this case it is plain concatenation, so `hasher.update()` will read as well as this version.

Sventimir: It works like this:
```
    hasher.update(constants.k.to_string().as_bytes());
    hasher.update(constants.slots_per_epoch.to_string().as_bytes());
    hasher.update(constants.slots_per_sub_window.to_string().as_bytes());
    hasher.update(constants.delta.to_string().as_bytes());
    hasher.update(tx_pool_max_size.to_string().as_bytes());
    hasher.update(genesis_timestamp.format(&time_format).unwrap().as_bytes());
```

Sventimir: Done.

core/src/chain_id.rs
Sventimir: Done.

core/src/chain_id.rs
Sventimir: Done.

core/src/chain_id.rs
Sventimir: Unfortunately, that's not so. `hasher.finalize()` returns `[u8]`, which cannot be safely converted into `[u8; L]`, That's just a shortcoming of the type system, as far as I understand (and I read about it quite a bit today).

core/src/chain_id.rs
Sventimir: But this function is being used down below in the implementation of `Deserialize` trait.

core/src/chain_id.rs
Sventimir: Done.

core/src/chain_id.rs
Sventimir: Okay.

node/src/action.rs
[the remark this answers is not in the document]
akoptelov: This action enables p2p, not called in it.

@binier if I just wrap it into another enum, would it help?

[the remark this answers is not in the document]
akoptelov: Ok, got that now, with EC for Action. Moved to P2pAction, as you suggested.

core/src/chain_id.rs
akoptelov: OK, I mixed up `multihash` library with e.g. `blake2`, that can use fixed-size outputs. We can use that, later.

Sventimir (approved):

binier (approved):

PR #424, opened 2024-05-23
Remove v1 messages, use limit-safe types on wire types.

tizoc:
`String` was replaced by `SmallString1k` in binprot-rs (so that we can still use it for short strings and binprot-rs tests), `Vec` still remains because it is used by the ledger. But every use of `Vec` on wire types was replaced by `List`.


tizoc:
binprot-rs change https://github.com/openmina/binprot-rs/pull/1 (still need some cleanup and tweaking)

ledger/src/proofs/gates.rs
```diff
@@ -298,7 +301,7 @@ pub fn read_constraints_data<F: FieldWitness>(
     // V.t option array list
     let rows_rev = std::fs::read(rows_rev_path).unwrap();
     // std::fs::read(Path::new(env!("CARGO_MANIFEST_DIR")).join("rows_rev.bin")).unwrap();
-    let rows_rev: Vec<Vec<Option<VRaw>>> = Vec::binprot_read(&mut rows_rev.as_slice()).unwrap();
+    let rows_rev: List<List<Option<VRaw>>> = List::binprot_read(&mut rows_rev.as_slice()).unwrap();
```
tizoc, line 304: Unsure about these, I changed this code before restoring `Vec` will probably revert it.

PR #434, opened 2024-05-27
Feat/clippy fixes

0xMimir:
Majority of changes where made automatically by clippy, some of warnings that where style related like naming conventions in enums and naming modules where ignored.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -36,4 +36,4 @@ jobs:
         name: clippy
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-targets -- -A clippy::too_many_arguments -A clippy::doc_markdown
+          args: --all-targets -- -D warnings


node/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::if_same_then_else)]
```
tizoc, line 1: where do we have this case? are we sure it is not a mistake?

node/src/rpc/rpc_effects.rs
```diff
@@ -500,23 +494,14 @@ pub fn rpc_effects<S: Service>(store: &mut Store<S>, action: RpcActionWithMeta)
                 ),
                 Err(err) => RpcSnarkerJobSpecResponse::Err(err),
             };
-            if store
```
tizoc, line 503: wondering if this wasn't a placeholder for some error handling/reporting. Lets add a `// TODO: handle potential errors` or something like that here and bellow on top of `let _ = ...`

tizoc (approved):
LGTM, left some comments about a few small details

node/src/lib.rs
0xMimir: There a few cases where this gets triggered, bcs it is used to dispatch one of actions, example:
```rs
pub fn transition_frontier_sync_ledger_snarked_success_effects<S: redux::Service>(
    _: &ActionMeta,
    store: &mut Store<S>,
) {
    if store.dispatch(TransitionFrontierSyncLedgerAction::Success) {
    } else if store.dispatch(TransitionFrontierSyncLedgerStagedAction::ReconstructEmpty) {
    } else if store.dispatch(TransitionFrontierSyncLedgerStagedAction::PartsFetchPending) {
    }
}
```

core/src/requests/request_id.rs
```diff
@@ -138,11 +138,7 @@ impl<T: RequestIdType> fmt::Debug for RequestId<T> {
 
 impl<T> Clone for RequestId<T> {
     fn clone(&self) -> Self {
-        Self {
-            locator: self.locator,
-            counter: self.counter,
-            _phantom_request_type: self._phantom_request_type.clone(),
-        }
+        *self
```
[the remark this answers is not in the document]
0xMimir: No, I don't think so, because generic, and phantom type, and for same reason other traits like `Eq`, `ParitalEq`, `Ord` are also manually implemented.

node/src/rpc/rpc_effects.rs
0xMimir: I wasn't sure about that, but there are a lot of bindings like that for `Result`.

tizoc: Yes, there are a a bunch of ignored errors (mostly unlikely ones or not very important ones, but it is not that good to silently ignore them). At some point we will need to handle all those (at least some logging so that such errors are not invisible).

akoptelov:
@0xMimir Can you also make this fail on warnings? Also it  is worth checking these allowed warnings, if we can fix them in code instead. 
https://github.com/openmina/openmina/blob/f768dde5a78f485485ba05f00859411ed38bf1ad/.github/workflows/lint.yml#L39

0xMimir:
> @0xMimir Can you also make this fail on warnings? Also it is worth checking these allowed warnings, if we can fix them in code instead.
> 
> https://github.com/openmina/openmina/blob/f768dde5a78f485485ba05f00859411ed38bf1ad/.github/workflows/lint.yml#L39

Updated pipeline to fail if there are warnings, I have removed `-A clippy::too_many_arguments -A clippy::doc_markdown `, since there were only few of these, and they can be easily fixed.

Issue #437, opened 2024-05-28
Make default ledger configuration compatible with berkeleynet


akoptelov:
Currently default configuration has wrong timestamp. Also the timestamp is not stored in prebuilt configuration (created with `ledger-tool`).

The idea is to have complete configuration generated by `ledger-tool`. Also it should be more consistent with used data types, instead of just being a sequence of encoded values.

PR #438, opened 2024-05-28
task(config): fix default ledger configuration

akoptelov:
fixes: #437


akoptelov:
BTW, cbor encoded prebuilt ledger is about 5M, while binprot-encoded one is only 2M. So might be worth switching back to binprot.

genesis_ledgers/berkeley.bin
tizoc, line 1: @akoptelov Old file was binprot and new one is cbor right? Size is more than double now

tools/ledger-tool/src/main.rs
```diff
@@ -23,151 +17,23 @@ struct Args {
...
-        mask.get_or_create_account(account_id, account.clone())
-            .expect("must not be full");
-    }
-
-    let root = mask.merkle_root();
-    let top_hash = v2::LedgerHash::from_fp(root);
-    println!("{top_hash}");
-
-    let mut hashes = vec![];
+    let daemon_json = serde_json::from_slice::<node::daemon_json::DaemonJson>(&data)?;
```
tizoc, line 31: Much better.

Btw the only reason we are doing this now is that the node doesn't cache the genesis ledger, but I think that is what should happen instead. And on startup, if the cache file is there (matched by hash?) it is loaded instead of parsing the config file to obtain the genesis ledger.

To speed up tests we can still have pre-generated files if necessary.

genesis_ledgers/berkeley.bin
tizoc: Ok, just saw your comment. @akoptelov for this file it may not be that important but for mainnet ledgers the impact is going to be much higher, so yes, I would go back to binprot.

tools/ledger-tool/src/main.rs
akoptelov: The thing is that genesis ledger is not serializable. The closest thing is what we store in that prebuilt configuration.

tizoc: @akoptelov ok but in that case we can still serialize the same data we are producing with this tool and if it exists load that instead of parsing and processing the full config file.

genesis_ledgers/berkeley.bin
akoptelov: @tizoc so, do you think 3M is worth using binprot? Actually this is quite easy to do.

tizoc: @akoptelov yes, but it is not about the 3mb, it is about the > 50% savings in space for any ledger size we end up after the hardfork (either when caching the results as mentioned in the other thread, or if we introduce support for downloading pre-constructed data like the OCaml node by referencing it just by hash)

tizoc (approved):
LGTM. Although not super important right now I would revert back to binprot to keep the preprocessed result files smaller.

tools/ledger-tool/src/main.rs
akoptelov: So, you mean if we use `--config ...`, we check if there's cached prebuilt configuration for it, and use it, otherwise generate and store it?

tizoc: Yes, at least for the accounts + hashes. Similar to what the OCaml node does with the genesis and epoch ledgers, which are stored in `~/.mina-config/genesis` and reused if they exists. This saves you from having to reconstruct the genesis ledger each time, and also from syncing up the epoch ledgers if you restart the node and the epoch is still the same as last time.

tizoc: You can see the stored ledgers (tagged by their hash):

```
> ls ~/.mina-config/genesis
epoch_ledger_3e2c138e3e014dcb66f5dcfbfd42e1c8cbbac7aeb5bbf90ed62907b91075a9f5
epoch_ledger_3e2c138e3e014dcb66f5dcfbfd42e1c8cbbac7aeb5bbf90ed62907b91075a9f5.tar.gz
epoch_ledger_b595fe2dbde04c83164b527df891ab445366a8dabe5ca448495e808a64c3ecdd
epoch_ledger_b595fe2dbde04c83164b527df891ab445366a8dabe5ca448495e808a64c3ecdd.tar.gz
genesis_ledger_7f8e3c72be57f296e86e1083379c7a09c6131b1567510cf0019eb4aff8b87813
genesis_ledger_7f8e3c72be57f296e86e1083379c7a09c6131b1567510cf0019eb4aff8b87813.tar.gz
```

tizoc: note that in OCaml they store the actual rocksdb database, but we don't need that, just accounts + hashes is enough to quickly reconstruct the ledger for us

PR #447, opened 2024-05-31
Small noise fix

vlad9486:


tizoc:
@vlad9486 remember to remove the MIN_STACK_SIZE setting in the test too

p2p/src/network/noise/p2p_network_noise_effects.rs
```diff
@@ -121,8 +121,12 @@ impl P2pNetworkNoiseAction {
                 }
             }
             Self::IncomingData { addr, .. } => {
-                if let Some(data) = incoming {
-                    store.dispatch(P2pNetworkNoiseAction::IncomingChunk { addr, data });
+                let mut incoming = incoming;
+                while let Some(data) = incoming.pop_front() {
```
tizoc, line 125: since `incoming` is not referenced anywhere else now you could just:

```rust
for data in state.incoming_chunks.clone() {
    // ...
}
```

as in https://github.com/openmina/openmina/blob/eeda46514ec68f91a582496cd18f72ca5d2a8870/p2p/src/network/yamux/p2p_network_yamux_effects.rs#L26-L30

PR #478, opened 2024-06-06
Build FE multiple FE images with different configurations

adonagy:
To build a separate image used for demos or for different environments in the future


.github/workflows/docker.yaml
```diff
@@ -46,14 +46,22 @@ jobs:
...
           tags: ${{ steps.meta.outputs.tags }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
   build-openmina-frontend-image:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        configuration:
+          - build_configuration: compose
```
akoptelov, line 61: Can we have more descriptive configuration name here?

adonagy: Sure, let's do it in another PR, as this is a change on the FE side directly 
cc @directcuteo

PR #490, opened 2024-06-11
p2p fixes

dkuehr:
- protocol select: improve/fix timeouts
- pnet: implement timeouts for nonce exchange
- identify: close streams
- pubsub: close streams

### DIFF
--- a/p2p/src/network/identify/p2p_network_identify_state.rs
+++ b/p2p/src/network/identify/p2p_network_identify_state.rs
@@ -39,4 +39,8 @@ impl P2pNetworkIdentifyState {
             .get_mut(peer_id)
             .map_or(false, |m| m.remove(stream_id).is_some())
     }
+
+    pub fn prune_peer_state(&mut self, peer_id: &PeerId) {
+        self.streams.remove(peer_id);
+    }
 }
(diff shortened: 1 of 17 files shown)


p2p/src/network/identify/p2p_network_identify_state.rs
```diff
@@ -39,4 +39,8 @@ impl P2pNetworkIdentifyState {
             .get_mut(peer_id)
             .map_or(false, |m| m.remove(stream_id).is_some())
     }
+
+    pub fn remove_peer(&mut self, peer_id: &PeerId) {
```
akoptelov, line 43: Let's give it a name that better describes its functionality.

p2p/src/p2p_config.rs
```diff
@@ -110,6 +111,7 @@ impl Default for P2pTimeouts {
                 Some(Duration::from_secs(5)),
             ),
             select: from_env_or("SELECT_TIMEOUT", Some(Duration::from_secs(5))),
+            pnet: Some(Duration::from_secs(5)),
```
akoptelov, line 114: Please make it configurable via env. BTW, isn't 5 secs too much for PNET?

akoptelov (dismissed):
LGTM, with some minor changes.

p2p/src/network/identify/p2p_network_identify_state.rs
dkuehr: Would `purge_peer_state` do?

p2p/src/p2p_config.rs
dkuehr: Sure, I'll make it configurable via env var like you suggest.
Five secs is probably too much but I wanted to be on the safe side for some of the tests, maybe we can reduce the default value now and increase it via env var for the tests.

PR #494, opened 2024-06-12
Fixed remaining clippy::large_enum_variant

0xMimir:
Removed remaining `#[allow(clippy::large_enum_variant)]`, there are still 2 places it is allowed on `Command` in [`cli/src/commands/mod.rs`](https://github.com/openmina/openmina/blob/develop/cli/src/commands/mod.rs) and `NodeTestingConfig` in [`node/testing/src/node/config.rs`](https://github.com/openmina/openmina/blob/develop/node/testing/src/node/config.rs).


tizoc (approved):
LGTM

PR #505, opened 2024-06-17
P2p add channel for transaction propagation

binier:
closes: #502


binier:
Two of the tests in CI fail because since today berkeley network is shut down and those tests rely on berkeley network.

PR #508, opened 2024-06-21
Add openmina node and tooling to frontend workflow

adonagy:


binier:
@adonagy Wouldn't it be easier if we create scenario for frontend tests, set up rust node there and run cypress tests from the scenario itself.

Should be easier and more flexible than that approach.

dannywillems:
Closing. Old.

PR #509, opened 2024-06-21
Implement genesis config caching.

Sventimir:
When genesis config is loaded and genesis ledger reconstructed from JSON config, it is also cached on disk. When the same JSON loaded again, the node finds the cache on disk and loads it instead, which is significantly faster than building the ledgers from JSON.

### DIFF
--- a/node/src/daemon_json/json_ledger.rs
+++ b/node/src/daemon_json/json_ledger.rs
@@ -75,34 +75,44 @@ impl Ledger {
     /// See: https://github.com/MinaProtocol/mina/blob/develop/src/lib/genesis_ledger_helper/genesis_ledger_helper.ml#L105
     pub fn ledger_name(&self) -> String {
         self.hash.clone().unwrap_or_else(|| {
-            let mut hash = Blake2b256::default();
-            hash.update(LEDGER_DEPTH.to_string().as_bytes());
-            let balances = self
-                .balances
-                .clone()
-                .unwrap_or_default()
-                .iter()
-                .fold(String::new(), |acc, (i, balance)| {
-                    format!("{} {} {}", acc, i, balance)
-                });
-            hash.update(balances.as_bytes());
-            let empty_account = ledger::Account::empty();
-            hash.update(empty_account.hash().to_string().as_bytes());
-            let mut empty_account_enc: Vec<u8> = Vec::new();
-            empty_account
-                .binprot_write(&mut empty_account_enc)
-                .expect("failed to write account");
-            hash.update(empty_account_enc.as_slice());
-            format!("{:x?}", hash.finalize())
+            build_ledger_name(
+                self.num_accounts.unwrap_or(0),
+                self.balances
+                    .clone()
+                    .unwrap_or_default()
+                    .iter()
+                    .map(Clone::clone),
+            )
         })
     }
+}
+
+pub fn build_ledger_name(
+    num_accounts: usize,
+    balances: impl Iterator<Item = (usize, RawCurrency)>,
+) -> String {
+    let mut hash = Blake2b256::default();
+    hash.update(LEDGER_DEPTH.to_string().as_bytes());
+    hash.update(num_accounts.to_string().as_bytes());
+    let balances = balances.fold(String::new(), |acc, (i, balance)| {
+        format!("{} {} {}", acc, i, balance)
+    });
+    hash.update(balances.as_bytes());
+    let empty_account = ledger::Account::empty();
+    hash.update(empty_account.hash().to_string().as_bytes());
+    let mut empty_account_enc: Vec<u8> = Vec::new();
+    empty_account
+        .binprot_write(&mut empty_account_enc)
+        .expect("failed to write account");
+    hash.update(empty_account_enc.as_slice());
+    format!("{:x?}", hash.finalize())
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Account {
     pk: String,
     sk: Option<String>,
-    balance: RawCurrency,
+    pub(super) balance: RawCurrency,
     delegate: Option<String>,
     token_id: Option<String>,
     token_symbol: Option<String>,
--- a/node/src/daemon_json/json_ledger.rs
+++ b/node/src/daemon_json/json_ledger.rs
@@ -1,5 +1,7 @@
 use core::str::FromStr;
 use mina_hasher::Fp;
+use mina_p2p_messages::binprot::BinProtWrite;
+use multihash::{Blake2b256, Hasher};
 use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::Value;
 use std::fmt::{self, Display, Formatter};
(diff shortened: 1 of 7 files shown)


node/src/daemon_json/json_ledger.rs
```diff
@@ -61,6 +65,36 @@ impl Ledger {
...
         }
     }
+
+    /// Typically a ledger is identified by its hash, but if hash is
+    /// not given in config, we construct a name by hashing the ledger
+    /// config so that we can recognise a ledger we already have
+    /// cached without reconstructing it (which is expensive).
+    /// Note that the name contains some constants - the purpose of this
+    /// is to invalidate all caches when these constants change.
+    pub fn ledger_name(&self) -> String {
```
tizoc, line 75: @Sventimir does OCaml perform this same hashing? if that is the case please include a permalink to the original implementation.

node/src/transition_frontier/genesis/transition_frontier_genesis_config.rs
```diff
@@ -143,67 +150,33 @@ impl GenesisConfig {
...
+                let config_name = ledger.ledger_name();
+                let cache_filename =
+                    openmina_cache_path(format!("prebuilt_configs/{}.bin", config_name)).unwrap();
+                let filename_str = cache_filename
+                    .clone()
+                    .into_os_string()
+                    .into_string()
+                    .unwrap();
+                if cache_filename.is_file() {
+                    let fd = std::fs::File::open(cache_filename.clone()).unwrap();
```
tizoc, line 171: Lets turn this unwrap call into a returned error

tizoc (changes requested):
@Sventimir this version will already provide a speedup, but the idea is to cache only the ledgers (which are the expensive part of processing these config files) so that they can be referenced by their hash without actually including the accounts, so that config files like this one (which includes the hashes but no account data) can be used in the future with support for downloading the prebuilt ledgers (out of scope for this PR):

```json
{
  "genesis": {
    "genesis_state_timestamp": "2024-04-09T21:00:00Z"
  },
  "proof": {
    "fork": {
      "state_hash": "3NKSvjaGSKiQuAt8BP1b1VCpLbJc9RcEFjYCaBYsJJFdrtd6tpaV",
      "blockchain_length": 296371,
      "global_slot_since_genesis": 445860
    }
  },
  "ledger": {
    "add_genesis_winner": false,
    "hash": "jy1wjiJgTkzXr7yL8r5x4ikaNJuikibsRMnkjdH6uqGCsDmR2sf",
  },
  "epoch_data": {
    "staking": {
      "seed": "2vbdR3vZUxh8NFnsnrefko7YCiLUKNPUbQpqSQAFXFijtTjd3531",
      "hash": "jxvdSnpgWjjERJFdcYNxChJ9F39v4tGrNSzDcaQXuXeJnicrYEP",
    },
    "next": {
      "seed": "2vaduyGw7uST2rT1HmvSxjrkZX26XnFvu4NJvCebMmLzoWaQtdGT",
      "hash": "jxjmYHK1tnGy5S6CDaLDTRrDbw17DmtEd6Jnw3L86QSsLm15w5g",
    }
  }
}
```

node/src/daemon_json/json_ledger.rs
Sventimir: I updated the last commit with the the following link:
https://github.com/MinaProtocol/mina/blob/develop/src/lib/genesis_ledger_helper/genesis_ledger_helper.ml#L105

node/src/transition_frontier/genesis/transition_frontier_genesis_config.rs
Sventimir: Done.

Sventimir:
@tizoc, I added another commit to cache just the genesis ledger, without epoch ledgers and everything else.

node/src/transition_frontier/genesis/transition_frontier_genesis_config.rs
```diff
@@ -374,6 +363,22 @@ fn genesis_producer_stake_proof(mask: &ledger::Mask) -> v2::MinaBaseSparseLedger
     (&sparse_ledger).into()
 }
 
+#[derive(Debug, BinProtRead, BinProtWrite)]
+struct LedgerAccountsWithHash {
+    accounts: Vec<MinaBaseAccountBinableArgStableV2>,
```
tizoc, line 368: @Sventimir this doesn't include the merkle tree hashes, this means that this still requires re-hashing of every node in the tree to recompute it, which is the most expensive part of producing these ledgers.

The data stored should be the same the prebuilt config stores and rebuilt in the same way as `GenesisConfig::build_ledger_from_accounts_and_hashes` which loads the accounts and sets the hashes directly without any rehashing:

```rust
    ledger_hash: LedgerHash,
    accounts: Vec<MinaBaseAccountBinableArgStableV2>,
    hashes: Vec<(u64, LedgerHash)>,
```

tizoc (changes requested):
@Sventimir See my comment regarding hashes, also lets have this same logic for the epoch ledgers too, otherwise only 1/3 of the total work will be optimized.

node/src/transition_frontier/genesis/transition_frontier_genesis_config.rs
Sventimir: Done.

Sventimir:
I added caching genesis epoch ledgers as well as the genesis ledger. Also added caching inner ledger hashes (as opposed to only the root hash).

node/src/transition_frontier/genesis/transition_frontier_genesis_config.rs
```diff
@@ -284,6 +244,69 @@ impl GenesisConfig {
...
+            ledger::Mask,
+            v2::CurrencyAmountStableV1,
+            Base58CheckOfBinProt<
+                v2::MinaBaseLedgerHash0StableV1,
+                Versioned<v2::MinaBaseLedgerHash0StableV1, 1>,
+                5,
+            >,
+        ),
+        GenesisConfigError,
+    > {
```
tizoc, line 261: @Sventimir clippy is complaining here

tizoc:
```
            Base58CheckOfBinProt<v2::MinaBaseLedgerHash0StableV1, ...  >
```
is just `LedgerHash`

node/src/transition_frontier/genesis/transition_frontier_genesis_config.rs
tizoc: small request, lets put `ledger_hash` as the first element, so that it can be read quickly without having to go through all the accounts first

Sventimir: Done.

node/src/transition_frontier/genesis/transition_frontier_genesis_config.rs
Sventimir: Done.

tizoc (approved):
@Sventimir LGTM. One last request, lets have an extra message in the logs mentioning that the node is trying to load the ledgers before it starts, otherwise it looks like the node is stuck (looks like so already and it will be much worse with the mainnet ledgers).

Then combine commits as you see fit and merge.

Sventimir:
@tizoc , very well, I added the log message.

PR #510, opened 2024-06-21
fix: Restore original `Action::is_enabled` condition during replays w…

tizoc:
…here the result is always `true`


tizoc:
Closing because actually this breaks replay.

PR #556, opened 2024-07-19
P2P: Introduce ConnectionAddr type to differentiate incoming/outgoing…

dkuehr:
The following PR includes the following changes:

## Minor Fixes and Enhancements
This includes the revision of expressions such as `Self::*action-variant*`, which have been updated to their explicit forms.

## Introduction of ConnectionAddr
The `ConnectionAddr` type essentially links a `SockAddr` to a `bool`, differentiating incoming connections from outgoing ones. Most actions and collections that previously utilized `SockAddr` to track a specific connection now employ `ConnectionAddr`.

This modification mitigates potential collisions that could occur if multiple connections with the same `ip:port` pair occur simultaneously. For instance, while OpenMina does not employ port-reuse (source port for outgoing connections is always different) the OCaml node reuses the same listening port as the source port for outgoing connections. If the OCaml node attempts to connect to us (OpenMina) concurrently with our connection to it, the OpenMina node could encounter two `SockAddr` with identical `ip:port`. This situation is feasible as OpenMina continues to use a random source port. This issue could be proactively mitigated (at the TCP level) if OpenMina adopted port-reuse in the same manner as the OCaml node. In such a scenario, the differentiation of connections introduced by this change wouldn't be necessary.

Related issue: https://github.com/openmina/openmina/issues/399

## Enforcing Peer-ID Verification
The peer-id, in conjunction with the peer's IP and port, constitutes a portion of a peer's (multi)address. Until now, when connecting to a peer, we have not verified that the peer-id reported by it (during the Noise handshake) aligns with the peer-id specified as part of its address. This update introduces such a check, and the Noise handshake will now fail (resulting in the peer's disconnection) if the peer-ids do not match.

### Limitations

This check is only performed for IDs we are aware of: either those from the initial peer list, or discovered by Kademlia.
We are unable to verify an incoming connection from a peer that hasn't been discovered before (we allow these to pass through).
A potential issue could arise if a malicious peer exploits the Kademlia protocol to report incorrect peer IDs (for valid IP addresses), thereby contaminating other nodes' routing tables. The nodes may attempt to connect to a valid IP and terminate the connection upon realizing the node's peer-id does not match the one obtained from Kademlia. This potential issue warrants further investigation

### DIFF
--- a/node/testing/src/scenarios/driver.rs
+++ b/node/testing/src/scenarios/driver.rs
@@ -10,8 +10,9 @@ use node::{
     p2p::{
         connection::outgoing::P2pConnectionOutgoingInitOpts,
         webrtc::{Host, HttpSignalingInfo, SignalingMethod},
-        P2pConnectionEvent, P2pEvent, P2pNetworkConnectionMuxState, P2pNetworkConnectionState,
-        P2pNetworkYamuxState, P2pPeerState, P2pPeerStatus, P2pState, PeerId,
+        ConnectionAddr, P2pConnectionEvent, P2pEvent, P2pNetworkConnectionMuxState,
+        P2pNetworkConnectionState, P2pNetworkYamuxState, P2pPeerState, P2pPeerStatus, P2pState,
+        PeerId,
     },
     State,
 };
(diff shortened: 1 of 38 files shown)


0xMimir (approved):
LGTM

vlad9486 (approved):

PR #558, opened 2024-07-24
Removed unused code in select protocol

0xMimir:
Removed optimized handshake, removed optional handshake sending, added proper way to respond to handshake token.

### DIFF
--- a/p2p/src/network/select/p2p_network_select_effects.rs
+++ b/p2p/src/network/select/p2p_network_select_effects.rs
@@ -1,3 +1,8 @@
+use self::token::{
+    AuthKind, DiscoveryAlgorithm, IdentifyAlgorithm, MuxKind, PingAlgorithm, Protocol,
+    RpcAlgorithm, StreamKind, Token,
+};
+
 use openmina_core::{fuzz_maybe, fuzzed_maybe};
 
 use crate::{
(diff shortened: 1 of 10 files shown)


p2p/src/network/select/p2p_network_select_effects.rs
```diff
@@ -288,20 +214,24 @@ impl P2pNetworkSelectAction {
                 }
             }
             P2pNetworkSelectAction::IncomingToken { addr, kind, .. } => {
-                if let P2pNetworkSelectStateInner::Error(error) = &state.inner {
```
dkuehr, line 291: Is it safe to remove this error check?

p2p/src/network/select/p2p_network_select_effects.rs
```diff
@@ -8,86 +13,39 @@ use crate::{
...
-        }
-
-        if self.negotiated.is_some() {
-            effects.push(forward_data_action(addr, kind, data, fin));
-            return;
-        }
-
-        let payload_data = self.recv.buffer.clone();
+/// Propagates incoming data to corresponding action
+macro_rules! forward_incoming_data {
```
dkuehr, line 17: This is up to you but in general we try to be conservative with the use of macros, and consider the pros and cons:

- Does the reduced code duplication/redundancy outweigh the increased complexity to understand and navigate the code?
- How can this affect coverage reports (profiling, fuzzing)?
- How can this affect debugging info/backtraces?

dkuehr (commented):
I left a couple of comments. The rest LGTM 👍

p2p/src/network/select/p2p_network_select_effects.rs
0xMimir: I have tried putting it into new Action, but that triggers `NoRecursion` invariant, before there was function in state that putted all actions into vector and after that they were dispatched, this made debugging that part much harder. 

Problem is that state is immutable reference on store, but for dispatch store needs to be mutable, so I think pros outweigh cons.

tizoc: @0xMimir @0xMimir re: mutable access to the store, this is solved in the new style reducers (which combine reducer + dispatcher in the same handler function), but to take advantage of it the whole component needs to be ported: https://github.com/openmina/openmina/blob/898750c4007e8351e6d4792da234a20851253518/ARCHITECTURE.md#substate-access-queued-reducer-dispatch-and-callbacks

Probably not worth it just for this change but eventually we should have all our components ported

PR #560, opened 2024-07-25
p2p/kad: fix buffer index miscalculation

dkuehr:
This fixes a bug found by p2p-fuzzer in kademlia stream reducer.

### DIFF
--- a/p2p/src/network/kad/stream/p2p_network_kad_stream_reducer.rs
+++ b/p2p/src/network/kad/stream/p2p_network_kad_stream_reducer.rs
@@ -31,30 +31,39 @@ impl P2pNetworkKadIncomingStreamState {
             }
             (S::WaitingForRequest { .. }, A::IncomingData { data, .. }) => {
                 let data = &data.0;
-
                 let mut reader = BytesReader::from_bytes(data);
-                let Ok(len) = reader.read_varint32(data).map(|v| v as usize) else {
+
+                let Ok(encoded_len) = reader.read_varint32(data).map(|v| v as usize) else {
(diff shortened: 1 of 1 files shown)


p2p/src/network/kad/stream/p2p_network_kad_stream_reducer.rs
```diff
@@ -31,30 +31,32 @@ impl P2pNetworkKadIncomingStreamState {
...
-                        len,
-                        data: data[(len - reader.len())..].to_vec(),
+                        len: encoded_len,
+                        data: data[(data.len() - remaining_len)..].to_vec(),
                     };
                     return Ok(());
                 }
 
-                self.handle_incoming_request(len, &data[data.len() - reader.len()..])
+                self.handle_incoming_request(encoded_len, &data[data.len() - remaining_len..])
```
sebastiencs, line 59:
```suggestion
                self.handle_incoming_request(encoded_len, data.get(data.len() - remaining_len..).unwrap()) // Without the `unwrap`
```

Is it parsing incoming data ? We should probably use `slice::get` instead ?
On line 54 too

dkuehr: It seems to be decoding the length from incoming data and then using the rest of it. Maybe I could just use the reader with `.read_message_by_len()`

dkuehr: Modified to use `slice::get` now. I think the new else condition is unreachable but I don't think that will be an issue.

0xMimir (approved):

PR #578, opened 2024-08-02
Fixed tests in connection discovery

0xMimir:
Fixed and completed test in `node/testing/src/scenarios/multi_node/connection_discovery.rs`, test take about 20 minutes to run locally, but 40-ish minutes to run in CI.

Removed async requests to OCaml graphql in `node/testing/src/node/ocaml/mod.rs`, this is because calling these function panics, since calling `reqwest::blocking::Client::new();` panics.

Rewrote `KademliaBootstrap` test in `node/testing/src/scenarios/p2p/kademlia.rs` and removed `IncomingFindNode` test, since these tests where only used for debugging during development of Kademlia layer, and added that test in CI.


node/testing/src/scenarios/multi_node/connection_discovery.rs
```diff
@@ -1,275 +1,209 @@
...
             initial_peers: vec![rust_node_dial_addr],
-            daemon_json: DaemonJson::Custom("/var/lib/coda/config_dc6bf78b.json".to_owned()),
-            block_producer: None,
+            ..Default::default()
         };
 
         let ocaml_node0 = runner.add_ocaml_node(ocaml_node_config.clone());
         let ocaml_peer_id0 = runner.ocaml_node(ocaml_node0).unwrap().peer_id();
 
+        sleep(Duration::from_secs(30)).await;
```
tizoc, line 39: @0xMimir why is this sleep here needed?

0xMimir: It is to make sure node0 connect before node1.

tizoc: @0xMimir ok good, please add a comment mentioning the purpose. In the future we should have a way to encode the fact that we are waiting for some specific event to happen (with a timeout) before continuing instead of plain sleeps.

PR #579, opened 2024-08-02
Update producer-demo.md

JanSlobodnik:
* merge steps of Linux/Mac/Windows
* add screenshot of UI that user should see upon completion of steps.

### DIFF
--- a/docs/producer-demo.md
+++ b/docs/producer-demo.md
@@ -78,7 +78,7 @@ This demo showcases the block production capabilities of OpenMina nodes within a
 
 ## Running the Producer Demo
 
-### Running on Debian-based Linux
+Once pre-requisites for your OS have been completed, follow these steps (they are the same for Debian-based Linux, Windows and MacOS):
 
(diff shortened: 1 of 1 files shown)


docs/producer-demo.md
```diff
@@ -98,46 +98,6 @@ This demo showcases the block production capabilities of OpenMina nodes within a
...
-    cd openmina
-    ```
-
-4. **Run the following command to start the demo:**
-    ```sh
-    docker compose -f docker-compose.local.producers.yml up
-    ```
-
-5. **Open you browser and visit http://localhost:8070**
+![image](https://github.com/user-attachments/assets/1eac34aa-41b7-42ab-a193-0dcdbda32c0f)
```
adonagy, line 103: We should use a more up to date image (the image there has the nanoMina representation). Also it is from a longer run, maybe we should show the screen right after the commands are run?

adonagy (commented):
Only merge after #255 is merged.

PR #619, opened 2024-09-02
Run ledger tests in wasm

sebastiencs:


dannywillems:
Code is old. Closing.

PR #632, opened 2024-09-04
feat(p2p-messages): Add `ZkappUri` and `TokenSymbol` type with size limit

### LINKED ISSUE #623: As a developer, I want to add limits to zkapp uris, to match the OCaml node
We are missing that limit for zkapp updates.

tizoc:
fixes #623

### DIFF
--- a/ledger/src/account/account.rs
+++ b/ledger/src/account/account.rs
@@ -113,15 +113,15 @@ impl Default for TokenSymbol {
     }
 }
 
-impl TryFrom<&mina_p2p_messages::string::ByteString> for TokenSymbol {
+impl TryFrom<&mina_p2p_messages::string::TokenSymbol> for TokenSymbol {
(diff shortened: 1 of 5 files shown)


sebastiencs:
@tizoc  Not really related to your changes, but I just realized that we will need to change our [TokenSymbol](https://github.com/openmina/openmina/blob/4a1bd7345b6a889df12f6e74df9e28abf0a5e7c2/ledger/src/account/account.rs#L74) and [ZkAppUri](https://github.com/openmina/openmina/blob/4a1bd7345b6a889df12f6e74df9e28abf0a5e7c2/ledger/src/account/account.rs#L559).
Right now they are Rust's `String` (with UTF-8 encoding), but in OCaml they are `string` which are just immutable `bytes` I think ?

tizoc:
@sebastiencs yes, and they can contain any sequence of bytes, not just valid utf8

PR #638, opened 2024-09-05
Improve crash resilience in p2p

vlad9486:


node/common/src/service/p2p.rs
```diff
@@ -87,4 +87,13 @@ impl P2pCryptoService for NodeService {
             .sign(&msg)
             .expect("unable to create signature")
     }
+
+    fn verify_publication(&mut self, pk: &[u8], publication: &[u8], sig: &[u8]) -> bool {
+        let Ok(pk) = libp2p_identity::PublicKey::try_decode_protobuf(pk) else {
+            return false;
```
dkuehr, line 93: Is it ok just to return false on decode errors or should we propagate errors?
We might end silently ignoring some errors

node/common/src/service/p2p.rs
```diff
@@ -87,4 +87,13 @@ impl P2pCryptoService for NodeService {
...
+        };
+        println!("pk {:?}", pk);
```
dkuehr, line 95: println can be removed

node/common/src/service/p2p.rs
vlad9486: Yes, you are right. I will add TODO. Higher, we just silently ignore the message, no matter if it is malformed or the signature is invalid. Anyway, we don't have a punishment system for peers who misbehave, so we can't do anything about it.

https://github.com/openmina/openmina/blob/cce612e7edcedffb8af01877b0c3ad7a5351a8c3/p2p/src/network/pubsub/p2p_network_pubsub_effects.rs#L146-L155

p2p/src/network/noise/p2p_network_noise_reducer.rs
```diff
@@ -68,13 +68,15 @@ impl P2pNetworkNoiseState {
                 let mut offset = 0;
                 loop {
                     let buf = &self.buffer[offset..];
-                    if buf.len() >= 2 {
-                        let len = u16::from_be_bytes(buf[..2].try_into().expect("cannot fail"));
+                    let len = buf
+                        .get(..2)
+                        .and_then(|buf| Some(u16::from_be_bytes(buf.try_into().ok()?)));
```
dkuehr, line 73: Here we silently ignore errors when converting the result to optional.
I implemented a `bug_condition!()` macro that I'll push in my next PR that can be used to either log the errors (and continue) or crash the node depending on an environment variable.
Please add TODO comment here so I can add a `bug_condition!()` call later.

p2p/src/network/noise/p2p_network_noise_reducer.rs
```diff
@@ -188,55 +190,72 @@ impl P2pNetworkNoiseState {
...
-                        let chunks = data
-                            .chunks(chunk_max_size)
-                            .map(|data| {
-                                let mut chunk = Vec::with_capacity(18 + data.len());
-                                chunk.extend_from_slice(&((data.len() + 16) as u16).to_be_bytes());
-                                chunk.extend_from_slice(data);
+                        let mut chunks = vec![];
+
+                        for data in data.chunks(chunk_max_size) {
+                            let mut chunk = Vec::with_capacity(18 + data.len());
```
dkuehr, line 196: Please add constant names or code comments explaining the origin of hardcoded values in size calculations to better understand them.

p2p/src/network/noise/p2p_network_noise_reducer.rs
```diff
@@ -188,55 +190,72 @@ impl P2pNetworkNoiseState {
...
+                            chunk.extend_from_slice(&((data.len() + 16) as u16).to_be_bytes());
+                            chunk.extend_from_slice(data);
+
+                            let mut nonce = GenericArray::default();
```
dkuehr, line 200: Could you add an explicit type signature for `nonce`? It is not obvious to figure out what array size the compiler has inferred.

p2p/src/network/noise/p2p_network_noise_reducer.rs
```diff
@@ -188,55 +190,72 @@ impl P2pNetworkNoiseState {
...
+                            nonce[4..].clone_from_slice(&send_nonce.to_le_bytes());
```
dkuehr, line 201: For later discussion: array indexes vs using `get()`. This is a similar problem to calling `unwrap()` if we do the calculation incorrectly node crashes.

For code that is obviously right I believe using indexes is more clear, but for other more complex indexing we probably want to use `get()`.

p2p/src/network/noise/p2p_network_noise_state.rs
```diff
@@ -127,7 +127,8 @@ impl NoiseState {
         let hkdf = Hkdf::<Sha256, Hmac<Sha256>>::new(Some(&self.chaining_key.0), &secret);
         secret.zeroize();
         let mut okm = [0; 64];
-        hkdf.expand(&[], &mut okm).unwrap();
+        hkdf.expand(&[], &mut okm)
+            .expect("the length is constant and small");
```
dkuehr, line 131: Is the `expect` here fine because sizes are hardcoded and can't break unless the code is changed? Could you add a brief comment explaining it?

vlad9486: Yes, this method returns an `InvalidLength` error and this is invalid if the size is greater than some constant (255 times the block size). In our case the size is always 64. Even if the constant changes in the future, this will cause any test to fail. So we will not miss this.

p2p/src/network/pubsub/p2p_network_pubsub_effects.rs
```diff
@@ -100,7 +100,9 @@ impl P2pNetworkPubsubAction {
             P2pNetworkPubsubAction::Broadcast { message } => {
                 // println!("(pubsub) {this} broadcast");
                 let mut buffer = vec![0; 8];
-                binprot::BinProtWrite::binprot_write(&message, &mut buffer).expect("msg");
+                if binprot::BinProtWrite::binprot_write(&message, &mut buffer).is_err() {
+                    return;
```
dkuehr, line 104: Please add TODO comment here so I can add `bug_condition!()` later

p2p/src/network/pubsub/p2p_network_pubsub_effects.rs
```diff
@@ -111,16 +113,55 @@ impl P2pNetworkPubsubAction {
...
             }
             P2pNetworkPubsubAction::BroadcastSigned { .. } => broadcast(store),
-            P2pNetworkPubsubAction::IncomingData { peer_id, .. } => {
+            P2pNetworkPubsubAction::IncomingData {
+                peer_id,
+                seen_limit,
+                ..
+            } => {
+                let Some(state) = state.clients.get(&peer_id) else {
+                    return;
```
dkuehr, line 135: Is this an error condition or something normal?

p2p/src/network/pubsub/p2p_network_pubsub_effects.rs
```diff
@@ -111,16 +113,55 @@ impl P2pNetworkPubsubAction {
...
+
+                for mut message in messages {
+                    if let (Some(signature), Some(from)) =
+                        (message.signature.take(), message.from.clone())
+                    {
+                        message.key = None;
+                        let mut data = vec![];
+                        #[allow(clippy::if_same_then_else)]
+                        if prost::Message::encode(&message, &mut data).is_err() {
+                            continue;
```
dkuehr, line 147: Please add TODO comment so I'll add `bug_condition!()`

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -227,6 +107,137 @@ impl P2pNetworkPubsubState {
...
+            P2pNetworkPubsubAction::IncomingMessage {
+                peer_id,
+                message,
+                seen_limit,
+            } => {
+                self.incoming_transactions.clear();
+                self.incoming_snarks.clear();
+
+                let Some(state) = self.clients.get_mut(peer_id) else {
+                    return;
```
dkuehr, line 173: Is this normal or an error condition?

dkuehr (changes requested):
I see some error conditions cause the dispatch of `P2pNetworkPubsubAction::OutgoingMessageError` but the handling of it seems to be empty.
Do we really need to dispatch this? Otherwise just leave comments in the code and I'll replace it for `bug_condition!()` later.

p2p/src/network/pubsub/p2p_network_pubsub_effects.rs
tizoc: if the enabling condition is correct, this should never fail because the existence of such peer_id in the collection should have been checked already.

But still in general, as part of the refactor (in a later phase) reducers should return a `Result` and all these cases should return an error that we can log so that this never fails silently if we have a bug

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
tizoc: I mean, "normal" as in, the pattern is one we have everywhere, but it should never return here

p2p/src/network/pubsub/p2p_network_pubsub_effects.rs
dkuehr: Maybe we could add later `bug_condition!()` in such cases, that doesn't require immediate changing the signatures of reducers and propagate the Result. Also I could include in the error message for `bug_condition!()` that the possible cause is that the enabling condition is not strong enough.

dkuehr:
> > I see some error conditions cause the dispatch of `P2pNetworkPubsubAction::OutgoingMessageError` but the handling of it seems to be empty. Do we really need to dispatch this? Otherwise just leave comments in the code and I'll replace it for `bug_condition!()` later.
> 
> @dkuehr even if the handling code is empty, it is still useful to have such error action dispatched because it will at least show up in the logs and recordings.

I prefer to have the logs at the place of the error otherwise we could have many possible sources for the same error action., and also the option to panic in place too.

dkuehr (approved):

PR #640, opened 2024-09-05
Replace fallible from implementations by try_from version

dkuehr:
This is a WIP with the purpose of removing code that can panic/crash the node.

This PR:
- Adds `bug_condition!()`: we can use it to log errors that should never happen in runtime. We can still set `PANIC_ON_BUG=true` to force a panic.
- Replaces `From` implementations that were fallible (they shouldn't) by a `TryForm` version. Most P2P code is affected by this change because of conversions between our `PeerID` type and the one used by libp2p. This propagates upwards to most conversions.

### DIFF
--- a/core/src/log.rs
+++ b/core/src/log.rs
@@ -140,3 +140,16 @@ pub trait ActionEvent {
 use tracing::Value;
 
 pub use crate::{debug, error, info, trace, warn};
+
+#[macro_export]
+macro_rules! bug_condition {
+    ($($arg:tt)*) => {{
+        if std::env::var("OPENMINA_PANIC_ON_BUG")
+        .map(|v| v.to_lowercase() == "true")
+        .unwrap_or(false) {
+            panic!($($arg)*)
+        } else {
+            $crate::log::inner::error!($($arg)*)
+        }
+    }};
+}
--- a/node/src/p2p/p2p_effects.rs
+++ b/node/src/p2p/p2p_effects.rs
@@ -1,5 +1,6 @@
 use mina_p2p_messages::v2::{MinaLedgerSyncLedgerAnswerStableV2, StateHash};
 use openmina_core::block::BlockWithHash;
+use openmina_core::bug_condition;
 use p2p::channels::streaming_rpc::{
     P2pChannelsStreamingRpcAction, P2pStreamingRpcRequest, P2pStreamingRpcResponseFull,
 };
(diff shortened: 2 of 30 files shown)


core/src/log.rs
```diff
@@ -140,3 +140,16 @@ pub trait ActionEvent {
 use tracing::Value;
 
 pub use crate::{debug, error, info, trace, warn};
+
+#[macro_export]
+macro_rules! bug_condition {
+    ($($arg:tt)*) => {{
+        if std::env::var("PANIC_ON_BUG")
```
tizoc, line 147: I would add an `OPENMINA_` prefix here and make sure we always use that in our env vars

node/src/p2p/p2p_effects.rs
```diff
@@ -290,31 +291,44 @@ pub fn node_p2p_effects<S: Service>(store: &mut Store<S>, action: P2pActionWithM
...
-                        store.dispatch(TransitionFrontierSyncAction::BlocksPeerQueryError {
-                            peer_id,
-                            rpc_id: id,
-                            error: PeerBlockFetchError::Timeout,
-                        });
-                        store.dispatch(P2pDisconnectionAction::Init {
-                            peer_id,
-                            reason: P2pDisconnectionReason::TransitionFrontierRpcTimeout(rpc_kind),
-                        });
+                        if let Some(peer) = store.state().p2p.get_ready_peer(&peer_id) {
```
tizoc, line 294: what about

```rust
let Some(peer) = .... else {
    bug_condition!("get_ready_peer({:?}) returned None", peer_id);
    return;
};
```

to avoid the nesting? Diff will be smaller too

p2p/src/network/kad/bootstrap/p2p_network_kad_bootstrap_state.rs
```diff
@@ -33,7 +33,7 @@ impl P2pNetworkKadBootstrapState {
     pub fn new(key: PeerId) -> Self {
         P2pNetworkKadBootstrapState {
             key,
-            kademlia_key: key.into(),
+            kademlia_key: key.try_into().expect("valid key"), // TODO: propagate error
```
tizoc, line 36: Pending `TODO` here

tizoc (changes requested):

p2p/src/network/kad/bootstrap/p2p_network_kad_bootstrap_state.rs
dkuehr: Leaving this for next PR

p2p/src/network/select/p2p_network_select_effects.rs
```diff
@@ -92,104 +92,117 @@ impl P2pNetworkSelectAction {
...
-                                    }
-                                    StreamKind::Ping(PingAlgorithm::Ping1_0_0) => {
-                                        //unimplemented!()
-                                    }
-                                    StreamKind::Bitswap(_) => {
-                                        //unimplemented!()
-                                    }
-                                    StreamKind::Status(_) => {
-                                        //unimplemented!()
+                let Some(Some(negotiated)) = &state.negotiated else {
```
tizoc, line 112: much better this way, now the error is immediately after the failed condition, and not after the happy-path block

PR #648, opened 2024-09-06
Fix configure keypair

vlad9486:


cli/src/commands/node/mod.rs
```diff
@@ -73,6 +78,8 @@ pub struct Node {
     /// Snark fee, in Mina
     #[arg(long, env, default_value_t = 1_000_000)]
     pub snarker_fee: u64,
+    #[arg(env = "MINA_PRIVKEY_PASS")]
```
tizoc, line 81: @vlad9486 small change, lets move it above, after `producer_key` and rename to `producer_key_password` and `libp2p_password` to `libp2p_keypair_password` (so that it is `X_password`)

PR #662, opened 2024-09-09
P2p state machine porting

0xMimir:


p2p/src/network/kad/bootstrap/p2p_network_kad_bootstrap_reducer.rs
```diff
@@ -64,11 +72,19 @@ impl P2pNetworkKadBootstrapState {
                     .take(requests_to_create) // and stop when we create enough requests so up to 3 will be executed in parallel
                     .collect::<Vec<_>>();
 
+                if bootstrap_state.requests.is_empty() && peer_id_req_vec.is_empty() {
```
tizoc, line 75: I would rather have this happen later after `.into_dispatcher_and_state()` gets called (even if that makes us go through the next block of code and a loop that will be a no-op). It is easier to follow the code if there is a clear split between sate updates, and state machine flow control. With this shortcuiting here we have state update + state machine flow control code interleaved.

p2p/src/network/kad/bootstrap/p2p_network_kad_bootstrap_reducer.rs
```diff
@@ -77,31 +93,58 @@ impl P2pNetworkKadBootstrapState {
...
             P2pNetworkKadBootstrapAction::RequestDone {
                 peer_id,
                 closest_peers,
             } => {
-                let Some(req) = self.requests.remove(peer_id) else {
-                    return Err(format!("cannot find reques for peer {peer_id}"));
+                let state: &mut P2pNetworkKadBootstrapState = state_context.get_substate_mut()?;
+                let Some(req) = state.requests.remove(peer_id) else {
+                    if state.successful_requests < 20 {
+                        let dispatcher = state_context.into_dispatcher();
```
tizoc, line 127: Same as above, unless it complicates the code too much I would rather have the dispatching happening later. Also I see that the dispatching that we see here happens later too. Maybe moving the body of this block (the part that doesn't perform the dispatch) to a helper function will help (so that it can still short-circuit with a return)

tizoc: What I mean is:

```
let result = create_requests_logic(...); // Could even be a method on the state

// Dispatch
let dispatcher = ...;
// .. dispatch logic
```

p2p/src/network/kad/bootstrap/p2p_network_kad_bootstrap_reducer.rs
```diff
@@ -111,29 +154,45 @@ impl P2pNetworkKadBootstrapState {
...
-                Ok(())
+                result
             }
             P2pNetworkKadBootstrapAction::RequestError { peer_id, error } => {
-                let Some(req) = self.requests.remove(peer_id) else {
-                    return Err(format!("cannot find reques for peer {peer_id}"));
+                let bootstrap_state: &mut P2pNetworkKadBootstrapState =
+                    state_context.get_substate_mut()?;
+
+                let Some(req) = bootstrap_state.requests.remove(peer_id) else {
```
tizoc, line 173: Same as above, lets remove the interleaving of state-updates and control flow logic

tizoc (changes requested):
@0xMimir still have to go through the rest of the code, but I left some comments. Overall it is looking quite good.

p2p/src/network/identify/stream/p2p_network_identify_stream_reducer.rs
```diff
@@ -2,32 +2,70 @@ use super::{
...
+                peer_id, stream_id, ..
+            } => substate
+                .create_identify_stream_state(peer_id, stream_id)
+                .map_err(|stream| {
+                    format!("Identify stream already exists for action {action:?}: {stream:?}")
+                })?,
+            P2pNetworkIdentifyStreamAction::Prune {
+                peer_id, stream_id, ..
+            } => {
+                return substate
```
tizoc, line 42: @0xMimir one kind of subtle difference between the new-style reducers and the old is that short-circuiting the reducer will not run the code that used to run in the effect handler before. Not about this particular case here just a general comment for you to keep in mind.

p2p/src/network/identify/stream/p2p_network_identify_stream_reducer.rs
```diff
@@ -62,52 +115,173 @@ impl P2pNetworkIdentifyStreamState {
...
                         };
                         Ok(())
                     } else {
-                        self.handle_incoming_identify_message(len, data)
+                        stream_state.handle_incoming_identify_message(len, data)?;
+                        if let P2pNetworkIdentifyStreamState::IdentifyReceived { data } =
+                            stream_state
+                        {
+                            let data = data.clone();
+                            let dispatcher = state_context.into_dispatcher();
```
tizoc, line 129: If possible I would obtain the dispatcher only once (instead of having a separate call per branch) to use it later. The idea is to have clear separations between state updates and validations, and control flow (dispatches) phases.

tizoc: Alternatively just a comment clearly separating the two sections (I see you still need to access the state at this point, but it is not going to be mutated anymore, so we are done with the state updates phase by now)

p2p/tests/identify.rs
```diff
@@ -104,6 +104,7 @@ async fn rust_node_to_rust_node() -> anyhow::Result<()> {
 }
 
 #[tokio::test]
+#[ignore = "TODO: Add override for reducer"]
```
tizoc, line 107: What is this TODO about?

tizoc (approved):
@0xMimir LGTM, left some comments. There is also a TODO which I am not sure what it is about.

p2p/tests/identify.rs
0xMimir: This test and same named test in `p2p/tests/kademlia.rs` work by having override effects so that node can send invalid data, I need to implement this functionality for reducer.

p2p/src/network/kad/request/p2p_network_kad_request_reducer.rs
```diff
@@ -1,52 +1,292 @@
...
+                    },
+                    peer_id: *peer_id,
+                    stream_id: *stream_id,
+                });
+                dispatcher.push(P2pNetworkKadRequestAction::Prune { peer_id: *peer_id });
+                Ok(())
             }
             P2pNetworkKadRequestAction::Prune { .. } => {
-                return Err(String::from("should never happen"))
+                Err("Handled above shouldn't happen".to_owned())
```
tizoc, line 269: should be a `bug_condition!` ?

tizoc (commented):
adding a few comments, still going through the rest

p2p/src/network/kad/request/p2p_network_kad_request_reducer.rs
```diff
@@ -1,52 +1,292 @@
...
             P2pNetworkKadRequestAction::PeerIsConnecting { .. } => {
-                self.status = super::P2pNetworkKadRequestStatus::WaitingForConnection
+                request_state.status = P2pNetworkKadRequestStatus::WaitingForConnection;
+                Ok(())
+            }
+            P2pNetworkKadRequestAction::MuxReady { peer_id, addr } => {
+                let (dispatcher, state) = state_context.into_dispatcher_and_state();
+                let p2p_state: &P2pState = state.substate()?;
+
+                let stream_id = p2p_state
```
tizoc, line 132: This is a general comment, unrelated to the PR. We should start moving all these complicated state accessors (and mutators) to helpers (either as methods in the state itself or elsewhere, with proper documentation) to simplify and shorten the reducer functions. Ideally in the reducers we will just call a few well-named methods/functions that perform the actual work and then proceed with the dispatches.

p2p/src/network/kad/stream/p2p_network_kad_stream_reducer.rs
```diff
@@ -214,60 +403,135 @@ impl P2pNetworkKadOutgoingStreamState {
...
+                    });
+                    dispatcher.push(P2pNetworkKademliaAction::UpdateFindNodeRequest {
+                        addr: *addr,
+                        peer_id: *peer_id,
+                        stream_id: *stream_id,
+                        closest_peers,
+                    });
+                    Ok(())
+                } else {
+                    todo!();
```
tizoc, line 481: @0xMimir there is a  `todo!()` here. Is it something that can happen or is this a `bug_condition` ?

p2p/src/network/identify/stream/p2p_network_identify_stream_reducer.rs
```diff
@@ -62,52 +115,171 @@ impl P2pNetworkIdentifyStreamState {
...
-                        *self = P2pNetworkIdentifyStreamState::IncomingPartialData {
+                        *stream_state = P2pNetworkIdentifyStreamState::IncomingPartialData {
                             len,
                             data: data.to_vec(),
                         };
                         Ok(())
                     } else {
-                        self.handle_incoming_identify_message(len, data)
+                        stream_state.handle_incoming_identify_message(len, data)?;
+                        let stream_state = stream_state.clone();
```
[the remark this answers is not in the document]
0xMimir: I clone it here, since value inside enum needs to be cloned either way, to be passed to dispatcher

p2p/src/network/noise/p2p_network_noise_reducer.rs
```diff
@@ -81,30 +110,36 @@ impl P2pNetworkNoiseState {
...
                 }
-                self.buffer = self.buffer[offset..].to_vec();
+                noise_state.buffer = noise_state.buffer[offset..].to_vec();
+
+                let mut incoming = noise_state.incoming_chunks.clone();
+                let dispatcher = state_context.into_dispatcher();
+                while let Some(data) = incoming.pop_front() {
+                    dispatcher.push(P2pNetworkNoiseAction::IncomingChunk {
+                        addr: *addr,
+                        data: Data::from(data),
```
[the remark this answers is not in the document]
0xMimir: There are few parts where data is passed in action but not actually used, I am not sure why.

p2p/src/network/kad/bootstrap/p2p_network_kad_bootstrap_reducer.rs
```diff
@@ -42,20 +44,26 @@ fn prepare_next_request(
...
         filter_addrs: bool,
-    ) -> Result<(), String> {
+    ) -> Result<(), String>
+    where
+        State: SubstateAccess<Self> + SubstateAccess<P2pState> + SubstateAccess<P2pNetworkKadState>,
+        Action: crate::P2pActionTrait<State>,
+    {
         let (action, meta) = action.split();
+        let discovery_state: &P2pNetworkKadState = state_context.get_state().substate()?;
+        let routing_table = &discovery_state.routing_table;
```
tizoc, line 58: @0xMimir something we were discussing with @dkuehr is that `get_state()` shouldn't be used (it kind of violates the phase separation between state updates and control flow, I see it was introduced for some testing needs, but it was not part of the original API because of I just mentioned).

It would be better to pass these routing table from the caller here, would that complicate things too much?

p2p/src/network/kad/request/p2p_network_kad_request_reducer.rs
```diff
@@ -1,52 +1,293 @@
...
+                let dispatcher = state_context.into_dispatcher();
+                let data =
+                    P2pNetworkKademliaRpcRequest::find_node(key).map_err(|e| e.to_string())?;
+                dispatcher.push(P2pNetworkKademliaStreamAction::SendRequest {
+                    addr,
+                    peer_id,
+                    stream_id,
+                    data,
+                });
+                dispatcher.push(P2pNetworkKadRequestAction::RequestSent { peer_id });
```
tizoc, line 199: @0xMimir does this action depend on the previous one? if that is the case, then it should be handled with a callback.

Doesn't need to be fixed in this PR but should at least be a TODO for it.

tizoc: relevant PR where I already updated some code using callbacks to enforce the dependency https://github.com/openmina/openmina/pull/628/files

tizoc: main reason is that if for some reason `P2pNetworkKademliaStreamAction::SendRequest` doesn't get dispatched (either gets cancelled or filtered out by the enabling conditions), we don't want to dispatch `P2pNetworkKadRequestAction::RequestSent` either, since it is a dependent action.

tizoc: @0xMimir also https://github.com/openmina/openmina/blob/develop/ARCHITECTURE.md#callbacks-instead-of-conditional-dispatches

p2p/src/network/kad/stream/p2p_network_kad_stream_reducer.rs
```diff
@@ -1,126 +1,252 @@
...
+                match state {
+                    P2pNetworkKadIncomingStreamState::RequestIsReady {
+                        data: P2pNetworkKademliaRpcRequest::FindNode { key },
+                    } => {
+                        dispatcher.push(P2pNetworkKademliaStreamAction::WaitOutgoing {
+                            addr: *addr,
+                            peer_id: *peer_id,
+                            stream_id: *stream_id,
+                        });
+                        dispatcher.push(P2pNetworkKademliaAction::AnswerFindNodeRequest {
```
tizoc, line 103: @0xMimir another potential action-dependency that requires a callback

core/src/substate.rs
```diff
@@ -64,9 +64,10 @@ where
     }
 
     /// Obtain an immutable reference to the state.
-    /// 
+    ///
+    ///
```
tizoc, line 68: this empty comment-line needs to be removed or the linter will complain

p2p/src/network/kad/bootstrap/p2p_network_kad_bootstrap_reducer.rs
```diff
@@ -45,21 +45,23 @@ fn prepare_next_request(
...
     {
         let (action, meta) = action.split();
-        let discovery_state: &P2pNetworkKadState = state_context.get_state().substate()?;
-        let routing_table = &discovery_state.routing_table;
-        let bootstrap_state = state_context.get_substate()?;
 
         match action {
             P2pNetworkKadBootstrapAction::CreateRequests => {
+                let discovery_state: &P2pNetworkKadState =
+                    state_context.get_substate()?.substate()?;
```
tizoc, line 61: Oh I didn't notice the routing table was part of the substate, this is better then. I wonder why it was being passed from the caller before.

tizoc: nevermind, I see this is reducing over the parent state (that works too).

tizoc: One concern I have is that the way it is defiend now, allows for this reducer to mutate the outer state, which is not ideal (the code doesn't do that, but it is not obvious from just the signature).

tizoc: (btw, not a blocker, we can figure it out later)

0xMimir: This is also case for reducer on requests and streams.

p2p/src/network/identify/stream_effectful/p2p_network_identify_stream_effectful_actions.rs
```diff
@@ -0,0 +1,44 @@
...
+pub enum P2pNetworkIdentifyStreamEffectfulAction {
+    /// Creates a new stream state.
+    SendIdentify {
+        addr: ConnectionAddr,
+        peer_id: PeerId,
+        stream_id: StreamId,
+    },
+}
+
+macro_rules! enum_field {
```
dkuehr, line 17: Do we need this macro just for 3 fields?

p2p/src/network/kad/bootstrap/p2p_network_kad_bootstrap_reducer.rs
```diff
@@ -77,31 +89,58 @@ impl P2pNetworkKadBootstrapState {
...
             P2pNetworkKadBootstrapAction::RequestDone {
                 peer_id,
                 closest_peers,
             } => {
-                let Some(req) = self.requests.remove(peer_id) else {
-                    return Err(format!("cannot find reques for peer {peer_id}"));
+                let state: &mut P2pNetworkKadBootstrapState =
+                    state_context.get_substate_mut()?.substate_mut()?;
+                let Some(req) = state.requests.remove(peer_id) else {
+                    return Err(format!("cannot find request for peer {peer_id}"));
```
dkuehr, line 127: Is this a possible condition or bug_condition?

p2p/src/network/kad/bootstrap/p2p_network_kad_bootstrap_reducer.rs
```diff
@@ -111,29 +150,43 @@ impl P2pNetworkKadBootstrapState {
                             closest_peers: closest_peers.clone(),
                         },
                     );
+                    Ok(())
                 } else {
-                    return Err(format!("cannot find stats for request {req:?}"));
+                    Err(format!("cannot find stats for request {req:?}"))
```
dkuehr, line 155: possible or bug_condition?

p2p/src/network/kad/bootstrap/p2p_network_kad_bootstrap_reducer.rs
```diff
@@ -111,29 +150,43 @@ impl P2pNetworkKadBootstrapState {
...
+                result
             }
             P2pNetworkKadBootstrapAction::RequestError { peer_id, error } => {
-                let Some(req) = self.requests.remove(peer_id) else {
-                    return Err(format!("cannot find reques for peer {peer_id}"));
+                let bootstrap_state: &mut P2pNetworkKadBootstrapState =
+                    state_context.get_substate_mut()?.substate_mut()?;
+
+                let Some(req) = bootstrap_state.requests.remove(peer_id) else {
+                    return Err(format!("cannot find request for peer {peer_id}"));
```
dkuehr, line 170: possible or bug_condition?

p2p/src/network/kad/bootstrap/p2p_network_kad_bootstrap_reducer.rs
```diff
@@ -143,11 +196,15 @@ impl P2pNetworkKadBootstrapState {
                             error: error.clone(),
                         },
                     );
+
+                    Ok(())
                 } else {
-                    return Err(format!("cannot find stats for request {req:?}"));
-                }
+                    Err(format!("cannot find stats for request {req:?}"))
```
dkuehr, line 202: possible or bug_condition?

p2p/src/network/kad/request/p2p_network_kad_request_reducer.rs
```diff
@@ -1,52 +1,301 @@
...
+
+                let on_connection_in_progress = |dispatcher: &mut Dispatcher<Action, State>| {
+                    dispatcher.push(P2pNetworkKadRequestAction::PeerIsConnecting { peer_id });
+                    Ok(())
+                };
+
+                let on_connection_established = |dispatcher: &mut Dispatcher<Action, State>| {
+                    let Some((_, conn_state)) = p2p_state.network.scheduler.find_peer(&peer_id)
+                    else {
+                        return Err(format!(
```
dkuehr, line 73: possible or bug_condition?

p2p/src/network/identify/stream_effectful/p2p_network_identify_stream_effectful_actions.rs
0xMimir: It is not need, but I left it there just in case.

p2p/src/network/kad/request/p2p_network_kad_request_reducer.rs
```diff
@@ -1,52 +1,301 @@
...
+                    if let Some(stream_id) = conn_state.mux.as_ref().and_then(
+                        |P2pNetworkConnectionMuxState::Yamux(yamux)| {
+                            yamux.next_stream_id(
+                                crate::YamuxStreamKind::Kademlia,
+                                conn_state.incoming,
+                            )
+                        },
+                    ) {
+                        // multiplexing is ready, open a stream
+                        dispatcher.push(P2pNetworkYamuxAction::OpenStream {
```
dkuehr, line 91: Include TODO comment to reimplement this with callbacks

p2p/src/network/kad/request/p2p_network_kad_request_reducer.rs
```diff
@@ -1,52 +1,301 @@
...
+                            .map(|mux| (mux, conn.incoming))
+                            .ok_or_else(|| format!("multiplexing is not ready for {addr}"))
+                    })
+                    .and_then(|(P2pNetworkConnectionMuxState::Yamux(yamux), incoming)| {
+                        yamux
+                            .next_stream_id(crate::YamuxStreamKind::Kademlia, incoming)
+                            .ok_or_else(|| format!("cannot get next stream for {addr}"))
+                    })?;
+
+                dispatcher.push(P2pNetworkYamuxAction::OpenStream {
```
dkuehr, line 149: Include TODO comment to reimplement this with callbacks

p2p/src/network/kad/request/p2p_network_kad_request_reducer.rs
```diff
@@ -1,52 +1,301 @@
...
+                        bug_condition!(
+                            "P2pNetworkKadRequestAction::StreamReady invalid request key error: {error}"
+                        );
+                        return Ok(());
+                    }
+                };
+
                 let message = super::super::Message::from(&find_node);
-                self.status = quick_protobuf::serialize_into_vec(&message).map_or_else(
+                request_state.status = quick_protobuf::serialize_into_vec(&message).map_or_else(
```
dkuehr, line 183: bug_condition?

p2p/src/network/kad/stream/p2p_network_kad_stream_reducer.rs
```diff
@@ -1,126 +1,253 @@
...
-                        P2pNetworkStreamProtobufError::Limit(
-                            encoded_len,
-                            limits.kademlia_request(),
-                        ),
-                    );
-                    return Ok(());
-                }
+                            state.handle_incoming_request(encoded_len, remaining_data)?;
+                        } else {
+                            *state = P2pNetworkKadIncomingStreamState::Error(
```
dkuehr, line 79: bug_condition?

p2p/src/network/kad/stream/p2p_network_kad_stream_reducer.rs
```diff
@@ -1,126 +1,253 @@
...
-                self.handle_incoming_request(*len, &data)
+                state.handle_incoming_request(*len, &data)?;
+                let state = state.clone();
+                let dispatcher = state_context.into_dispatcher();
+
+                match state {
+                    P2pNetworkKadIncomingStreamState::RequestIsReady {
+                        data: P2pNetworkKademliaRpcRequest::FindNode { key },
+                    } => {
+                        dispatcher.push(P2pNetworkKademliaStreamAction::WaitOutgoing {
```
dkuehr, line 153: Not sure about this one but it is possible that we should use a callback

p2p/src/network/kad/stream/p2p_network_kad_stream_reducer.rs
```diff
@@ -152,122 +279,265 @@ impl P2pNetworkKadIncomingStreamState {
...
+    where
+        State: SubstateAccess<P2pNetworkKadState> + SubstateAccess<P2pState>,
+        Action: crate::P2pActionTrait<State>,
+    {
+        let (action, meta) = action.split();
+        let Some(P2pNetworkKadStreamState::Outgoing(state)) = state_context
+            .get_substate_mut()?
+            .find_kad_stream_state_mut(action.peer_id(), action.stream_id())
+        else {
+            return Err("invalid stream".to_owned());
```
dkuehr, line 296: bug_condition ?

dkuehr (approved):
Left some comments but in general LGTM

p2p/src/network/kad/request/p2p_network_kad_request_reducer.rs
0xMimir: `request_state.key` is our PeerId so this should be bug_condition with us

PR #699, opened 2024-09-12
Transition Frontier Sync: handle block application error and add a blacklist for failed blocks

binier:


tizoc:
@binier not exactly what you are fixing, but some time ago I added an error condition for when the block snark fails to verify, but never implemented the error logic:

https://github.com/openmina/openmina/blob/ed4e893a1a87fa3318f42b79a7a5b7d3ec5b7918/node/src/consensus/consensus_reducer.rs#L94-L96

binier:
@tizoc basically this is the handling for block application error.

tizoc:
@binier yes, I am just mentioning it because it is another case where we possibly want to add a peer to that blacklist, but it doesn't happen in the transition frontier (and the blacklist is there). Not sure if it should be moved up, or it should be two different blacklists (both are about blocks, but for different types of failures)

tizoc:
@binier no, it all sounds good to me, didn't know exactly which case you were trying to handle, just saw that it was potentially related to that currently-unhandled case for the block proof verification failure.

PR #701, opened 2024-09-13
More powerful clippy lints

adonagy:
Enable more clippy lints to deal with potential panics

### DIFF
--- a/node/src/block_producer/block_producer_reducer.rs
+++ b/node/src/block_producer/block_producer_reducer.rs
@@ -145,9 +145,10 @@ impl BlockProducerEnabled {
                 if let Some(won_slot) = state.current.won_slot() {
                     if let Some(chain) = best_chain.last().map(|best_tip| {
                         if best_tip.global_slot() == won_slot.global_slot() {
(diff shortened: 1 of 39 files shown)


tizoc:
if possible I would skip build scripts too in addition to tests

node/src/block_producer/block_producer_reducer.rs
```diff
@@ -545,7 +557,9 @@ impl BlockProducerEnabled {
             &global_slot_since_genesis,
         );
         let consensus_state = v2::ConsensusProofOfStakeDataConsensusStateValueStableV2 {
-            blockchain_length: v2::UnsignedExtendedUInt32StableV1((pred_block.height() + 1).into()),
+            blockchain_length: v2::UnsignedExtendedUInt32StableV1(
+                (pred_block.height().saturating_add(1)).into(),
```
tizoc, line 561: @dkuehr I am thinking if cases like these shouldn't be fatal failures. If this happens the state will be messed up and we cannot trust it, continuing as if nothing happened will just delay the error until later making things more confusing.

tizoc (approved):

PR #705, opened 2024-09-13
fix: Index staged ledgers by staged ledger hash

tizoc:


tizoc:
closing in favor of #706

PR #713, opened 2024-09-16
RPC layer and Select protocol layer port

0xMimir:
RPC layer and Select protocol layer port


p2p/src/network/noise/p2p_network_noise_actions.rs
```diff
@@ -75,7 +75,16 @@ impl From<P2pNetworkNoiseAction> for crate::P2pAction {
 }
 
 impl redux::EnablingCondition<P2pState> for P2pNetworkNoiseAction {
-    fn is_enabled(&self, _state: &P2pState, _time: redux::Timestamp) -> bool {
+    fn is_enabled(&self, state: &P2pState, _time: redux::Timestamp) -> bool {
+        let Some(_noise_state) = state
+            .network
+            .scheduler
+            .connection_state(self.addr())
+            .and_then(|state| state.noise_state())
```
dkuehr, line 83: maybe `.is_none()` instead of else?

0xMimir: I left it as `_state` just so it is easier to add state later but I can change it to `.is_none`

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
```diff
@@ -1,33 +1,83 @@
...
+                addr,
+                peer_id,
+                stream_id,
+            } => {
+                rpc_state.buffer.extend_from_slice(data);
                 let mut offset = 0;
                 // TODO(akoptelov): there shouldn't be the case where we have multiple incoming messages at once (or at least other than heartbeat)
                 loop {
-                    let buf = &self.buffer[offset..];
+                    let buf = &rpc_state.buffer[offset..];
```
dkuehr, line 75: Lets take the opportunity to change indexing for `get(offset..)` and `bug_condition` for the `None` case

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
```diff
@@ -48,53 +98,233 @@ impl P2pNetworkRpcState {
...
                                 }
                             };
-                            self.incoming.push_back(msg);
+                            rpc_state.incoming.push_back(msg);
                             continue;
                         }
                     }
 
+                    if offset != 0 {
+                        rpc_state.buffer = rpc_state.buffer[offset..].to_vec();
```
dkuehr, line 112: Same here

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
```diff
@@ -1,33 +1,83 @@
...
 
 use self::p2p_network_rpc_state::P2pNetworkRpcError;
 
 use super::*;
 
 impl P2pNetworkRpcState {
-    pub fn reducer(
-        &mut self,
+    pub fn reducer<State, Action>(
+        mut state_context: Substate<Action, State, P2pNetworkState>,
```
tizoc, line 32: @0xMimir when `Substate` is not over `Self` please add a comment mentioning what other substate is accessed, just for future reference once we figure out what is the best way to narrow it.

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
```diff
@@ -1,33 +1,83 @@
...
-                self.is_incoming = *incoming;
+    ) -> Result<(), String>
+    where
+        State: crate::P2pStateTrait,
+        Action: crate::P2pActionTrait<State>,
+    {
+        let rpc_state = state_context
+            .get_substate_mut()?
+            .find_rpc_state_mut(action.action())
+            .ok_or_else(|| format!("RPC state not found for action: {:?}", action.action()))?;
```
tizoc, line 43: should this be a bug_condition? I would assume that the enabling condition should filter the action so that it doesn't get dispatched if the state doesn't exist, right?

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
```diff
@@ -48,53 +98,233 @@ impl P2pNetworkRpcState {
...
+
+                match &message {
+                    RpcMessage::Handshake => {
+                        if !state.is_incoming {
+                            dispatcher.push(P2pChannelsRpcAction::Ready { peer_id: *peer_id });
+                        }
+                    }
+                    RpcMessage::Heartbeat => {}
+                    RpcMessage::Query { header, bytes } => {
+                        if let Err(e) = rpc_query_effects(*peer_id, header, bytes, dispatcher) {
```
dkuehr, line 173: It could be a bit confusing to have a function named `rpc_query_effects` called by the reducer

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
tizoc, line 75: this whole loop makes me a bit nervous, can it be abstracted?

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
tizoc: looking at the rest of the code I am still not sure why this cannot be done over `Self` btw, is that really required?

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
```diff
@@ -48,53 +98,233 @@ impl P2pNetworkRpcState {
...
+                            }
+                        };
+                        // unset pending
+                        dispatcher.push(P2pNetworkRpcAction::PrunePending {
+                            peer_id: *peer_id,
+                            stream_id: *stream_id,
+                        });
+
+                        if let Err(e) =
+                            rpc_response_effects(*peer_id, &query_header, bytes, dispatcher)
```
tizoc, line 204: Just in case, the logic inside this functions does not depend on the `PrunePending` above having been processed already right?

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
```diff
@@ -135,3 +365,206 @@ impl P2pNetworkRpcState {
         }
     }
 }
+
+fn rpc_query_effects<'a, State, Action>(
```
dkuehr, line 369: Rename to replace "effects" part

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
```diff
@@ -135,3 +365,206 @@ impl P2pNetworkRpcState {
...
+                id,
+                request: Box::new(P2pRpcRequest::InitialPeers),
+            });
+        }
+        (name, version) => return Err(RpcQueryError::Unimplemented(name, version)),
+    }
+    Ok(())
+}
+
+fn rpc_response_effects<State, Action>(
```
dkuehr, line 441: Same here

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
tizoc: nevermind, it doesn't receive as input the state object, so there is no dependency

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
tizoc: ok I see, this is what you mentioned on Slack right? regarding the selection of the substate based on the kind of action (which here happens when calling `find_rpc_state_mut`)

tizoc: But even in that case.... I think you can still use `Substate<...., Self>` ? it is just that you will have to select the substate in the caller

p2p/src/network/yamux/p2p_network_yamux_actions.rs
```diff
@@ -58,24 +58,23 @@ impl From<P2pNetworkYamuxAction> for crate::P2pAction {
...
 
 impl redux::EnablingCondition<P2pState> for P2pNetworkYamuxAction {
-    fn is_enabled(&self, _state: &P2pState, _time: redux::Timestamp) -> bool {
-        #[allow(unused_variables)]
+    fn is_enabled(&self, state: &P2pState, _time: redux::Timestamp) -> bool {
+        let Some(_yamux_state) = state
+            .network
+            .scheduler
+            .connection_state(self.addr())
+            .and_then(|state| state.yamux_state())
```
dkuehr, line 66: maybe use `is_none()` and avoid the `else`

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
tizoc: true, when it comes to logic that just makes more dispatches we should move away from the "effects" terminology and reserve that for interactions with the "external world" (services).

@dkuehr any suggestion for a better naming?

dkuehr (commented):
LGTM.
Btw, I see some inconsistent use of `error!` vs `openmina_core::error!` maybe due moving code from different files?

tizoc (approved):
LGTM. Left a bunch of comments (none is a blocker, just good-to-have)

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
0xMimir: It doesn't depend on `PrunePending`, but  `PrunePending` action is used twice only to set `pending` to `None` this can be move to reducers.

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
dkuehr: I'm not really sure, I remember there was some naming conventions in redux for these cases (an action that produces more actions) but I think we should stay away from it.

p2p/src/network/rpc/p2p_network_rpc_reducer.rs
0xMimir: It should only iterate if there is data in `rpc_state.buffer` by default it breaks out of it

PR #718, opened 2024-09-17
Ported scheduler

0xMimir:


p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
```diff
@@ -1,54 +1,122 @@
...
             }
             P2pNetworkSchedulerAction::InterfaceExpired { ip, .. } => {
-                self.interfaces.remove(ip);
+                scheduler_state.interfaces.remove(ip);
+                Ok(())
             }
             P2pNetworkSchedulerAction::ListenerReady { listener } => {
-                self.listeners.insert(*listener);
+                scheduler_state.listeners.insert(*listener);
+                Ok(())
```
tizoc, line 57: do you need `Ok(())` everywhere? instead of for example a single `Ok(())` at the end, and any case that produces an error would short-circuit everything with a `return`.

0xMimir: I don't need `Ok(())` for every action but I think it is better to know that nothing runs after that.

p2p/src/network/scheduler_effectful/p2p_network_scheduler_effectful_effects.rs
```diff
@@ -0,0 +1,82 @@
...
+                if state.network.scheduler.connections.len()
+                    >= state.config.limits.max_connections()
+                {
+                    store.service().send_mio_cmd(MioCmd::Refuse(listener));
+                } else {
+                    store.service().send_mio_cmd(MioCmd::Accept(listener));
+                }
+            }
+            P2pNetworkSchedulerEffectfulAction::IncomingDidAccept { addr, .. } => {
+                let nonce = store.service().generate_random_nonce();
```
tizoc, line 31: This is for later (out of scope for this PR), but we will need to split this eventually. Any result that is not deterministic will need to "re-enter" the state machine through an event. So eventually we will be able to implement replays with fully mocked services by replaying the events (since all non-deterministic results will be contained there).

p2p/src/network/pnet/p2p_network_pnet_reducer.rs
```diff
@@ -34,21 +40,92 @@ impl Half {
...
+
+                let (dispatcher, state) = state_context.into_dispatcher_and_state();
+                let scheduler_state: &P2pNetworkSchedulerState = state.substate()?;
+                let pnet_state = &scheduler_state
+                    .connection_state(addr)
+                    .ok_or_else(|| format!("Missing connection for action: {:?}", action))?
+                    .pnet;
+
+                let Half::Done { to_send, .. } = &pnet_state.incoming else {
+                    return Ok(());
```
dkuehr, line 71: Can we clarify under what conditions we can reach other states here?
Should this be filtered out by an enabling condition?

p2p/src/network/pnet/p2p_network_pnet_reducer.rs
```diff
@@ -34,21 +40,92 @@ impl Half {
...
+
+                let (dispatcher, state) = state_context.into_dispatcher_and_state();
+                let scheduler_state: &P2pNetworkSchedulerState = state.substate()?;
+                let pnet_state = &scheduler_state
+                    .connection_state(addr)
+                    .ok_or_else(|| format!("Missing connection for action: {:?}", action))?
+                    .pnet;
+
+                let Half::Done { to_send, .. } = &pnet_state.outgoing else {
+                    return Ok(());
```
dkuehr, line 96: Same as above

p2p/src/network/pnet/p2p_network_pnet_reducer.rs
```diff
@@ -34,21 +40,92 @@ impl Half {
...
+                Ok(())
             }
-            P2pNetworkPnetAction::SetupNonce { nonce, .. } => {
-                self.outgoing.reduce(&self.shared_secret, nonce)
+            P2pNetworkPnetAction::SetupNonce {
+                nonce,
+                addr,
+                incoming,
+            } => {
+                pnet_state.outgoing.reduce(&pnet_state.shared_secret, nonce);
```
dkuehr, line 112: Should we add state checks and bug_condition?

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
```diff
@@ -64,148 +132,480 @@ impl P2pNetworkSchedulerState {
...
+                    }
+                    Err(error) => {
+                        let Some((peer_id, peer_state)) = p2p_state.peer_with_connection(*addr)
+                        else {
+                            bug_condition!(
+                                "outgoing connection to {addr} failed, but there is no peer for it"
+                            );
+                            return Ok(());
+                        };
+                        if matches!(
```
dkuehr, line 161: Not sure if needed but we could add en else clause here with a bug_condition

p2p/src/network/pnet/p2p_network_pnet_reducer.rs
0xMimir: If incoming data is small other state can be reached, it is not bug and it is normal for other incoming to be in `Half::Buffering`

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
```diff
@@ -64,148 +132,480 @@ impl P2pNetworkSchedulerState {
...
+                        }
+                    }
+                }
+                Ok(())
             }
-            P2pNetworkSchedulerAction::IncomingDataIsReady { .. } => {}
             P2pNetworkSchedulerAction::IncomingDataDidReceive { result, addr } => {
-                let Some(state) = self.connections.get_mut(addr) else {
-                    return;
+                if let Some(state) = scheduler_state.connections.get_mut(addr) {
```
dkuehr, line 175: Maybe add some comment explaining this part since in the original reducer code there was an early return but now that the "effects" code was appended (which is always executed) it is a bit more difficult to follow

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
```diff
@@ -64,148 +132,480 @@ impl P2pNetworkSchedulerState {
...
                 addr,
                 kind,
                 protocol,
                 incoming,
                 expected_peer_id,
-                ..
             } => {
-                let Some(connection) = self.connections.get_mut(addr) else {
-                    return;
+                if let Some(connection) = scheduler_state.connections.get_mut(addr) {
```
dkuehr, line 203: General observation: now that "effects" and reducer code are together, what happens when we can't find connection state for some address? Does still make sense to execute the "effects" part and dispatch further actions? Is it a bug condition not being able to find associated state to an address?

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
```diff
@@ -64,148 +132,480 @@ impl P2pNetworkSchedulerState {
...
+                            }
+                        }
+                    }
+                    None => {
+                        match &select_kind {
+                            SelectKind::Authentication => {
+                                // TODO: close the connection
+                            }
+                            SelectKind::MultiplexingNoPeerId => {
+                                // WARNING: must not happen
```
dkuehr, line 353: add bug_condition here?

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
```diff
@@ -64,148 +132,480 @@ impl P2pNetworkSchedulerState {
...
-                ..
+                peer_id,
             } => {
-                if let Some(cn) = self.connections.get_mut(addr) {
-                    if let Some(P2pNetworkConnectionMuxState::Yamux(yamux)) = &mut cn.mux {
-                        yamux.init = true;
-                        yamux.message_size_limit = *message_size_limit;
-                    }
+                let Some(cn) = scheduler_state.connections.get_mut(addr) else {
+                    return Ok(());
```
dkuehr, line 423: Can this happen? Filter it in enabling condition?

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
0xMimir: In most cases it doesn't make sense to execute "effects" if reducer is unable to be executed, but there are cases in which we need to execute "effects" even if reducer fails, most of these cases are to dispatch error, so just in case I try to port the reducer so even if it can't execute to still go to effects

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
```diff
@@ -64,148 +132,480 @@ impl P2pNetworkSchedulerState {
...
+                let peer_id = *peer_id;
+                let addr = *addr;
+
+                if incoming {
+                    dispatcher.push(P2pConnectionIncomingAction::Libp2pReceived { peer_id });
+                } else {
+                    dispatcher.push(P2pConnectionOutgoingAction::FinalizeSuccess { peer_id });
+                }
+
+                // for each negotiated yamux conenction open a new outgoing RPC stream
```
dkuehr, line 441: Now that I see this comment I wonder if we should do this after receiving the identity message from the peer (which should include Rpc in the list of supported protocols).

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
```diff
@@ -64,148 +132,480 @@ impl P2pNetworkSchedulerState {
...
+
+                if kad_state.map_or(false, |state| state.request(&peer_id).is_some()) {
+                    dispatcher.push(P2pNetworkKadRequestAction::MuxReady { peer_id, addr });
                 }
+                Ok(())
             }
             P2pNetworkSchedulerAction::Disconnect { addr, reason } => {
-                let Some(conn_state) = self.connections.get_mut(addr) else {
+                let Some(conn_state) = scheduler_state.connections.get_mut(addr) else {
                     error!(meta.time(); "P2pNetworkSchedulerAction::Disconnect: connection {addr} does not exist");
```
dkuehr, line 469: Are we still seeing these kind of errors? I think they are bug conditions

p2p/src/network/pnet/p2p_network_pnet_reducer.rs
tizoc: Please add a small comment mentioning that

p2p/src/network/scheduler_effectful/p2p_network_scheduler_effectful_effects.rs
dkuehr: We have a Crypto service we could use, but we also had some problems with the use of entropy in the p2p layer to keep things working with the network debugger. Maybe @vlad9486 can add some insight.

p2p/src/network/scheduler_effectful/p2p_network_scheduler_effectful_effects.rs
```diff
@@ -0,0 +1,82 @@
...
+use super::{super::*, *};
+
+impl P2pNetworkSchedulerEffectfulAction {
+    pub fn effects<Store, S>(self, _meta: &ActionMeta, store: &mut Store)
+    where
+        Store: crate::P2pStore<S>,
+        Store::Service: P2pMioService + P2pCryptoService,
+    {
+        match self {
+            P2pNetworkSchedulerEffectfulAction::InterfaceDetected { ip, port } => {
```
dkuehr, line 15: Now that I'm noting this, do we want to listen on any detected interface? Do we really want this behavior? @vlad9486

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
tizoc: could also be simplified by moving the logic to a method in `scheduler_state`

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
tizoc: such cases where the reducer cannot be executed and the dispatches must not happen either should be filtered by enabling conditions right?

tizoc: btw this whole block of code here is another good candidate for being moved to a helper

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
tizoc: and bug_condition in the else it is not supposed to happen

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
0xMimir: Since both actions that are dispatcher in "effects" part require connection state, and enabling conditions require connection to exists I will change it so that if connection state is not found to not run anything, and just add comment

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
0xMimir: I have moved both parts to helpers, it looks much better and cleaner now

p2p/src/network/scheduler_effectful/p2p_network_scheduler_effectful_effects.rs
vlad9486: @dkuehr May be useful on mobile devices where cellular and wifi are turned on/off during runtime.

PR #729, opened 2024-09-19
Keep state consistent during disconnection

vlad9486:
Delay removal of the connection from the network state machine until the disconnection actually occurs at the TCP level.

### DIFF
--- a/node/src/event_source/event_source_effects.rs
+++ b/node/src/event_source/event_source_effects.rs
@@ -115,6 +115,17 @@ pub fn event_source_effects<S: Service>(store: &mut Store<S>, action: EventSourc
                             });
                         }
                     }
+                    MioEvent::ConnectionDidCloseOnDemand(addr) => {
+                        // enabling condition
+                        // it could be the case that the connection initialized again, so should not remove the state
+                        let exist = store.state().p2p.ready().unwrap().peers.iter().any(|(_, x)| match &x.dial_opts {
+                            Some(p2p::connection::outgoing::P2pConnectionOutgoingInitOpts::LibP2P(o)) => o.matches_socket_addr(addr.sock_addr),
+                            _ => false,
+                        });
+                        if !exist {
+                            store.dispatch(P2pNetworkSchedulerAction::Prune { addr });
+                        }
+                    }
                 },
                 P2pEvent::Connection(e) => match e {
                     P2pConnectionEvent::OfferSdpReady(peer_id, res) => match res {
--- a/node/src/event_source/event_source_effects.rs
+++ b/node/src/event_source/event_source_effects.rs
@@ -118,11 +118,13 @@ pub fn event_source_effects<S: Service>(store: &mut Store<S>, action: EventSourc
                     MioEvent::ConnectionDidCloseOnDemand(addr) => {
                         // enabling condition
                         // it could be the case that the connection initialized again, so should not remove the state
-                        let exist = store.state().p2p.ready().unwrap().peers.iter().any(|(_, x)| match &x.dial_opts {
-                            Some(p2p::connection::outgoing::P2pConnectionOutgoingInitOpts::LibP2P(o)) => o.matches_socket_addr(addr.sock_addr),
-                            _ => false,
+                        let ready = store.state().p2p.ready().map_or(false, |ready| {
+                            ready.peers.iter().any(|(_, state)| state.status.as_ready().is_some() && match &state.dial_opts {
+                                Some(p2p::connection::outgoing::P2pConnectionOutgoingInitOpts::LibP2P(o)) => o.matches_socket_ip(addr.sock_addr),
+                                _ => false,
+                            })
                         });
-                        if !exist {
+                        if !ready {
                             store.dispatch(P2pNetworkSchedulerAction::Prune { addr });
                         }
                     }
--- a/node/src/event_source/event_source_effects.rs
+++ b/node/src/event_source/event_source_effects.rs
@@ -116,17 +116,7 @@ pub fn event_source_effects<S: Service>(store: &mut Store<S>, action: EventSourc
                         }
                     }
                     MioEvent::ConnectionDidCloseOnDemand(addr) => {
-                        // enabling condition
-                        // it could be the case that the connection initialized again, so should not remove the state
-                        let ready = store.state().p2p.ready().map_or(false, |ready| {
-                            ready.peers.iter().any(|(_, state)| state.status.as_ready().is_some() && match &state.dial_opts {
-                                Some(p2p::connection::outgoing::P2pConnectionOutgoingInitOpts::LibP2P(o)) => o.matches_socket_ip(addr.sock_addr),
-                                _ => false,
-                            })
-                        });
-                        if !ready {
-                            store.dispatch(P2pNetworkSchedulerAction::Prune { addr });
-                        }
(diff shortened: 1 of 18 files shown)


p2p/src/disconnection/p2p_disconnection_actions.rs
```diff
@@ -21,8 +21,16 @@ pub enum P2pDisconnectionAction {
 impl redux::EnablingCondition<P2pState> for P2pDisconnectionAction {
     fn is_enabled(&self, state: &P2pState, _time: redux::Timestamp) -> bool {
         match self {
-            P2pDisconnectionAction::Init { peer_id, .. }
-            | P2pDisconnectionAction::Finish { peer_id } => {
+            P2pDisconnectionAction::Init { peer_id, .. } => {
+                state.peers.get(peer_id).map_or(false, |peer| {
+                    !matches!(peer.status, P2pPeerStatus::Disconnected { .. })
+                }) && state.network.scheduler.connections.iter().any(|(_, st)| {
```
binier, line 27: You are expecting all peers to be libp2p, but webrtc peers won't have libp2p state so it will break that.
Why do you need this extra check anyways? If peer is not in disconnected state, but is in some other state, we need to enable this action.

binier (changes requested):

p2p/testing/src/redux.rs
```diff
@@ -228,6 +228,9 @@ pub(super) fn event_effect(store: &mut crate::redux::Store, event: P2pEvent) ->
                     )
                 }
             }
+            MioEvent::ConnectionDidCloseOnDemand(addr) => {
+                SubStore::dispatch(store, P2pNetworkSchedulerAction::Prune { addr })
```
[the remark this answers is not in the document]
binier: In fact I'd put `SubStore::dispatch(store, P2pNetworkSchedulerAction::Prune { addr })` action in effects of the `P2pDisconnectionAction::Finish`

p2p/src/disconnection/p2p_disconnection_actions.rs
vlad9486: Without this check, I see timeout disconnects repeated many times before the actual disconnect happens. Is it better to add a `P2pPeerStatus::Disconnecting` state to indicate that a disconnect is pending?

binier: Why does it take that long to disconnect a peer after disconnection was initiated though?

p2p/src/network/scheduler/p2p_network_scheduler_reducer.rs
```diff
@@ -374,7 +374,6 @@ impl P2pNetworkSchedulerState {
                 let state: &P2pState = state.substate()?;
 
                 let peer_with_state = state.peer_with_connection(*addr);
-                dispatcher.push(P2pNetworkSchedulerAction::Prune { addr: *addr });
```
0xMimir, line 377: Why is `P2pNetworkSchedulerAction::Prune` dispatch removed, it seems to be a reason why p2p tests are failing

vlad9486: This is postponed. We remove the connection after mio responds that the connection is closed, not immediately.

p2p/src/disconnection/p2p_disconnection_actions.rs
vlad9486: It takes microseconds. The problem was that the peer was not marked as disconnected because the connection removal is postponed and [this](https://github.com/openmina/openmina/blob/4b4dadfb2c381cc46205c172a313050636264e0f/p2p/src/disconnection/p2p_disconnection_reducer.rs#L68) check was triggered. I fixed it by the condition `conn_state.peer_id() == Some(peer_id) && conn_state.closed.is_none()`.

p2p/testing/src/redux.rs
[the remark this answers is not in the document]
vlad9486: It is too big change and require more debugging to make it work properly. Let's allow connection where `conn_state.closed` is `Some(..)` to exist after `Finish` done.

p2p/src/disconnection/p2p_disconnection_actions.rs
vlad9486: I removed this from enabling conditions. WebRTC should be ok.

p2p/src/disconnection/p2p_disconnection_reducer.rs
```diff
@@ -22,27 +22,32 @@ impl P2pDisconnectedState {
...
+                        };
+                        peer.status = P2pPeerStatus::Disconnecting { time: meta.time() };
+
+                        let dispatcher = state_context.into_dispatcher();
                         dispatcher.push(P2pNetworkSchedulerAction::Disconnect {
-                            addr: *addr,
+                            addr,
                             reason: reason.clone(),
                         });
                         dispatcher.push(P2pDisconnectionAction::Finish { peer_id: *peer_id });
```
binier, line 45: If it gets insta finished, without any event from mio, then disconnecting state seems useless.

binier (changes requested):

node/src/event_source/event_source_effects.rs
```diff
@@ -115,6 +115,17 @@ pub fn event_source_effects<S: Service>(store: &mut Store<S>, action: EventSourc
                             });
                         }
                     }
+                    MioEvent::ConnectionDidCloseOnDemand(addr) => {
+                        // enabling condition
+                        // it could be the case that the connection initialized again, so should not remove the state
+                        let exist = store.state().p2p.ready().unwrap().peers.iter().any(|(_, x)| match &x.dial_opts {
+                            Some(p2p::connection::outgoing::P2pConnectionOutgoingInitOpts::LibP2P(o)) => o.matches_socket_addr(addr.sock_addr),
```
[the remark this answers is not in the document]
binier: To properly fix this, it might require bigger changes, so maybe this is fine for now.

node/src/event_source/event_source_effects.rs
```diff
@@ -115,6 +115,17 @@ pub fn event_source_effects<S: Service>(store: &mut Store<S>, action: EventSourc
...
+                            _ => false,
+                        });
+                        if !exist {
```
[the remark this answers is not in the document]
vlad9486: I should rename `exist` -> `ready`. When we removing connection, the peer should be in `Disconnecting` state, not `Ready(..)`. In webrtc case `MioEvent` will not happen, so it is irrelevant.

To properly handle the case when new connection is initializing during old connection is still `Disconnecting`, maybe we need a counter in the state or in the key, in this map:
```
    pub peers: BTreeMap<PeerId, P2pPeerState>,
```
but anyway, disconnection is a millisecond at worst, and reconnection timeout is much bigger, so it should not be possible.

p2p/src/disconnection/p2p_disconnection_reducer.rs
vlad9486: Let's keep it for further improvement. I will make `Finish` an effect of `Prune`, but not now.

node/src/event_source/event_source_effects.rs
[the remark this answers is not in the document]
vlad9486: Oh, I just notice the unwrap, I remove it now.

binier (approved):

PR #735, opened 2024-09-23
Ported peer reducer and connection reducer

0xMimir:


node/src/p2p/p2p_effects.rs
```diff
@@ -474,12 +464,7 @@ pub fn node_p2p_effects<S: Service>(store: &mut Store<S>, action: P2pActionWithM
                                     work: snark.clone(),
                                 });
                             }
-                            Some(P2pRpcResponse::InitialPeers(peers)) => {
-                                store.dispatch(P2pDiscoveryAction::Success {
-                                    peer_id,
-                                    peers: peers.iter().cloned().collect(),
-                                });
-                            }
+                            Some(P2pRpcResponse::InitialPeers(_)) => {}
```
dkuehr, line 467: Do we still need this?

0xMimir: I think it is not need now, but it might be needed by webnode later, @binier is it so?

p2p/src/connection/outgoing_effectful/p2p_connection_outgoing_effectful_effects.rs
```diff
@@ -0,0 +1,67 @@
...
+impl P2pConnectionOutgoingEffectfulAction {
+    pub fn effects<Store, S>(self, _meta: &ActionMeta, store: &mut Store)
+    where
+        Store: crate::P2pStore<S>,
+        Store::Service: P2pConnectionService,
+    {
+        match self {
+            P2pConnectionOutgoingEffectfulAction::RandomInit => {
+                let peers = store.state().disconnected_peers().collect::<Vec<_>>();
+                let picked_peer = store.service().random_pick(&peers);
```
dkuehr, line 24: I was wondering what happens if peers is empty and just noted that `random_pick` unwraps. We probably need a better way to handle this

0xMimir: List of peers not being empty is currently handled by enabling conditions, I could make `random_pick` return `Option` to remove unwrap and add `bug_condition` if it is empty.

dkuehr (approved):
LGTM

p2p/src/connection/incoming/p2p_connection_incoming_reducer.rs
```diff
@@ -1,162 +1,533 @@
...
 use super::{
-    P2pConnectionIncomingAction, P2pConnectionIncomingActionWithMetaRef, P2pConnectionIncomingState,
+    super::{incoming::P2pConnectionIncomingState, RejectionReason},
+    IncomingSignalingMethod, P2pConnectionIncomingAction,
 };
 
 impl P2pConnectionIncomingState {
-    pub fn reducer(&mut self, action: P2pConnectionIncomingActionWithMetaRef<'_>) {
+    pub fn reducer<Action, State>(
+        mut state_context: Substate<Action, State, P2pState>,
```
tizoc, line 26: Please add a comment here explaining why this reducer is over `P2pState` instead of `Self`

p2p/src/connection/outgoing_effectful/p2p_connection_outgoing_effectful_actions.rs
```diff
@@ -0,0 +1,48 @@
...
+        peer_id: PeerId,
+        answer: Box<webrtc::Answer>,
+    },
+}
+
+impl redux::EnablingCondition<P2pState> for P2pConnectionOutgoingEffectfulAction {
+    fn is_enabled(&self, state: &P2pState, _: redux::Timestamp) -> bool {
+        match self {
+            P2pConnectionOutgoingEffectfulAction::RandomInit => {
+                !state.already_has_min_peers() && state.disconnected_peers().next().is_some()
```
tizoc, line 37: @0xMimir I don't think this should be here right? it should be in a stateful action that will then trigger this `RandomInit` effectful action. Effectful actions should always be enabled and should not be checking the state (the fact that they have enabling condition is just an artifact of the full refactoring still not being complete, in the final version they will not have such thing)

p2p/src/identify/p2p_identify_reducer.rs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        let (action, _meta) = action.split();
+        let p2p_state = state_context.get_substate_mut()?;
+
+        match action {
+            P2pIdentifyAction::NewRequest { addr, .. } => {
+                let (dispatcher, state) = state_context.into_dispatcher_and_state();
+                let p2p_state: &P2pState = state.substate()?;
+                let scheduler = &p2p_state.network.scheduler;
+                let stream_id = scheduler
```
tizoc, line 29: question about these short-circuiting accessors, should these be bug conditions? if yes, I would remove the `?` at the end, perform the check manually so that `bug_condition` can be called, and then at that point yes, the error can be returned.

NOTE: I am assuming the caller is not doing any special handling with the error, if it does ignore this comment

p2p/src/p2p_reducer.rs
```diff
@@ -1,149 +1,70 @@
...
+                P2pDisconnectedState::reducer(state_context, meta.with_action(action))
             }
+            P2pAction::Peer(action) => P2pPeerState::reducer(
+                Substate::from_compatible_substate(state_context),
+                meta.with_action(action),
+            ),
             P2pAction::Channels(action) => {
                 let Some(peer_id) = action.peer_id() else {
-                    return;
+                    return Ok(());
```
tizoc, line 39: I assume these are not `bug_condition`s and expected

p2p/src/peer/p2p_peer_reducer.rs
```diff
@@ -1,41 +1,98 @@
...
+                            dispatcher.push(P2pChannelsStreamingRpcAction::Init { peer_id });
+                        }
+                    }
+                }
+
+                Ok(())
+            }
+            P2pPeerAction::BestTipUpdate { peer_id, best_tip } => {
+                let Some(peer) = p2p_state.get_ready_peer_mut(peer_id) else {
+                    return Ok(());
```
tizoc, line 90: I assume this is not a `bug_condition` either but just in case I am mentioning it.

p2p/src/connection/outgoing/p2p_connection_outgoing_actions.rs
```diff
@@ -14,9 +14,6 @@ pub type P2pConnectionOutgoingActionWithMetaRef<'a> =
 #[derive(Serialize, Deserialize, Debug, Clone, ActionEvent)]
 #[action_event(fields(display(opts), display(peer_id), display(error)))]
 pub enum P2pConnectionOutgoingAction {
-    /// Initialize connection to a random peer.
-    #[action_event(level = trace)]
-    RandomInit,
```
tizoc, line 19: @0xMimir re `RandomInit` in the effectful action, this should still be here, not removed. Then it should implement the enabling condition currently in the effectful action, and when enabled, dispatch the effectful counterpart.

tizoc (changes requested):
Left some comments, but the important one is the `RandomInit` one, after that is solved LGTM

p2p/src/identify/p2p_identify_reducer.rs
0xMimir: In most cases yes, they should be bug conditions expecally for `P2pState` but if they happen we will know, because nothing would work, since root reducer from node waits for `P2pState` to be ready before anything is action is passed there.

p2p/src/p2p_reducer.rs
0xMimir: Yes, there was one like this for connections but I updated and removed it

tizoc (approved):
LGTM

PR #742, opened 2024-09-23
Changed log levels

0xMimir:


ledger/src/transaction_pool.rs
```diff
@@ -2012,10 +2012,7 @@ impl TransactionPool {
                         account.liquid_balance_at_slot(global_slot_since_genesis),
                     ) {
                         Ok(x) => Ok(x),
-                        Err(e) => {
-                            eprintln!();
```
tizoc, line 2016: wonder what this was even about

PR #743, opened 2024-09-23
fix(block-producer): Use the correct staking ledger after epoch boundaries

tizoc:
If the epoch of the block being producer is greater than the epoch of the parent block, then the next epoch ledger must be used.

### DIFF
--- a/node/src/block_producer/block_producer_effects.rs
+++ b/node/src/block_producer/block_producer_effects.rs
@@ -151,6 +151,16 @@ pub fn block_producer_effects<S: crate::Service>(
                 .collect();
             // TODO(binier)
             let supercharge_coinbase = true;
+            // We want to know if this is a new epoch to decide which staking ledger to use
+            // (staking epoch ledger or next epoch ledger).
+            let is_new_epoch = won_slot.epoch()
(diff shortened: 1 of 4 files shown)


node/src/block_producer/block_producer_effects.rs
```diff
@@ -151,6 +151,16 @@ pub fn block_producer_effects<S: crate::Service>(
                 .collect();
             // TODO(binier)
             let supercharge_coinbase = true;
+            // We want to know if this is a new epoch to decide which staking ledger to use
+            // (staking epoch ledger or next epoch ledger).
+            let is_new_epoch = won_slot.epoch()
```
tizoc, line 156: I am not entirely sure about these comparisons here, these slot/epoch values confuse me (they are are already confusing in the OCaml node, and in our case I think we have even more variants).

I am assuming the `WonSlot` values and the `epoch_count` in the consensus state is compatible.

@adonagy do you know if this is correct?

tizoc: ok just confirmed that both use the same point of reference (genesis vs hardfork) and it is ok to compare them.

PR #749, opened 2024-09-25
fix(vrf-evaluator) Improve readiness check logic

tizoc:
We don't need to keep track of the last heigh of the previous epoch. If the epoch of the best tip and the block at the root of the transition frontier match, we know that the snarked ledger at the root belongs to a block that has been finalized already because there are 290 blocks between it and the best tip.


tizoc:
related to #700

PR #759, opened 2024-09-30
Ported channels

0xMimir:
Ported channels from old style reducers to new style reducers


node/src/p2p/p2p_effects.rs
```diff
@@ -231,28 +229,22 @@ pub fn node_p2p_effects<S: Service>(store: &mut Store<S>, action: P2pActionWithM
...
+                P2pChannelsSnarkAction::Libp2pReceived { peer_id, snark, .. } => {
+                    store.dispatch(SnarkPoolCandidateAction::WorkReceived {
+                        peer_id,
+                        work: *snark,
+                    });
+                }
+                _ => {}
+            },
             P2pChannelsAction::SnarkJobCommitment(action) => {
-                // TODO: does the order matter here? if not this clone can be removed
```
tizoc, line 254: @0xMimir I guess I will figure out by reading the rest of the diff, but I was wondering where did these `effects` calls go

0xMimir: Effects that don't require service are in reducer and ones that do are in `P2pChannelsSnarkJobCommitmentEffectfulAction::effects`.

p2p/src/channels/p2p_channels_reducer.rs
```diff
@@ -1,30 +1,210 @@
...
+        State: crate::P2pStateTrait,
+        Action: crate::P2pActionTrait<State>,
+    {
+        let (action, meta) = action.split();
+        let time = meta.time();
+
+        let peer_id = action.peer_id;
+        let chain_id = action.message.channel_id();
+
+        let mut is_enabled = |action: Action| dispatcher.push_if_enabled(action, state, time);
```
tizoc, line 76: I have to look at it more thoroughly, but be careful with this. `push_is_enabled` will return a result based on the state at the time the action was queued, which may not be the same as the state when the action is finally dispatched.

p2p/src/channels/snark/p2p_channels_snark_reducer.rs
```diff
@@ -1,63 +1,119 @@
...
+            .peer_id()
+            .and_then(|peer_id| p2p_state.get_ready_peer_mut(peer_id))
+            .map(|peer_state| &mut peer_state.channels.snark)
+            .ok_or_else(|| format!("Invalid state for: {action:?}"));
+
         match action {
-            P2pChannelsSnarkAction::Init { .. } => {
-                *self = Self::Init { time: meta.time() };
+            P2pChannelsSnarkAction::Init { peer_id } => {
+                let state = state?;
```
tizoc, line 31: Are these bug-conditions? will they be correctly processed as such?

0xMimir: These are bug conditions but I didn't add them in places like this, I will added them just in case

p2p/src/channels/rpc/p2p_channels_rpc_reducer.rs
```diff
@@ -31,35 +58,96 @@ impl P2pChannelsRpcState {
...
                     time: meta.time(),
                     id: *id,
                     request: request.clone(),
                 };
+
+                let dispatcher = state_context.into_dispatcher();
+
+                #[cfg(feature = "p2p-libp2p")]
+                if is_libp2p {
+                    if let Some((query, data)) =
```
dkuehr, line 87: Under what conditions is this None?

p2p/src/channels/rpc/p2p_channels_rpc_reducer.rs
```diff
@@ -31,35 +58,96 @@ impl P2pChannelsRpcState {
...
+                    return Ok(());
                 };
                 *local = P2pRpcLocalState::Responded {
                     time: meta.time(),
                     id: *id,
                     request: std::mem::take(request),
                 };
+
+                let dispatcher = state_context.into_dispatcher();
+                if let Some(P2pRpcResponse::BestTipWithProof(resp)) = response.as_deref() {
```
dkuehr, line 134: What other kind of responses can we get and how should be handled?

p2p/src/channels/snark_effectful/p2p_channels_snark_effectful_actions.rs
```diff
@@ -0,0 +1,31 @@
...
+
+impl redux::EnablingCondition<P2pState> for P2pChannelsSnarkEffectfulAction {
+    fn is_enabled(&self, _state: &P2pState, _time: redux::Timestamp) -> bool {
+        true
+    }
+}
+
+impl From<P2pChannelsSnarkEffectfulAction> for crate::P2pAction {
+    fn from(action: P2pChannelsSnarkEffectfulAction) -> Self {
+        Self::ChannelsEffectful(P2pChannelsEffectfulAction::Snark(action))
```
dkuehr, line 29: Can we change `Self` here for the action type?

p2p/src/channels/snark_job_commitment_effectful/p2p_channels_snark_job_commitment_effectful_actions.rs
```diff
@@ -0,0 +1,32 @@
...
+
+impl redux::EnablingCondition<P2pState> for P2pChannelsSnarkJobCommitmentEffectfulAction {
+    fn is_enabled(&self, _state: &P2pState, _time: redux::Timestamp) -> bool {
+        true
+    }
+}
+
+impl From<P2pChannelsSnarkJobCommitmentEffectfulAction> for crate::P2pAction {
+    fn from(action: P2pChannelsSnarkJobCommitmentEffectfulAction) -> Self {
+        Self::ChannelsEffectful(P2pChannelsEffectfulAction::SnarkJobCommitment(action))
```
dkuehr, line 30: Can we change `Self` for `P2pAction`?

p2p/src/channels/rpc/p2p_channels_rpc_reducer.rs
0xMimir: Only if request is `P2pRpcRequest::Snark`

p2p/src/channels/streaming_rpc/p2p_channels_streaming_rpc_reducer.rs
```diff
@@ -154,16 +221,42 @@ impl P2pChannelsStreamingRpcState {
...
+                });
+                Ok(())
             }
-            P2pChannelsStreamingRpcAction::ResponsePartNextSend { .. } => {}
-            P2pChannelsStreamingRpcAction::ResponsePartSend { .. } => {
+            P2pChannelsStreamingRpcAction::ResponsePartNextSend { id, .. } => {
+                let (dispatcher, state) = state_context.into_dispatcher_and_state();
+                let state: &P2pState = state.substate()?;
+
+                let Some(response) = None.or_else(|| {
```
dkuehr, line 237: This looks a bit weird

dkuehr (approved):
LGTM

p2p/src/channels/rpc/p2p_channels_rpc_reducer.rs
0xMimir: It can be `None` or other variants of `P2pRpcResponse`, they are handle in `node_p2p_effects`

p2p/src/channels/snark_effectful/p2p_channels_snark_effectful_actions.rs
tizoc: @0xMimir one clarification, the problem is not the `Self` in the return, is the `Self` in the body. But I don't think it is an issue here, it is just in dispatchers and effect handlers (when `Self::` or aliases are used it makes it harder to grep for the action). I would revert to the original.

@dkuehr agree?

tizoc (approved):
left a comment regarding the `Self` change, I would revert that.

LGTM

p2p/src/channels/snark_effectful/p2p_channels_snark_effectful_actions.rs
dkuehr: Yes I meant in the return, but is not really an issue here. It just makes things easier to grep

PR #779, opened 2024-10-07
Fix subtract with overflow in p2p_network_yamux_reducer.rs

olton:
Fix panic 
thread 'main' panicked at /openmina/p2p/src/network/yamux/p2p_network_yamux_reducer.rs:211:33: attempt to subtract with overflow. Also, same problem in line 339.


vlad9486:
Of course, please, don't forget `.wrapping_add()` on line 425 of the same file. So it will be something like this:

```
stream.window_theirs = stream.window_theirs.wrapping_sub(data.len() as u32);
``` 

and 

```
*window = (*window).wrapping_add(increasing);
```

olton:
I have no objections. I agree.

PR #784, opened 2024-10-08
Ported reducer and p2p effects in node

0xMimir:


node/src/effects.rs
```diff
@@ -37,8 +36,6 @@ pub fn effects<S: Service>(store: &mut Store<S>, action: ActionWithMeta) {
             store.dispatch(ExternalSnarkWorkerAction::Start);
 
             if store.state().p2p.ready().is_some() {
-                p2p_timeout_effects(store, &meta);
```
tizoc, line 40: @0xMimir what replaces this?

0xMimir: This is replaced by `P2pState::p2p_timeout_dispatch` in reducer

tizoc: Ok, I see it in the handler of `Action::CheckTimeouts`

p2p/src/p2p_state.rs
```diff
@@ -491,38 +491,56 @@ type OptionalCallback<T> = Option<Callback<T>>;
...
 
+    /// Callback for [`P2pDisconnectionAction::Finish`]
     pub on_p2p_disconnection_finish: OptionalCallback<PeerId>,
 
-    pub on_p2p_connection_outgoing_error:
-        OptionalCallback<(RequestId<RpcIdType>, P2pConnectionOutgoingError)>,
-    pub on_p2p_connection_outgoing_success: OptionalCallback<RequestId<RpcIdType>>,
+    /// Callback for [`P2pConnectionOutgoingAction::Error`]
+    pub on_p2p_connection_outgoing_error: OptionalCallback<(RpcId, P2pConnectionOutgoingError)>,
+    /// Callback for [`P2pConnectionOutgoingAction::Success`]
```
tizoc, line 514: @0xMimir that would be `P2pConnectionOutgoingAction::Init` right? Basically, this callback will be called as a result of the `Init` action.

So on the caller which now has:

```rust
store.dispatch(P2pConnectionOutgoingAction::Init {
    opts,
    rpc_id: Some(rpc_id),
});
```

if we were to add the callbacks, we would have:

```rust
store.dispatch(P2pConnectionOutgoingAction::Init {
    opts,
    rpc_id: Some(rpc_id),
    on_success: ....this callback....,
    on_error: ...the error callback...,
});

**Review by dkuehr — APPROVED** [review]
LGTM. Just one question are the callbacks optional just to support `Default` or is there any other reason?

File: node/src/p2p/callbacks/p2p_callbacks_actions.rs [hunk 3]
```diff
@@ -0,0 +1,56 @@
...
+        rpc::{P2pRpcId, P2pRpcRequest, P2pRpcResponse},
+        streaming_rpc::P2pStreamingRpcResponseFull,
+    },
+    PeerId,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, Clone, ActionEvent)]
+#[action_event(level = debug)]
+pub enum P2pCallbacksAction {
```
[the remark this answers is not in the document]
0xMimir: Currently they are only RPC related, but they don't have to be

node/src/p2p/callbacks/p2p_callbacks_actions.rs
```diff
@@ -0,0 +1,56 @@
...
+        peer_id: PeerId,
+        id: P2pRpcId,
+        response: Option<P2pStreamingRpcResponseFull>,
+    },
+}
+
+impl redux::EnablingCondition<crate::State> for P2pCallbacksAction {
+    fn is_enabled(&self, _state: &crate::State, _time: redux::Timestamp) -> bool {
+        match self {
+            P2pCallbacksAction::P2pChannelsRpcReady { .. } => true,
```
tizoc, line 47: I think these callbacks (that replace the interception of an action) are like effectful actions in the sense that they will always be enabled right? so the implementation here could be just `true`

node/src/p2p/callbacks/p2p_callbacks_reducer.rs
```diff
@@ -0,0 +1,384 @@
...
+                        });
+                    }
+                    Some(P2pRpcResponse::BestTipWithProof(resp)) => {
+                        let (body_hashes, root_block) = &resp.proof;
+
+                        let (Ok(best_tip), Ok(root_block)) = (
+                            BlockWithHash::try_new(resp.best_tip.clone()),
+                            BlockWithHash::try_new(root_block.clone()),
+                        ) else {
+                            openmina_core::error!(meta.time(); "P2pRpcResponse::BestTipWithProof: invalid blocks");
```
tizoc, line 128: this can be replaced by `bug_condition!` right?

tizoc: Nevermind, this is not a case we want to panic when the flag is enabled, it is just that we got bad data in the response.

node/src/p2p/callbacks/p2p_callbacks_reducer.rs
```diff
@@ -0,0 +1,384 @@
...
+                    }
+                    Some(P2pRpcResponse::Snark(snark)) => {
+                        dispatcher.push(SnarkPoolCandidateAction::WorkReceived {
+                            peer_id,
+                            work: snark.clone(),
+                        });
+                    }
+                    Some(P2pRpcResponse::InitialPeers(_)) => {}
+                }
+                dispatcher.push(TransitionFrontierSyncLedgerSnarkedAction::PeersQuery);
```
tizoc, line 231: Not an issue with the PR, the old version already had this bug. But in the cases in which the flow is short-circuited by a `return` because of an error, these dispatches will never happen.

tizoc: Cannot come up with an obvious solution right now, but lets at least add a `FIXME` comment here mentioning that these will not be dispatched in the case of errors.

tizoc: Well, actually, there is an obvious solution. The body of these match branch is too long, the handling could be moved to an outer helper function/method, then the `return` will not be an issue.

node/src/p2p/callbacks/p2p_callbacks_reducer.rs
```diff
@@ -0,0 +1,384 @@
...
+                dispatcher.push(TransitionFrontierSyncLedgerStagedAction::PartsPeerFetchInit {});
```
tizoc, line 232: the `{}` part can be removed

node/src/p2p/callbacks/p2p_callbacks_reducer.rs
```diff
@@ -0,0 +1,384 @@
...
+                        dispatcher.push(
+                            TransitionFrontierSyncLedgerStagedAction::PartsPeerFetchSuccess {
+                                peer_id,
+                                rpc_id,
+                                parts: parts.clone(),
+                            },
+                        );
+                    }
+                }
+                dispatcher.push(TransitionFrontierSyncLedgerStagedAction::PartsPeerFetchInit {});
```
tizoc, line 380: same here, the `{}` is not necessary

node/src/rpc/rpc_effects.rs
```diff
@@ -279,6 +279,17 @@ pub fn rpc_effects<S: Service>(store: &mut Store<S>, action: RpcActionWithMeta)
             }
         }
         RpcAction::P2pConnectionIncomingPending { .. } => {}
+        RpcAction::P2pConnectionIncomingAnswerReady {
```
tizoc, line 282: This is a case of a new action that is required only because you need to dispatch two different actions from the callback?

p2p/tests/kademlia.rs
```diff
@@ -402,7 +402,7 @@ fn bad_node_effects(
                 event_mapper_effect(store, a);
             }
             Action::Idle(_) => {
-                p2p_timeout_effects(store, &meta);
+                // p2p_timeout_effects(store, &meta);
```
tizoc, line 405: Is this a leftover comment? or something that should be enabled?

p2p/tests/identify.rs
```diff
@@ -268,7 +268,7 @@ fn bad_node_effects(
                 event_mapper_effect(store, a);
             }
             Action::Idle(_) => {
-                p2p_timeout_effects(store, &meta);
+                // p2p_timeout_effects(store, &meta);
```
tizoc, line 271: Leftover comment or required?

node/src/transition_frontier/sync/ledger/snarked/transition_frontier_sync_ledger_snarked_reducer.rs
```diff
@@ -601,6 +611,88 @@ impl TransitionFrontierSyncLedgerSnarkedState {
                     target: target.clone(),
                 };
             }
+            TransitionFrontierSyncLedgerSnarkedAction::P2pDisconnection { peer_id } => {
```
tizoc, line 614: I am not very convinced about this one. It was originally part of the `node/p2p` module and the logic made sense there (because it is the upper layer connecting everything), but now this logic is inside the transition frontier module, doesn't feel right. It is also interacting with the snark pool candidate, which has no place here. 

I think it is better to move it with the other callbacks.

node/src/transition_frontier/transition_frontier_actions.rs
```diff
@@ -69,6 +74,9 @@ impl redux::EnablingCondition<crate::State> for TransitionFrontierAction {
                             .map_or(false, |s| s.is_apply_error()),
                     }
             }
+            TransitionFrontierAction::RpcRespondBestTip { .. } => {
+                state.transition_frontier.best_tip().is_some()
```
tizoc, line 78: what happens if we don't have a best tip? we just never answer? doesn't feel right. This is a bug in the original code, not introduced in the PR, but just now noticed it.

node/src/transition_frontier/transition_frontier_reducer.rs
```diff
@@ -101,6 +103,18 @@ impl TransitionFrontierState {
                 }
                 state.sync = TransitionFrontierSyncState::Synced { time: meta.time() };
             }
+            TransitionFrontierAction::RpcRespondBestTip { peer_id } => {
```
tizoc, line 106: Same as the other action that was added to the transition frontier, I don't think this belongs here, but instead with the other p2p handler callbacks.

tizoc (changes requested):
Added a bunch of comments. Some things are probably out of scope to the PR but it would be good to at least get a TODO/FIXME comment for those so that we don't forget later. Same for adding callbacks to the actions directly associated to them, that is for another iteration (comments mentioning the associated actions is good enough for now, unless you find that it is easy enough to do in this PR).

I would like to have the new actions added to the transition frontier moved out from there and into the p2p related callbacks, they feel out of place in the transition frontier and introduce more inter-component dependencies, which we can reduce.

node/src/p2p/callbacks/p2p_callbacks_actions.rs
0xMimir: Yes, but I wrote it out just in case some of them needed it

node/src/rpc/rpc_effects.rs
0xMimir: Yes, but I can remove it, set callback to `RpcAction::P2pConnectionIncomingRespond` and dispatch `P2pConnectionIncomingAction::AnswerSendSuccess` after callback in p2p crate

p2p/tests/kademlia.rs
0xMimir: I need to rewrite override functionality in p2p testing framework so I just left it here

p2p/tests/identify.rs
0xMimir: I need to update this and kademlia bad node tests after I update p2p testing framework

node/src/rpc/rpc_effects.rs
tizoc: No, it is fine, was just asking to make sure I was interpreting it correctly.

p2p/src/p2p_state.rs
tizoc: @0xMimir I see this still contains the previous comments, so let me clarify. The idea is that it should mention the original action to which it relates (from the point of view of the caller). Like in the example above, the caller will never dispatch `Success` or `Error`, he will dispatch `Init`, so this is a callback for `Init`. But it is no big deal, just a nice-to-have to make it easier to modification I mentioned above in the future.

tizoc (approved):
LGTM

PR #788, opened 2024-10-09
feat(transition-frontier): Verify blocks as soon as possible during catchup

tizoc:


tizoc:
Closing this for now, not going with this approach.

Issue #772, opened 2024-10-02
As a developer, I need to implement WebRTC peer discovery and signaling via connected common peer


binier:
- [ ] WebRTC signaling via common peer. 
- [ ] Peer discovery on WebRTC transport.

binier:
- [x] Implement state machine for relaying offers and answers and for receiving relayed ones.
- [ ] Implement discovery state machine where peers will be discovered and offer relaying will be initiated.
- [ ] Handle disconnections/timeouts.

PR #794, opened 2024-10-10
P2p WebRTC: peer discovery and p2p based signaling

binier:
closes: #772


binier:
I still need to add timeouts, some error handling and do general review, but it's much better than nothing so I'll merge it once CI passes

PR #806, opened 2024-10-15
Updated reducer to accept action instead of & action

0xMimir:
Update reducer to accept action instead of reference to action, this is to make p2p layer ready for full stateful vs effetcful split between actions.

The cloning of an action shouldn't impact the stack size, since the size of `P2pAction` is only 184 bytes. In most cases, it needs to be either fully or partially cloned to update the state and/or pass data to new actions.

### DIFF
--- a/p2p/src/peer/p2p_peer_reducer.rs
+++ b/p2p/src/peer/p2p_peer_reducer.rs
@@ -9,7 +9,7 @@ impl P2pPeerState {
     /// Substate is accessed
     pub fn reducer<Action, State>(
         mut state_context: Substate<Action, State, P2pState>,
-        action: ActionWithMeta<&P2pPeerAction>,
+        action: ActionWithMeta<P2pPeerAction>,
     ) -> Result<(), String>
     where
         State: crate::P2pStateTrait,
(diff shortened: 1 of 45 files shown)


dkuehr (commented):
Could you add a description of why this change is needed?
Have you measured this change has a big impact in the amount of stack used?

p2p/src/peer/p2p_peer_reducer.rs
```diff
@@ -23,41 +23,41 @@ impl P2pPeerState {
...
+                        dial_opts: None,
                         identify: None,
                         status: P2pPeerStatus::Disconnected {
                             time: Timestamp::ZERO,
                         },
                     });
 
                 if let Some(dial_opts) = dial_opts {
-                    peer_state.dial_opts.get_or_insert(dial_opts.clone());
+                    let _ = peer_state.dial_opts.insert(dial_opts);
```
tizoc, line 37: I am confused about what is going on here (more about the old version than the new one).

tizoc (approved):
Left a few comments, but LGTM in general. Having said that, I would leave these kind of refactors for later, when we can get an actual benefit from the change, I don't think we get a noticeable improvement from these changes right now (but I may be missing something).

tizoc:
I am even worried that it may introduce many conflicts to https://github.com/openmina/openmina/pull/794, so I would wait for that one first before merging this PR @0xMimir to avoid slowing it down.

p2p/src/peer/p2p_peer_reducer.rs
0xMimir: I have changed logic in this one, in old one `dial_opts` would only be current if peer didn't exist in peers already or if it didn't have `dial_opts`

tizoc:
@0xMimir out of curiosity, how complicated was the rebase after #794 got merged?

PR #817, opened 2024-10-17
Introduce a queue that allows to respect peer window size

vlad9486:


p2p/src/network/yamux/p2p_network_yamux_state.rs
```diff
@@ -182,6 +184,35 @@ impl YamuxFrame {
...
         vec
     }
+
+    /// If this data is bigger then `pos`, keep only first `pos` bytes and return some remaining
+    /// otherwise return none
+    pub fn split_at(&mut self, pos: usize) -> Option<Self> {
+        use std::ops::Sub;
+
+        if let YamuxFrameInner::Data(data) = &mut self.inner {
+            if data.len() == pos {
```
vlad9486, line 194: Maybe here should be `if data.len() <= pos`

p2p/src/network/yamux/p2p_network_yamux_reducer.rs
```diff
@@ -318,21 +349,45 @@ impl P2pNetworkYamuxState {
                             });
                         }
                     }
+                    YamuxFrameInner::WindowUpdate { .. } => {
+                        while let Some(frame) = pending_outgoing.pop_front() {
```
dkuehr, line 353: Is there any bound or practical limit in how many elements we can have in `pending_outgoing`?
I'm concerned that someone could try to fill the dispatch queue and and cause blockings in the state machine

p2p/src/network/yamux/p2p_network_yamux_reducer.rs
```diff
@@ -211,11 +214,37 @@ impl P2pNetworkYamuxState {
...
                         YamuxFrameInner::WindowUpdate { difference } => {
-                            yamux_state
+                            let stream = yamux_state
                                 .streams
                                 .entry(frame.stream_id)
-                                .or_insert_with(YamuxStreamState::incoming)
-                                .update_window(false, difference);
+                                .or_insert_with(YamuxStreamState::incoming);
+                            stream.update_window(false, difference);
+                            if difference > 0 {
```
tizoc, line 222: this can be 0 or negative? in which cases does that happen?

p2p/src/network/yamux/p2p_network_yamux_state.rs
```diff
@@ -54,6 +54,7 @@ pub struct YamuxStreamState {
     pub writable: bool,
     pub window_theirs: u32,
     pub window_ours: u32,
+    pub pending: VecDeque<YamuxFrame>,
```
dkuehr, line 57: We should make this bounded

dkuehr (commented):
Please add a more detailed description of the problem(s) this PR solves.
Also, it would be nice to do some testing of if with our p2p fuzzer since the change is not trivial.

p2p/src/network/yamux/p2p_network_yamux_reducer.rs
```diff
@@ -318,21 +349,45 @@ impl P2pNetworkYamuxState {
...
+                            if let Some(remaining) = frame.split_at(stream.window_theirs as usize) {
+                                stream.pending.push_back(remaining);
+                            }
+                            // the window will be zero after sending
+                            stream.window_theirs = 0;
+                        } else {
+                            // either the window cannot accept any byte,
+                            // or the queue is already not empty
+                            // in both cases the whole frame goes in the queue and nothing to send
+                            stream.pending.push_back(frame);
```
vlad9486, line 388: Need to add a limit here. Malicious peer can keep its window small and send an RPC request in a loop.

p2p/src/network/yamux/p2p_network_yamux_state.rs
vlad9486: Yes, I missed this

vlad9486:
@dkuehr 
Without this change, our implementation doesn't respect the yamux protocol. It sends the peer more than the peer allows to send. This doesn't cause a real problem, because the OCaml node doesn't penalise us for this. But Juraj wants us to use this yamux limit in the future to prioritise block production over other tasks (bootstrap and kademlia) and to support slow hardware (raspberry pi). So it will be necessary that our implementation respects it.

p2p/src/network/yamux/p2p_network_yamux_reducer.rs
tizoc: ok, please add a `bug_condition!` call in the else branch so that if this ever happens for some reason we will know about it. That is something we want to do in general, to avoid such unexpected (unenforced) cases happening silently without us noticing

vlad9486: I add a runtime error, not bug_condition, because peer can send us this and we should not crash.

tizoc: You are right, this is based on external input, so it is not exactly a `bug_condition!`. Btw `bug_condition` only panics if a flag is enabled in the environment, otherwise it just logs an error (but again this specific case is not for it)

p2p/src/network/yamux/p2p_network_yamux_reducer.rs
```diff
@@ -318,21 +344,61 @@ impl P2pNetworkYamuxState {
                             });
                         }
                     }
+                    YamuxFrameInner::WindowUpdate { difference } => {
+                        if *difference < 0 {
```
tizoc, line 348: @vlad9486  what about `0` ? that case is not covered. The other condition uses `> 0` and this one `< 0`, which means there is no handling for `0`

vlad9486: Yes, initially I wrote `<=`, but apparently the peer sends window update with zero if it wants to open the stream but not send data, so it sends flag SYN. Of course it could send data with zero length body, but it sends window update.

PR #824, opened 2024-10-24
Correcting epoch progress

garethtdavies:
CleanShot 2024-10-24 at 09 10 19@2x

Currently, this stat doesn't show the progress of the current epoch. This fix displays the percentage of the current epoch completed.


directcuteo:
Nice catch. Looks good!

PR #833, opened 2024-10-28
Ported RPC layer in node to new style reducers

0xMimir:


node/src/rpc/rpc_reducer.rs
```diff
@@ -1,114 +1,395 @@
...
+                let dispatcher = state_context.into_dispatcher();
+                dispatcher.push(P2pConnectionOutgoingAction::Init {
+                    opts: opts.clone(),
+                    rpc_id: Some(*rpc_id),
+                });
+                dispatcher.push(RpcAction::P2pConnectionOutgoingPending { rpc_id: *rpc_id });
             }
             RpcAction::P2pConnectionOutgoingPending { rpc_id } => {
-                let Some(rpc) = self.requests.get_mut(rpc_id) else {
+                let Some(rpc) = state.requests.get_mut(rpc_id) else {
```
tizoc, line 85: these should be `bug_condition!` right? The enabling condition should already be filtering out the action if the entry for that `rpc_id` cannot be found.

node/src/rpc/rpc_reducer.rs
```diff
@@ -1,114 +1,395 @@
...
+                        });
+                        dispatcher.push(RpcAction::ScanStateSummaryGetSuccess {
+                            rpc_id: *rpc_id,
+                            scan_state: Ok(Vec::new()),
+                        });
+                        return;
+                    }
+                };
+
+                // TODO: add callbacks here
```
tizoc, line 235: May be a good time to solve this `TODO`. Pattern is the same as the `on_init` callback in other actions. See `Callbacks instead of conditional dispatches` in the architecture document.

node/src/ledger/ledger_effects.rs
```diff
@@ -219,13 +233,27 @@ fn next_read_requests_init<S: redux::Service>(store: &mut Store<S>) {
...
+        store.dispatch(LedgerReadAction::Init {
+            request,
+            callback: LedgerReadInitCallback::P2pChannelsResponsePending
+         {      callback: redux::callback!(on_ledger_read_init_p2p_channels_response_pending((is_streaming: bool, id: P2pRpcId, peer_id: PeerId)) -> crate::Action{
+                    if is_streaming {
+                        P2pAction::from(P2pChannelsStreamingRpcAction::ResponsePending {
+                            peer_id,
+                            id,
+                        })
+                                    } else {
```
tizoc, line 246: @0xMimir indentation is weird here

0xMimir: Fixed, I am not sure why fmt formatted it that way.

tizoc: It doesn't handle code inside macros very well.

node/src/rpc/rpc_reducer.rs
```diff
@@ -118,60 +433,232 @@ impl RpcState {
...
                 };
-                self.requests.insert(*rpc_id, rpc_state);
+                state.requests.insert(*rpc_id, rpc_state);
+
+                let dispatcher = state_context.into_dispatcher();
+                dispatcher.push(RpcAction::TransactionInjectPending { rpc_id: *rpc_id });
+
+                // sort the commands by nonce
+
+                // let Ok(commands) = commands
```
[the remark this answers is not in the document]
0xMimir: I have left it just in case it's a todo, comment about sorting commands, but I can remove it

tizoc:
for the failed tests you will need to rebase btw, the fix is on develop

tizoc (approved):
LGTM

PR #849, opened 2024-10-30
Ported block producer and VRF evaluator

0xMimir:


node/src/block_producer/vrf_evaluator/block_producer_vrf_evaluator_reducer.rs
```diff
@@ -1,26 +1,48 @@
...
         let (action, meta) = action.split();
         match action {
             BlockProducerVrfEvaluatorAction::EvaluateSlot { vrf_input } => {
-                self.status = BlockProducerVrfEvaluatorStatus::SlotEvaluationPending {
+                state.status = BlockProducerVrfEvaluatorStatus::SlotEvaluationPending {
                     time: meta.time(),
                     global_slot: vrf_input.global_slot,
                 };
+
+                state_context.into_dispatcher().push(
```
tizoc, line 33: lets `let dispatcher = ...;` for now even if it is just for a single action like everywhere else.

If the case is too common we can add a helper method to the context later.

node/src/block_producer/vrf_evaluator/block_producer_vrf_evaluator_reducer.rs
```diff
@@ -164,7 +314,40 @@ impl BlockProducerVrfEvaluatorState {
...
+                        best_tip_epoch,
+                        best_tip_slot,
+                        best_tip_global_slot,
+                        next_epoch_first_slot,
+                        staking_epoch_data,
+                        ..
+                    },
+                )) = get_slot_and_status()
+                else {
+                    // error here!
```
tizoc, line 339: `bug_condition` here and in the return above?

node/src/block_producer/block_producer_reducer.rs
```diff
@@ -12,68 +12,104 @@ use mina_p2p_messages::{
...
     }
 }
 
 impl BlockProducerEnabled {
-    pub fn reducer(
-        &mut self,
-        action: BlockProducerActionWithMetaRef<'_>,
-        best_chain: &[AppliedBlock],
-    ) {
+    pub fn reducer(mut state_context: Substate<State>, action: BlockProducerActionWithMetaRef<'_>) {
```
tizoc, line 46: Please add a comment here explaining that `State` is accessed here and why. It will make our lives easier in the future.

node/src/block_producer/block_producer_reducer.rs
```diff
@@ -12,68 +12,104 @@ use mina_p2p_messages::{
...
+                BlockProducerVrfEvaluatorState::reducer(
+                    Substate::from_compatible_substate(state_context),
+                    meta.with_action(action),
+                );
             }
             BlockProducerAction::BestTipUpdate { best_tip } => {
-                self.injected_blocks.remove(best_tip.hash());
+                state.injected_blocks.remove(best_tip.hash());
                 // set the genesis timestamp on the first best tip update
                 // TODO: move/remove once we can generate the genesis block
```
tizoc, line 67: @adonagy we do generate the genesis block already, do you remember the context of this comment?

tizoc: I assume we can just set this timestamp as soon as the genesis block gets injected right? (maybe we are even doing it already?)

In which case there is no need to wait for the best tip.

node/src/block_producer/block_producer_reducer.rs
```diff
@@ -12,68 +12,104 @@ use mina_p2p_messages::{
...
             }
-            BlockProducerAction::WonSlotSearch => {}
             BlockProducerAction::WonSlot { won_slot } => {
-                self.current = BlockProducerCurrentState::WonSlot {
+                state.current = BlockProducerCurrentState::WonSlot {
                     time: meta.time(),
                     won_slot: won_slot.clone(),
                 };
+
+                state_context
```
tizoc, line 91: same as in the other case, lets not introduce this new pattern and assign `dispatcher` instead.

node/src/block_producer/block_producer_reducer.rs
```diff
@@ -519,35 +294,424 @@ impl BlockProducerEnabled {
...
+            curr_global_slot_since_hard_fork,
+            global_slot_since_genesis,
+            staking_epoch_data,
+            next_epoch_data,
+            has_ancestor_in_same_checkpoint_window,
+            block_stake_winner,
+            block_creator,
+            coinbase_receiver,
+            // TODO(binier): Staged_ledger.can_apply_supercharged_coinbase_exn
+            supercharge_coinbase: constraint_constants().supercharged_coinbase_factor != 0,
```
tizoc, line 572: hmmm out of scope for this PR but this logic is not right. And we already have the above function implemented already (but cannot be called from here)

node/src/block_producer/vrf_evaluator/block_producer_vrf_evaluator_reducer.rs
tizoc, line 339: I guess this should be a `bug_condition`. If this fails it means that we don't have a best tip yet or that the vrf evaluator isn't ready (then we shouldn't be handling this action)

tizoc (changes requested):
LGTM in general, but I need to do another pass. Also lets not merge it before tomorrow's release just in case.

tizoc (approved):
@0xMimir LGTM, please rebase on top of current `develop` before merging (just to make sure that all commits show up after the v0.11 merge and history doesn't get confusing).

Btw this second pass I did was pretty quick, is there anything I should have paid attention to other than the code moving around and the new callback? (it seems not but just in case)

PR #860, opened 2024-11-01
fix subtract with overflow in relative_min_window_density() consensus.rs

olton:
Panic with message: `Subtract with overflow`
at core/src/consensus.rs:78:27

When debug vars:
max_slot: 98879
global_slot + 1: 98880

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -75,7 +75,7 @@ pub fn relative_min_window_density(b1: &MinaConsensusState, b2: &MinaConsensusSt
 
     let projected_window = {
         // Compute shift count
-        let shift_count = (max_slot - global_slot(b1) - 1).clamp(0, SUB_WINDOWS_PER_WINDOW);
+        let shift_count = (max_slot.wrapping_sub(global_slot(b1) + 1)).clamp(0, SUB_WINDOWS_PER_WINDOW);
 
         // Initialize projected window
         let mut projected_window = b1
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -75,7 +75,8 @@ pub fn relative_min_window_density(b1: &MinaConsensusState, b2: &MinaConsensusSt
 
     let projected_window = {
         // Compute shift count
-        let shift_count = (max_slot.wrapping_sub(global_slot(b1) + 1)).clamp(0, SUB_WINDOWS_PER_WINDOW);
+        let shift_count = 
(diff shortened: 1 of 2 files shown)


core/src/consensus.rs
```diff
@@ -75,7 +75,8 @@ pub fn relative_min_window_density(b1: &MinaConsensusState, b2: &MinaConsensusSt
 
     let projected_window = {
         // Compute shift count
-        let shift_count = (max_slot - global_slot(b1) - 1).clamp(0, SUB_WINDOWS_PER_WINDOW);
+        let shift_count = 
+            (max_slot.wrapping_sub(global_slot(b1) + 1)).clamp(0, SUB_WINDOWS_PER_WINDOW);
```
tizoc, line 79: @olton thank you! this should use `saturating_sub` instead.

tizoc: This version would provide the intended behavior:

```rust
let shift_count = max_slot
    .saturating_sub(global_slot(b1) + 1)
    .clamp(0, SUB_WINDOWS_PER_WINDOW);
```

as per [the spec](https://github.com/MinaProtocol/mina/tree/develop/docs/specs/consensus#5412-relative-minimum-window-density):

```
let shift_count = min(
    max(max_slot - B1.curr_global_slot - 1, 0), // same as `saturating_sub`
    sub_windows_per_window
)
```

tizoc:
One last request, once you are done please squash everything into a single commit.

tizoc:
Lint check is failing because of the formatting, you just need to run `cargo fmt`.

core/src/consensus.rs
```diff
@@ -75,7 +75,8 @@ pub fn relative_min_window_density(b1: &MinaConsensusState, b2: &MinaConsensusSt
 
     let projected_window = {
         // Compute shift count
-        let shift_count = (max_slot - global_slot(b1) - 1).clamp(0, SUB_WINDOWS_PER_WINDOW);
+        let shift_count = 
+            (max_slot.saturating_sub(global_slot(b1) + 1)).clamp(0, SUB_WINDOWS_PER_WINDOW);
```
tizoc, line 79: Thinking about it a bit more, we don't even need `clamp` now that we use `saturating_sub` and it can be simplified a bit:

```suggestion
        let shift_count = max_slot
            .saturating_sub(global_slot(b1) + 1)
            .min(SUB_WINDOWS_PER_WINDOW);
```

PR #861, opened 2024-11-02
Fix `grace_period_end` value when building inputs for block proof

sebastiencs:


node/src/block_producer/block_producer_reducer.rs
```diff
@@ -493,15 +499,15 @@ impl BlockProducerEnabled {
...
-                    if is_same_global_sub_window || are_windows_overlapping && !within_range {
+                    if is_same_global_sub_window || (are_windows_overlapping && !within_range) {
                         density
                     } else {
                         0
                     }
                 })
                 .collect::<Vec<_>>();
 
-            let grace_period_end = grace_period_end(pred_block.constants());
```
tizoc, line 504: I am trying to figure out the source of this `grace_period_end` function, I cannot find it in the OCaml code. Other than that, I see that the OCaml implementation doesn't call it and refers directly to the constant like in your updated version: https://github.com/openmina/mina/blob/96f69bfa3aadeba319dbc09ee4a3525b4c2a825f/src/lib/consensus/proof_of_stake.ml#L1311-L1319

tizoc: https://github.com/openmina/mina/blob/96f69bfa3aadeba319dbc09ee4a3525b4c2a825f/src/lib/consensus/proof_of_stake.ml#L3056 and they are passed here (this comes from the context, not extracted from the pred block)

tizoc (approved):
LGTM

node/src/block_producer/block_producer_reducer.rs
tizoc: It used to exist when the code prior to this PR was written, but was removed in https://github.com/MinaProtocol/mina/pull/14776/commits/587c2a34e3e319b2208ee05feb7208c8f0a1e886

PR #871, opened 2024-11-05
Fix block proof generation with zero supply increase

sebastiencs:


tizoc:
@sebastiencs anything pending for this one?

sebastiencs:
@tizoc Yes, I will push more changes

PR #878, opened 2024-11-09
feat(consensus): Implement block pre-validation

binier:
@tizoc Please review this commit  once you have some time: https://github.com/openmina/openmina/pull/878/commits/23e0d35fdeea7d08d6cc21710aba0204dafd2151

### DIFF
--- a/node/src/state.rs
+++ b/node/src/state.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use mina_p2p_messages::v2::{MinaBaseUserCommandStableV2, MinaBlockBlockStableV2};
+use openmina_core::constants::PROTOCOL_VERSION;
 use rand::prelude::*;
 
 use openmina_core::block::BlockWithHash;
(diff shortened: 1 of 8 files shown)


node/src/state.rs
```diff
@@ -356,6 +368,71 @@ impl State {
...
+        block: &ArcBlockWithHash,
+    ) -> Result<(), BlockPrevalidationError> {
+        let Some((genesis, cur_global_slot)) =
+            None.or_else(|| Some((self.genesis_block()?, self.cur_global_slot()?)))
+        else {
+            // we don't have genesis block. This should be impossible
+            // because we don't even know chain_id before we have genesis
+            // block, so we can't be connected to any peers from which
+            // we would receive a block.
+            return Err(BlockPrevalidationError::GenesisNotReady);
```
binier, line 382: Maybe we should add bug_condition here?

node/src/state.rs
```diff
@@ -356,6 +368,71 @@ impl State {
...
+        if !protocol_versions_are_valid {
+            return Err(BlockPrevalidationError::InvalidProtocolVersion);
+        } else if !protocol_version_matches_daemon {
+            return Err(BlockPrevalidationError::MismatchedProtocolVersion);
+        }
+
+        // TODO(tizoc): check for InvalidDeltaBlockChainProof
+        // https://github.com/MinaProtocol/mina/blob/d800da86a764d8d37ffb8964dd8d54d9f522b358/src/lib/mina_block/validation.ml#L369
+        // https://github.com/MinaProtocol/mina/blob/d800da86a764d8d37ffb8964dd8d54d9f522b358/src/lib/transition_chain_verifier/transition_chain_verifier.ml
+
```
binier, line 432: Can we bring back constants check here?. Ocaml node doesn't have it, because those constants are never actually used (ones from the block), but in our case, we do.

node/src/state.rs
tizoc: Good pot. Btw do you know what's up with the failed p2p tests? If I run them locally they fail every time too but it is a different one. Changes in this PR shouldn't matter I think

node/src/state.rs
tizoc: @binier I am not sure if that will always be true (that these constants must match the genesis), but checking the current version of the OCaml node I don't see any logic that could change them.

PR #883, opened 2024-11-11
Inspect pubsub behavior, optimize broadcasting

vlad9486:


node/src/block_producer/block_producer_reducer.rs
```diff
@@ -304,8 +304,24 @@ impl BlockProducerEnabled {
...
+
                 let dispatcher = state_context.into_dispatcher();
                 dispatcher.push(BlockProducerAction::BlockInject);
+
+                #[cfg(feature = "p2p-libp2p")]
+                if let Some(block) = to_broadcast {
+                    use mina_p2p_messages::gossip::GossipNetMessageV2;
+
+                    let message = Box::new(GossipNetMessageV2::NewState(block));
+                    dispatcher.push(p2p::P2pNetworkPubsubAction::Broadcast { message });
```
tizoc, line 323: why is it not broadcasted after the `Inject` above? this means that it will be broadcasted even if the transition frontier rejects it right?

tizoc: (I don't mean inside the `BlockInject` action itself but as the result of it, same way as other broadcasted blocks, after successful application and integration into the frontier)

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -158,7 +168,10 @@ impl P2pNetworkPubsubState {
                     if !could_accept {
                         if let Some(topic_state) = map.get(&peer_id) {
                             if topic_state.on_mesh() {
-                                dispatcher.push(P2pNetworkPubsubAction::Prune { peer_id, topic_id })
+                                dispatcher.push(P2pNetworkPubsubAction::Prune {
+                                    peer_id,
+                                    topic_id: dbg!(topic_id),
```
tizoc, line 173: leftover `dbg!` here

binier (changes requested):

node/testing/src/simulator/mod.rs
```diff
@@ -217,19 +220,42 @@ impl Simulator {
         self.wait_for_all_nodes_synced(runner).await;
     }
 
+    pub async fn setup_and_run_with_listener<'a, AL, ALF>(
```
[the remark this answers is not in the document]
tizoc: I think it is for that graphviz visualization?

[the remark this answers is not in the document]
binier: Yeah but it doesnt matter, it basically wraps action handler and doesnt add anything useful to the existing api. I just don't get why we need it.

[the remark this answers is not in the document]
binier: Oh it's for simulator code, mb. Ok it's fine for now, I'll clean up api afterwards.

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -285,12 +285,22 @@ impl P2pNetworkPubsubState {
                 Ok(())
             }
             P2pNetworkPubsubAction::OutgoingMessageError { .. } => Ok(()),
-            P2pNetworkPubsubAction::Broadcast { message } => {
-                let mut seqno = pubsub_state.seq;
-                let (dispatcher, state) = state_context.into_dispatcher_and_state();
-                let config: &P2pConfig = state.substate()?;
-                seqno += config.meshsub.initial_time.as_nanos() as u64;
+            P2pNetworkPubsubAction::BroadcastBlock { block } => {
```
tizoc, line 288: lets avoid this custom action is wrapping the `NewState` block in `Arc` works

node/src/block_producer/block_producer_reducer.rs
```diff
@@ -353,6 +337,11 @@ impl BlockProducerEnabled {
                         }
                     )),
                 });
+
+                #[cfg(feature = "p2p-libp2p")]
+                dispatcher.push(p2p::P2pNetworkPubsubAction::BroadcastBlock {
```
tizoc, line 342: @vlad9486 @binier  is this needed? shouldn't the best tip update logic itself handle it? otherwise how do we broadcast blocks that we didn't produce?

tizoc: And if it is really needed, I would move it to `BlockInjected` bellow, we want to make sure we succeeded before broadcasting it right?

tizoc:
Superceded by https://github.com/openmina/openmina/pull/950 which already includes these changes

tizoc:
Closing in favor of #950 which already includes these changes

PR #890, opened 2024-11-12
Ledger: use eval pt name instead of fst and snd

dannywillems:
Initially, it was used to reflect the OCaml code in `src/lib/pickles/plonk_checks`. However, I do think we have to change that as it makes the code more readable from a cryptography point of view.


dannywillems:
FWIW: @sebastiencs was next to me at DevCon when doing this PR. We were commenting on some aesthetic improvements of Pickles and OpenMina

tizoc:
@dannywillems please go ahead and merge (with the approval you should already be able to do so)

sebastiencs (approved):

PR #920, opened 2024-11-20
fix(p2p): Avoid unnecessary buffer allocations on recv

tizoc:


p2p/src/service_impl/mio/mod.rs
```diff
@@ -476,20 +478,29 @@ where
                     )),
                 };
             }
-            Recv(addr, mut buf) => {
+            Recv(addr, limit) => {
                 if let Some(mut connection) = self.connections.remove(&addr) {
+                    // Ensure the buffer has enough space for the requested limit
+                    if limit > self.recv_buf.len() {
+                        // TODO: upper bound? resize to `limit` or try to allocate some extra space too?
+                        self.recv_buf.resize(limit, 0);
```
binier, line 486: Is this resize needed at all? We are reading `self.recv_buf[..limit]` anyways.

tizoc: @binier is slice-indexing supposed to automatically grow the vector? `limit` may be greater than the buffer size

PR #924, opened 2024-11-21
Ported ledger

0xMimir:


node/src/ledger/read/ledger_read_reducer.rs
```diff
@@ -1,20 +1,419 @@
...
+    LedgerAddress, LedgerReadAction, LedgerReadActionWithMetaRef, LedgerReadIdType,
+    LedgerReadInitCallback, LedgerReadRequest, LedgerReadResponse,
+    LedgerReadStagedLedgerAuxAndPendingCoinbases, LedgerReadState,
+};
 
 impl LedgerReadState {
-    pub fn reducer(&mut self, action: LedgerReadActionWithMetaRef<'_>) {
+    pub fn reducer(mut state_context: Substate<Self>, action: LedgerReadActionWithMetaRef<'_>) {
         let (action, meta) = action.split();
+        let Ok(ledger_read_state) = state_context.get_substate_mut() else {
```
tizoc, line 26: we (almost?) always call it just `state`, is there any good reason for this one to be the exception?

node/src/ledger/read/ledger_read_reducer.rs
```diff
@@ -1,20 +1,419 @@
...
+        }
+    }
+
+    fn propagate_read_response(
+        dispatcher: &mut Dispatcher<Action, State>,
+        state: &State,
+        id: RequestId<LedgerReadIdType>,
+        response: LedgerReadResponse,
+    ) {
+        let Some(request) = state.ledger.read.get(id) else {
```
tizoc, line 70: `bug_condition` ?

node/src/ledger/read/ledger_read_reducer.rs
```diff
@@ -1,20 +1,419 @@
...
+        match (request.request(), response) {
+            (
+                LedgerReadRequest::DelegatorTable(ledger_hash, pub_key),
+                LedgerReadResponse::DelegatorTable(table),
+            ) => {
+                let expected = state.block_producer.vrf_delegator_table_inputs();
+                if !expected.map_or(false, |(expected_hash, producer)| {
+                    ledger_hash == expected_hash && pub_key == producer
+                }) {
+                    eprintln!("delegator table unexpected");
```
tizoc, line 83: lets replace this with `bug_condition`

node/src/ledger/read/ledger_read_reducer.rs
```diff
@@ -1,20 +1,419 @@
...
+                    accounts,
+                    account_query,
+                });
+            }
+        }
+    }
+
+    fn next_read_requests_init(dispatcher: &mut Dispatcher<Action, State>, state: &State) {
+        // fetching delegator table
+        dispatcher.push(BlockProducerVrfEvaluatorAction::BeginDelegatorTableConstruction);
```
tizoc, line 206: hmmm I am wondering why this is even here. @binier do you remember?

node/src/ledger/write/ledger_write_reducer.rs
```diff
@@ -1,34 +1,160 @@
...
+
+use super::{
+    LedgerWriteAction, LedgerWriteActionWithMetaRef, LedgerWriteResponse, LedgerWriteState,
+};
 
 impl LedgerWriteState {
-    pub fn reducer(&mut self, action: LedgerWriteActionWithMetaRef<'_>) {
+    pub fn reducer(mut state_context: Substate<Self>, action: LedgerWriteActionWithMetaRef<'_>) {
         let (action, meta) = action.split();
+        let Ok(ledger_write_state) = state_context.get_substate_mut() else {
```
tizoc, line 18: Same as above, we always use `state`

node/src/ledger_effectful/ledger_effectful_effects.rs
```diff
@@ -0,0 +1,60 @@
...
+{
+    let (action, _meta) = action.split();
+
+    match action {
+        LedgerEffectfulAction::WriteInit { request, on_init } => {
+            store.service.write_init(request.clone());
+            store.dispatch(LedgerWriteAction::Pending);
+            store.dispatch_callback(on_init, request);
+        }
+        LedgerEffectfulAction::WriteSuccess => {
```
tizoc, line 31: this one is not really an effect right? never calls any service, it just dispatches further actions

tizoc (changes requested):
LGTM in general, but left some comments.

node/src/ledger_effectful/ledger_effectful_effects.rs
tizoc: btw the `ifs` here are just an optimization to avoid calling the enabling conditions for the other actions. In theory dispatching all the actions should be the same (assuming there are no bugs in the enabling conditions)

node/src/ledger/read/ledger_read_reducer.rs
binier: @tizoc Because delegator table construction requires reading from ledger. It could be that ledger read quota was reached when vrf tried to initiate that read, so we need to "retry" it if that's the case.

tizoc: Ok, can you add a comment mentioning that @0xMimir ?

PR #930, opened 2024-11-22
Move poseidon code out of `ledger`

sebastiencs:


ledger/src/database/database_impl.rs
```diff
@@ -328,7 +328,7 @@ impl DatabaseImpl<V2> {
         //     path
         // );
 
-        std::fs::create_dir_all(&path).ok();
+        // std::fs::create_dir_all(&path).ok();
```
tizoc, line 331: @sebastiencs we don't currently use this persistence at all right? it just leaves leftover files on `/tmp`

sebastiencs: @tizoc No, this was necessary when this Rust ledger was used in the OCaml node. The OCaml node was detecting if the directory existed or not

PR #938, opened 2024-11-25
Updated channels to reuse logic

0xMimir:
Added abstraction over channels service (`P2pChannelsService`), main reason is to simplify channels layer and reduce number of actions that perform same thing, before this each channel had its effectful counterpart which would interact with service, now each channel interacts with new actions that are abstraction over methods on `P2pChannelsService`.

Mental model of channels shouldn't change, only chain of actions from:
```
[Channel]Action -> [Channel]EffectfulAction -> Service call
```
to:
```
[Channel]Action -> Service call action -> Service call
```
Every channel still has its own ID and messages

Changes:

* Removed effectful action that send response/request, instead of sending message on channel like following:
```rs
dispatcher.push(P2pChannelsBestTipEffectfulAction::RequestSend { ... });
// or
dispatcher.push(P2pChannelsBestTipEffectfulAction::ResponseSend { ... });
```
Request/Response would be sent like following:
```rs
dispatcher.push(P2pChannelsEffectfulAction::RequestSend {
    msg: ChannelMsg::BestTipPropagation(BestTipPropagationChannelMsg::RequestMessage),
    ...
});
// or
dispatcher.push(P2pChannelsEffectfulAction::RequestSend {
    msg: ChannelMsg::BestTipPropagation(BestTipPropagationChannelMsg::ResponseMessage),
    ...
});
```

* Moved actions to initialize channels to single action with a callback, so instead of every channel having its own initialization action, channel id and callback should be passed to `P2pChannelsEffectfulAction::InitChannel` action.

Benefits:
* Easier to follow actions, only one action to open channel and one to send message, instead of every channel having its own effectful action to open channel, and instead of every message having its own action there is only one.
* Less code and less actions


tizoc:
@0xMimir when you are done please add a description on the summary of this PR explaining the change. Skimming through it I am still not sure I understand the benefits, and it seems to be changing the mental model of how channels work (we used to have one channel per kind of message, and each channel had one specific message, now it is one channel per peer with multiple types of messages?). It is important that the difference is clear so that we can properly evaluate it.

tizoc:
> when you are done please add a description on the summary
or maybe even before depending on how much work is left (seems to be close to ready but I may be wrong).

p2p/src/channels/best_tip/p2p_channels_best_tip_reducer.rs
```diff
@@ -136,9 +152,12 @@ impl P2pChannelsBestTipState {
                 let dispatcher = state_context.into_dispatcher();
 
                 if !is_libp2p {
-                    dispatcher.push(P2pChannelsBestTipEffectfulAction::ResponseSend {
+                    dispatcher.push(P2pChannelsEffectfulAction::RequestSend {
```
dkuehr, line 155: Previous code had `ResponseSend` and now is `RequestSend`, is this correct?

p2p/src/channels/rpc/p2p_channels_rpc_reducer.rs
```diff
@@ -238,10 +248,10 @@ impl P2pChannelsRpcState {
                     return Ok(());
                 }
 
-                dispatcher.push(P2pChannelsRpcEffectfulAction::ResponseSend {
+                dispatcher.push(P2pChannelsEffectfulAction::RequestSend {
                     peer_id,
-                    id,
-                    response,
+                    msg_id: MsgId::first(),
+                    msg: ChannelMsg::Rpc(RpcChannelMsg::Response(id, response.map(|v| *v))),
```
dkuehr, line 254: Ok, I see that now even "response" messages are wrapped into a `RequestSend` variant. I'm not fully convinced about this approach.

0xMimir: I could make one action `RequestSend` and one `ResponseSend` to make it clearer, or change name to something like `MessageSend`

tizoc: @0xMimir please spend more time explaining anything relevant about the changes in the summary of the issue so that reviewers don't need to figure it on their own. It has to be very clear what is going on and why this is a worthwhile improvement. There is obviously more going on than what is explained in the summary.

dkuehr (commented):
Could you provide a detailed list of the benefits of the channel reuse logic?

I'm not fully convinced about the extra level of wrapping that is required for pushing actions.

p2p/src/channels/p2p_channels_effectful_effects.rs
```diff
@@ -0,0 +1,133 @@
...
+        match self {
+            P2pChannelsEffectfulAction::InitChannel {
+                peer_id,
+                id,
+                on_success,
+            } => {
+                store.service().channel_open(peer_id, id);
+                store.dispatch_callback(on_success, peer_id);
+            }
+            P2pChannelsEffectfulAction::RequestSend {
```
tizoc, line 29: based on what @dkuehr mentioned in other comment, this probably needs a better name, otherwise it is confusing (as it is now we use `Request` to send responses too)

tizoc:
@binier if you have time, it would be useful to get your feedback on this one

p2p/src/channels/p2p_channels_effectful_effects.rs
```diff
@@ -0,0 +1,133 @@
...
+                }
+            }
+            P2pChannelsEffectfulAction::SignalingDiscoveryOfferEncryptAndSend {
+                peer_id,
+                pub_key,
+                offer,
+            } => {
+                match store.service().encrypt(&pub_key, offer.as_ref()) {
+                    Err(_) => {
+                        // todo!("Failed to encrypt webrtc offer. Handle it.")
```
tizoc, line 66: lets make this a `bug_condition` with `TODO` in the message so that it is at least not invisible if it happens

p2p/src/channels/rpc/p2p_channels_rpc_reducer.rs
```diff
@@ -40,7 +37,17 @@ impl P2pChannelsRpcState {
                 *rpc_state = Self::Init { time: meta.time() };
 
                 let dispatcher = state_context.into_dispatcher();
-                dispatcher.push(P2pChannelsRpcEffectfulAction::Init { peer_id });
+
+                // TODO(akoptelov): open a new stream, if we decide not to forcibly do that on connection established
```
binier, line 41: is this comment still relevant?

p2p/src/channels/signaling/discovery/p2p_channels_signaling_discovery_reducer.rs
```diff
@@ -74,9 +82,10 @@ impl P2pChannelsSignalingDiscoveryState {
 
                 let dispatcher = state_context.into_dispatcher();
                 let message = SignalingDiscoveryChannelMsg::GetNext;
-                dispatcher.push(P2pChannelsSignalingDiscoveryEffectfulAction::MessageSend {
+                dispatcher.push(P2pChannelsEffectfulAction::SendMessage {
```
binier, line 85: Could we keep the same naming convention for actions? First nouns then verbs. `MessageSend`

p2p/src/channels/signaling/discovery/p2p_channels_signaling_discovery_reducer.rs
```diff
@@ -112,11 +121,11 @@ impl P2pChannelsSignalingDiscoveryState {
...
                 let dispatcher = state_context.into_dispatcher();
-                dispatcher.push(P2pChannelsSignalingDiscoveryEffectfulAction::MessageSend {
+                let message = SignalingDiscoveryChannelMsg::Discovered { target_public_key };
+                dispatcher.push(P2pChannelsEffectfulAction::SendMessage {
                     peer_id,
-                    message: SignalingDiscoveryChannelMsg::Discovered {
-                        target_public_key: target_public_key.clone(),
-                    },
+                    msg_id: MsgId::first(),
+                    msg: ChannelMsg::SignalingDiscovery(message),
```
binier, line 128: `ChannelMsg` derives `From` so you could just use `message.into()` here.

binier (changes requested):
Looks good. Needs just some minor cosmetic changes.

p2p/src/channels/p2p_channels_effectful_effects.rs
binier: @tizoc I'd prefer just logging because this is the error case. It can happen if peer is (most likely) malicious.

p2p/src/channels/signaling/discovery/p2p_channels_signaling_discovery_reducer.rs
0xMimir: I know, but it think it is easier to read.

tizoc: @0xMimir in cases where the message is not obvious from the context I agree with you.

But in this case in particular (and others like it), since the contents of the message are immediately visible in the context, does it really aid with readability?

p2p/src/channels/p2p_channels_effectful_effects.rs
tizoc: @binier btw whats the proper way to handle this situation?

p2p/src/channels/p2p_channels_effectful_effects.rs
```diff
@@ -60,21 +60,18 @@ impl P2pChannelsEffectfulAction {
...
-                        store.service().channel_send(
-                            peer_id,
-                            super::MsgId::first(),
-                            message.into(),
-                        );
-                    }
+            } => match store.service().encrypt(&pub_key, offer.as_ref()) {
+                Err(_) => {
+                    // TODO: handle
+                    bug_condition!("Failed to encrypt webrtc offer")
```
tizoc, line 66: @0xMimir sorry lets do as @binier suggests and use regular logging, because if a malicious peer can trigger this then this isn't an internal bug.

binier (approved):

p2p/src/channels/snark/p2p_channels_snark_reducer.rs
```diff
@@ -178,7 +197,21 @@ impl P2pChannelsSnarkState {
                 };
 
                 let dispatcher = state_context.into_dispatcher();
-                dispatcher.push(P2pChannelsSnarkEffectfulAction::ResponseSend { peer_id, snarks });
+                dispatcher.push(P2pChannelsEffectfulAction::MessageSend {
+                    peer_id,
+                    msg_id: MsgId::first(),
+                    msg: ChannelMsg::SnarkPropagation(SnarkPropagationChannelMsg::WillSend {
```
tizoc, line 203: this one could use `into()` too

p2p/src/channels/snark/p2p_channels_snark_reducer.rs
```diff
@@ -65,7 +78,13 @@ impl P2pChannelsSnarkState {
                 };
 
                 let dispatcher = state_context.into_dispatcher();
-                dispatcher.push(P2pChannelsSnarkEffectfulAction::RequestSend { peer_id, limit });
+                dispatcher.push(P2pChannelsEffectfulAction::MessageSend {
+                    peer_id,
+                    msg_id: MsgId::first(),
+                    msg: ChannelMsg::SnarkPropagation(SnarkPropagationChannelMsg::GetNext {
```
tizoc, line 84: this one could use `into()` too

p2p/src/channels/transaction/p2p_channels_transaction_reducer.rs
```diff
@@ -66,8 +75,13 @@ impl P2pChannelsTransactionState {
                 };
 
                 let dispatcher = state_context.into_dispatcher();
-                dispatcher
-                    .push(P2pChannelsTransactionEffectfulAction::RequestSend { peer_id, limit });
+                dispatcher.push(P2pChannelsEffectfulAction::MessageSend {
+                    peer_id,
+                    msg_id: MsgId::first(),
+                    msg: ChannelMsg::TransactionPropagation(
```
tizoc, line 81: this one could use `into()` too

tizoc (approved):
left a few more comments, but overall it LGTM now + it reduces the service interaction surface which is a big plus

dkuehr (approved):

PR #940, opened 2024-11-27
Tx fuzzer updates

dkuehr:


sebastiencs:
@dkuehr Looks good.
Could you just add a test in CI ? To make sure we don't break that tool again.
`cargo check` might be enough

PR #944, opened 2024-11-28
Ported remaining callbacks in p2p layer and snark layer

0xMimir:
Added remaining callbacks in p2p and snark layer.

Removed state accesses from effectful functions.

### DIFF
--- a/node/src/rpc/rpc_reducer.rs
+++ b/node/src/rpc/rpc_reducer.rs
@@ -1,11 +1,12 @@
 use openmina_core::{
     block::AppliedBlock,
     bug_condition,
-    requests::{RequestId, RpcIdType},
+    requests::{RequestId, RpcId, RpcIdType},
 };
 use p2p::{
     connection::{incoming::P2pConnectionIncomingAction, outgoing::P2pConnectionOutgoingAction},
     webrtc::P2pConnectionResponse,
+    PeerId,
 };
 use redux::ActionWithMeta;
 
@@ -79,11 +80,21 @@ impl RpcState {
                 state.requests.insert(*rpc_id, rpc_state);
 
                 let dispatcher = state_context.into_dispatcher();
+
+                let callback = redux::callback!(
+                    on_p2p_connection_outgoing_rpc_connection_success((peer_id: PeerId, rpc_id: Option<RpcId>)) -> crate::Action {
+                        let Some(rpc_id) = rpc_id else {
+                            panic!("RPC ID not provided");
+                        };
+
+                        RpcAction::P2pConnectionOutgoingPending{ rpc_id }
+                    }
+                );
                 dispatcher.push(P2pConnectionOutgoingAction::Init {
                     opts: opts.clone(),
                     rpc_id: Some(*rpc_id),
+                    on_success: Some(callback),
                 });
-                dispatcher.push(RpcAction::P2pConnectionOutgoingPending { rpc_id: *rpc_id });
             }
             RpcAction::P2pConnectionOutgoingPending { rpc_id } => {
                 let Some(rpc) = state.requests.get_mut(rpc_id) else {
--- a/node/src/rpc/rpc_reducer.rs
+++ b/node/src/rpc/rpc_reducer.rs
@@ -81,19 +81,18 @@ impl RpcState {
 
                 let dispatcher = state_context.into_dispatcher();
 
-                let callback = redux::callback!(
-                    on_p2p_connection_outgoing_rpc_connection_success((peer_id: PeerId, rpc_id: Option<RpcId>)) -> crate::Action {
-                        let Some(rpc_id) = rpc_id else {
-                            panic!("RPC ID not provided");
-                        };
-
-                        RpcAction::P2pConnectionOutgoingPending{ rpc_id }
-                    }
-                );
                 dispatcher.push(P2pConnectionOutgoingAction::Init {
                     opts: opts.clone(),
                     rpc_id: Some(*rpc_id),
-                    on_success: Some(callback),
+                    on_success: Some(redux::callback!(
+                        on_p2p_connection_outgoing_rpc_connection_success((peer_id: PeerId, rpc_id: Option<RpcId>)) -> crate::Action {
+                            let Some(rpc_id) = rpc_id else {
+                                unreachable!("RPC ID not provided");
+                            };
+
+                            RpcAction::P2pConnectionOutgoingPending{ rpc_id }
+                        }
+                    )),
                 });
             }
             RpcAction::P2pConnectionOutgoingPending { rpc_id } => {
--- a/p2p/src/connection/outgoing/p2p_connection_outgoing_reducer.rs
+++ b/p2p/src/connection/outgoing/p2p_connection_outgoing_reducer.rs
@@ -36,8 +36,10 @@ impl P2pConnectionOutgoingState {
 
         match action {
             P2pConnectionOutgoingAction::RandomInit => {
-                let dispatcher = state_context.into_dispatcher();
-                dispatcher.push(P2pConnectionOutgoingEffectfulAction::RandomInit);
+                let (dispatcher, state) = state_context.into_dispatcher_and_state();
+                let p2p_state: &P2pState = state.substate()?;
+                let peers = p2p_state.disconnected_peers().collect::<Vec<_>>();
+                dispatcher.push(P2pConnectionOutgoingEffectfulAction::RandomInit { peers });
                 Ok(())
             }
             P2pConnectionOutgoingAction::Init {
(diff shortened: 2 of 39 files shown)


node/src/rpc/rpc_reducer.rs
```diff
@@ -79,11 +80,21 @@ impl RpcState {
                 state.requests.insert(*rpc_id, rpc_state);
 
                 let dispatcher = state_context.into_dispatcher();
+
+                let callback = redux::callback!(
+                    on_p2p_connection_outgoing_rpc_connection_success((peer_id: PeerId, rpc_id: Option<RpcId>)) -> crate::Action {
+                        let Some(rpc_id) = rpc_id else {
+                            panic!("RPC ID not provided");
```
tizoc, line 87: is there any way to avoid this panic here?

0xMimir: I could add enum callback like for `P2pNetworkKademliaStreamWaitOutgoingCallback`

tizoc: @0xMimir maybe it is not that bad being that the `rpc_id` argument is immediately after this code (or immediately before if you inline the callback). So it could be made into an `unreachable!` with a comment explaining why. I think I did something similar with other callbacks

p2p/src/connection/outgoing/p2p_connection_outgoing_reducer.rs
```diff
@@ -52,8 +58,13 @@ impl P2pConnectionOutgoingState {
                                 &opts,
                             )),
                             identify: None,
+                            on_connect_success: None,
                         });
 
+                if let Some(on_connect_success) = on_success {
+                    peer_state.on_connect_success = Some(on_connect_success);
```
tizoc, line 65: To make sure I understand here, this is required because the peer state may exists because this is an incoming connection, so in such case there is no need for the callback?

p2p/src/connection/outgoing/p2p_connection_outgoing_reducer.rs
```diff
@@ -509,19 +529,30 @@ impl P2pConnectionOutgoingState {
                     return Ok(());
                 }
 
+                let Some(peer_state) = p2p_state.peers.get_mut(&peer_id) else {
+                    bug_condition!("Outgoing peer state not found for: {}", peer_id);
+                    return Ok(());
+                };
+                let callback = peer_state.on_connect_success.take();
```
tizoc, line 536: Related to the content I left above, it would be good to have a comment explaining this a bit, because it is not immediately obvious

p2p/src/connection/outgoing/p2p_connection_outgoing_reducer.rs
0xMimir: Yes

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -123,14 +123,26 @@ impl P2pNetworkPubsubState {
...
             } => {
                 pubsub_state.reduce_incoming_data(&peer_id, data, meta.time())?;
 
-                let dispatcher: &mut Dispatcher<Action, State> = state_context.into_dispatcher();
+                let (dispatcher, state) = state_context.into_dispatcher_and_state();
+                let p2p_state: &P2pState = state.substate()?;
+                let state = &p2p_state.network.scheduler.broadcast_state;
+                let Some(state) = state.clients.get(&peer_id) else {
+                    // TODO: investigate, cannot reproduce this
+                    // bug_condition!("{:?} not found in state.clients", peer_id);
```
tizoc, line 136: should this be silently ignored? do you remember what was the context?

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -123,14 +123,26 @@ impl P2pNetworkPubsubState {
...
+                    return Ok(());
+                };
+
+                let messages = state.incoming_messages.clone();
```
tizoc, line 140: I am worried about all this cloning here and bellow, is it really needed? cannot the messages use `Arc` ?

probably out of scope for this requests, but worth exploring given that we are already having memory issues because of data (mis?)handling in the p2p layer

tizoc: If the answer is not immediately obvious please create an issue for that task so that you don't forget later.

p2p/src/network/scheduler/p2p_network_scheduler_actions.rs
```diff
@@ -107,10 +104,6 @@ pub enum P2pNetworkSchedulerAction {
         addr: ConnectionAddr,
     },
     /// Prune streams.
-    PruneStreams {
```
tizoc, line 110: @0xMimir why is this not needed anymore? (I saw there was a TODO to remove the action bellow, but I don't know the context)

p2p/tests/identify.rs
```diff
@@ -203,46 +202,12 @@ fn bad_node_effects(
                     multiaddr!(Https),
                 ];
 
-                let public_key = Some(SecretKey::rand().public_key());
```
tizoc, line 206: I don't know much about these tests but seems a lot if changing here, I will just assume it is correct.

tizoc (changes requested):
Left some comments (but mostly for later, only the outgoing connection callback storage is actionable I think). Overall looks much better than what we had before.

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
0xMimir: This was created by @vlad9486 so I just copied the comment, but since it is bug condition I could leave it in code and it would just log error in production

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
0xMimir: This was already cloned in original but in effectful effects, so it won't make difference, but most likely it could be optimized to reduce memory

p2p/src/network/scheduler/p2p_network_scheduler_actions.rs
0xMimir: It is action that is not longer needed it has no reducer and no effects anymore so it can be removed

p2p/tests/identify.rs
0xMimir: This is test to check we don't accept "bad" behavior in identify layer in this case for multiple same addresses

p2p/src/connection/outgoing/p2p_connection_outgoing_reducer.rs
```diff
@@ -488,40 +542,50 @@ impl P2pConnectionOutgoingState {
...
+                else {
                     bug_condition!(
                         "Invalid state for `P2pConnectionOutgoingAction::Success`: {:?}",
                         state
                     );
                     return Ok(());
-                }
+                };
+
+                // remove callback from state if it exists and call it, this is so same callback doesn't get called repeatedly
```
tizoc, line 560: @0xMimir this is more about retaining the callback so that it can be used later right? the previous state is discarded on the next line anyway. Lets get rid of the comment because it adds confusion

p2p/src/connection/outgoing/p2p_connection_outgoing_reducer.rs
```diff
@@ -109,6 +105,7 @@ impl P2pConnectionOutgoingState {
                         time,
                         opts: opts.clone(),
                         rpc_id,
+                        on_success: None,
```
tizoc, line 108: what does it mean for it to be `None` here? why do we care about success in other instances but not during reconnection?

p2p/src/connection/p2p_connection_state.rs
```diff
@@ -20,6 +20,7 @@ impl P2pConnectionState {
             time: Timestamp::ZERO,
             opts: opts.clone(),
             rpc_id: None,
+            on_success: None,
```
tizoc, line 23: same here, does it make sense for this to be `None` I see that in the only call that exists to this constructor, the callback is immediately assigned after

p2p/src/connection/outgoing/p2p_connection_outgoing_reducer.rs
0xMimir: `None` is here because we only need callback if we open outgoing connection from `rpc` layer or `kademlia` and need to update those layers after peer is connected

p2p/src/connection/p2p_connection_state.rs
0xMimir: I think that this is just a helper function to create "empty" state if it doesn't exist, if we created actual state we would have to create that state twice, once if it doesn't exist to create it and then once to update it

p2p/src/connection/outgoing/p2p_connection_outgoing_reducer.rs
tizoc: Ok, maybe a dumb question, but would it make sense to differentiate between `Init` and `Reconnecting` state then? only `Init` would have the callback. The fact that one requires a callback and the other doesn't makes me think they are not really the same state then.

tizoc (approved):
@0xMimir before merging please add the TODO's in the places where I mentioned the memory issues and create a new issue. I know it is out of scope for this PR and that the code was already like that before the PR but every time we spot those issues they have to be fixed or marked for fixing, otherwise it will never be done.

PR #950, opened 2024-11-29
fix(p2p): Pubsub optimizations

tizoc:
Continuation of #883 by @vlad9486  with extra changes by me and @binier 

- Less wasteful broadcasting
- Store blocks in `Arc` to avoid duplication
- Don't clone message data in actions, reference it from the state instead


node/testing/src/scenarios/multi_node/pubsub_advanced.rs
```diff
@@ -20,9 +29,61 @@ use crate::{
...
+                    let len = st.len();
+                    st[(len - 6)..len].to_owned()
+                };
+                let this = cut(&this);
+
+                match action.action() {
+                    Action::P2p(P2pAction::Network(P2pNetworkAction::Pubsub(
+                        P2pNetworkPubsubAction::OutgoingMessage { peer_id },
+                    ))) => {
+                        // FIXME: this doesn't work because this is called after the action
```
tizoc, line 55: @binier @vlad9486 now that the message is not part of the state I cannot fetch it here to collect the data for the test, do you have any suggestion?

Obvious solution I can come up is to add an intermediary action before `P2pNetworkPubsubAction::OutgoingMessage` which is basically a non-op so that I can intercept that instead, but feels dirty.

tizoc: alternatively, an post-Action `OutgoingMessageClear` that clears the message there, that feels like a better option.

tizoc: I went with the second option.

node/testing/src/server/simulator.rs
```diff
@@ -85,9 +85,11 @@ async fn simulation_create(
             while let Some(cluster_mutex) = cluster_mutex.upgrade() {
                 let mut cluster = cluster_mutex.lock().await;
                 let mut runner = ClusterRunner::new(&mut cluster, |_| {});
-                let _ =
-                    tokio::time::timeout(Duration::from_millis(500), simulator.run(&mut runner))
-                        .await;
+                let _ = tokio::time::timeout(
+                    Duration::from_millis(500),
+                    simulator.setup_and_run(&mut runner),
```
binier, line 90: this change is wrong and will break the webnodes custom cluster setup.
It will cause simulator to be `setup` over and over, which is not the goal.

node/testing/src/simulator/mod.rs
```diff
@@ -217,19 +220,42 @@ impl Simulator {
...
         self.set_up_block_producer_nodes(runner).await;
     }
 
-    pub async fn setup_and_run<'a>(&mut self, runner: &mut ClusterRunner<'a>) {
-        self.setup(runner).await;
-        self.run(runner).await;
-    }
-
-    pub async fn run<'a>(&mut self, runner: &mut ClusterRunner<'a>) {
+    pub async fn run_with_listener<'a, AL, ALF>(
```
binier, line 249: please bring back `run` as well and use that for testing/src/server/simulator.rs

binier (changes requested):
Just one change needed in testing framework. Rest looks ok.

node/testing/src/server/simulator.rs
tizoc: tbh I am not even sure what the change is about

binier: Yeah idk. @vlad9486 ?

node/testing/src/simulator/mod.rs
tizoc: done

node/testing/src/server/simulator.rs
vlad9486: Yes, it should be just `run` without `setup`.

PR #955, opened 2024-12-01
feat(p2p): More pubsub improvements, docs, cleanup

tizoc:
Organized so that each commit can be reviewed independently to make the review easier.

### DIFF
--- a/p2p/src/network/pubsub/pubsub_effectful/p2p_network_pubsub_effectful_effects.rs
+++ b/p2p/src/network/pubsub/pubsub_effectful/p2p_network_pubsub_effectful_effects.rs
@@ -30,7 +30,7 @@ impl P2pNetworkPubsubEffectfulAction {
                     store.dispatch(P2pNetworkPubsubAction::BroadcastSigned { signature });
                 }
             }
-            P2pNetworkPubsubEffectfulAction::IncomingData {
+            P2pNetworkPubsubEffectfulAction::ValidateIncomingMessages {
                 peer_id,
                 seen_limit,
                 addr,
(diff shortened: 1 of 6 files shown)


p2p/src/network/pubsub/pubsub_effectful/p2p_network_pubsub_effectful_effects.rs
```diff
@@ -30,57 +44,89 @@ impl P2pNetworkPubsubEffectfulAction {
...
-                            let Ok(data) = encode_without_signature_and_key(&mut message) else {
-                                // should never happen;
-                                // we just decode this message, so it should encode without error
-                                bug_condition!("serialization error");
-                                return;
+                    match validate_message(message, store) {
+                        Ok(valid_msg) => valid_messages.push(valid_msg),
+                        Err(error) => {
+                            let error_msg = match error {
+                                PubSubError::MissingSignature => {
```
0xMimir, line 60: Could this be done using `thiserror::Error` so that the string generated by the error can be reused and traced back to the enum variant more easily

0xMimir (approved):
Left a minor comment, but LGMT

vlad9486 (approved):
Looks great. Thank you for a wonderful improvement.

p2p/src/network/pubsub/pubsub_effectful/p2p_network_pubsub_effectful_effects.rs
tizoc: Done, much better now as I don't need to do that match to convert it to a string.

Issue #383, opened 2024-05-06
As a developer, I want to integrate archive node capability into the rust node, so we can use it with the producer dashboard


adonagy:

adonagy:
Update:

- Added workaround to run the producer-dashboard using an ocaml archive node

TODO: 

- Implement the RPCs so the archive node can pull(push?) data from the openmina node

adonagy:
Added archive service (WIP), conversions to the correct binprot type. 

Need to implement/work around the janestreet RPC protocol

adonagy:
- Implemented the RPC protocol
- Fixed some of the issues with token_ids
- Now the openmina node sends the diffs to the archive process

Working on a test that compares the data in openmina archive and ocaml archive

adonagy:
Still mainly focusing on the test, done some other stuff (libp2p key decryption, webnode testing)

adonagy:
Implemented Bruno's suggestion of modifying the archive process so it saves the recieved breadcrumbs. Working on the images/GHA part. 

(Implemented vrf slot evaluation stats for FE (stats rpc))

adonagy:
Testing uncovered a few bugs:

- In the initial catchup phase, not all the 290 blocks are received by the archive process (it varies but 284-288 are received, a few blocks are always missing)
- Some minor data missmatch between ocaml and openmina (in ocaml sometimes the block body hash is null)

adonagy:
Works in native (all blocks arrive into the archive), but in docker, there are still some issues. Investigating.

adonagy:
Fixed the issues with the rpc protocol. Now we are not missing blocks.

Fixing some mismatches between openmina vs ocaml node.

adonagy:
Investigating last missmatch (Protocol state proof mismatch). Either a ser/deser issue, might be a bug in the node itself?

adonagy:
Will address the review, and some finishing touches to the compose (mainly cleanup)

PR #956, opened 2024-12-02
Archive node compatibility

adonagy:
closes [#383](https://github.com/openmina/openmina/issues/383)


node/common/src/service/archive.rs
```diff
@@ -0,0 +1,321 @@
...
+    fn process_rpc(address: SocketAddr, data: &[u8]) -> io::Result<HandleResult> {
+        println!("[archive-service] Data length: {}", data.len());
+        let mut poll = Poll::new()?;
+        let mut events = Events::with_capacity(128);
+        let mut event_count = 0;
+
+        // We still need a token even for one connection
+        const TOKEN: Token = Token(0);
+
+        let mut stream = TcpStream::connect(address)?;
```
tizoc, line 156: do we really need to re-connect every time?

node/common/src/service/archive.rs
```diff
@@ -0,0 +1,321 @@
...
+                    // Other errors we'll consider fatal.
+                    Err(err) => return Err(err),
+                }
+                return Ok(HandleResult::ConnectionAlive);
+            }
+
+            if *handshake_received && *handshake_sent {
+                match connection.write_all(data) {
+                    Ok(_) => {
+                        println!("[archive-service] Message sent");
```
tizoc, line 237: here we would need to read the response (it is just unit, but still)

mina-p2p-messages/src/string.rs
```diff
@@ -4,7 +4,7 @@ use binprot::Nat0;
 use serde::{de::Visitor, Deserialize, Serialize};
 use serde_bytes;
 
-const MINA_STRING_MAX_LENGTH: usize = 100_000_000;
+const MINA_STRING_MAX_LENGTH: usize = 200_000_000;
```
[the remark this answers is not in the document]
adonagy: I need to doublecheck this.

node/common/src/service/archive.rs
```diff
@@ -0,0 +1,321 @@
...
+            loop {
+                match connection.read(&mut received_data[bytes_read..]) {
+                    Ok(0) => {
+                        connection_closed = true;
+                        break;
+                    }
+                    Ok(n) => {
+                        bytes_read += n;
+                        if bytes_read >= 15 {
+                            let handshake = [7, 0, 0, 0, 0, 0, 0, 0, 2, 253, 82, 80, 67, 0, 1];
```
tizoc, line 278: this was just captured and copied or do you know the source?

docker-compose.archive.devnet.compare.yml
```diff
@@ -0,0 +1,159 @@
...
+      LOG_LEVEL: Info
+      MINA_LIBP2P_PASS: v3rys3cr3t
+    command: >
+      daemon
+      --libp2p-keypair /keys/mina-privkey
+      --archive-address archive-ocaml:3086
+      --insecure-rest-server
+      --rest-port 5000
+      --enable-peer-exchange true
+      --peer-list-url https://bootnodes.minaprotocol.com/networks/devnet.txt
```
tizoc, line 138: Here and for the node it may be better to use our own peers list (from the `k8s-peers` job output), will be more reliable that way.

ledger/src/database/database_impl.rs
```diff
@@ -20,6 +20,7 @@ pub struct DatabaseImpl<T: TreeVersion> {
     accounts: Vec<Option<T::Account>>,
     pub hashes_matrix: HashesMatrix,
     id_to_addr: HashMap<AccountId, Address>,
+    token_to_account: HashMap<T::TokenId, AccountId>,
```
tizoc, line 23: IIRC maintaining this table was costly (thats why @sebastiencs removed it), so we need to add a runtime flag that will toggle it to be a noop when no archive node is used (for later, once everything else is solved I can add that)

ledger/src/mask/mask_impl.rs
```diff
@@ -1087,6 +1112,32 @@ impl BaseLedger for MaskImpl {
...
+
+        if let Some(account_id) = token_owners.get(&token_id).cloned() {
+            return Some(account_id);
+        };
+
+        parent.as_ref()?.token_owner(token_id)
+    }
+
+    fn token_owners(&self) -> HashSet<AccountId> {
+        // TODO: Not sure if it's the correct impl
```
tizoc, line 1134: this returns every account so probably not correct (not sure what it is supposed to do, maybe just accounts that own custom tokens?)

node/common/src/service/archive.rs
tizoc: it is defined on a `HANDSHAKE_MSG` constant above btw, then repeated here

node/common/src/service/archive.rs
```diff
@@ -0,0 +1,321 @@
...
+
+        if let Err(e) = Message::Query(Query {
+            tag: Method::NAME.into(),
+            version: Method::VERSION,
+            id: 1,
+            data: NeedsLength(data),
+        })
+        .binprot_write(&mut v)
+        {
+            node::core::warn!(
```
tizoc, line 133: this continues even if there is a serialization error

node/common/src/service/archive.rs
adonagy: @tizoc Yes, that was one reason. Also once synced, the node will send data only a few times every 3 minutes realistically. No need to keep the connection alive IMO.

ledger/src/bin/ledger.rs
```diff
@@ -6,7 +6,7 @@ fn main() {
 
         let now = redux::Instant::now();
 
-        let mut db = Database::<V2>::create(20);
+        let mut db = Database::<V2>::create(20, false);
```
tizoc, line 9: since most of the calls remain the same, wouldn't it be better to leave `create` as is and add a new `create_with_token_accounts` constructor for when it needs to be enabled?

mina-p2p-messages/src/v2/manual.rs
```diff
@@ -713,6 +714,26 @@ pub type NonZeroCurvePoint = Base58CheckOfBinProt<
...
+pub enum ArchiveTransitionFronntierDiff {
+    BreadcrumbAdded {
+        block: (MinaBlockBlockStableV2, (Option<StateBodyHash>, StateHash)),
+        accounts_accessed: List<(crate::number::UInt64, MinaBaseAccountBinableArgStableV2)>,
+        accounts_created: List<(MinaBaseAccountIdStableV2, CurrencyFeeStableV1)>,
+        tokens_used: List<(MinaBaseTokenIdStableV2, Option<MinaBaseAccountIdStableV2>)>,
+        sender_receipt_chains_from_parent_ledger:
+            List<(MinaBaseAccountIdStableV2, MinaBaseReceiptChainHashStableV1)>,
+    },
+    // TODO(adonagy): I think this is legacy stuff, doublecheck
```
tizoc, line 727: were you able to figure out if this is used at all? (I couldn't find any use myself, but it wasn't a super thorough search)

node/common/src/service/archive.rs
```diff
@@ -0,0 +1,490 @@
...
+        mut archive_receiver: mpsc::UnboundedReceiver<ArchiveTransitionFronntierDiff>,
+        address: SocketAddr,
+    ) {
+        while let Some(breadcrumb) = archive_receiver.blocking_recv() {
+            let mut retries = ARCHIVE_SEND_RETRIES;
+            while retries > 0 {
+                match rpc::send_diff(address, v2::ArchiveRpc::SendDiff(breadcrumb.clone())) {
+                    Ok(result) => {
+                        if result.should_retry() {
+                            node::core::warn!(node::core::log::system_time(); summary = "Archive suddenly closed connection, retrying...");
```
tizoc, line 30: `node::core::log::system_time()` is not the default if omitted btw, Sebastien added that [here](https://github.com/openmina/openmina/commit/3c97aab7e713d2099b75bde46686d8eb19950b29)

node/common/src/service/archive.rs
```diff
@@ -0,0 +1,490 @@
...
+    use mio::event::Event;
+    use mio::net::TcpStream;
+    use mio::{Events, Interest, Poll, Registry, Token};
+    use std::io::{self, Read, Write};
+    use std::net::SocketAddr;
+
+    const MAX_RECURSION_DEPTH: u8 = 25;
+
+    // messages
+    const HEADER_MSG: [u8; 7] = [2, 253, 82, 80, 67, 0, 1];
```
tizoc, line 91: probably not worth the effort to change this now if it is already working, but are the raw bytes really needed? iirc we had already found the source of this in the OCaml code.

node/common/src/service/p2p.rs
```diff
@@ -75,11 +75,7 @@ impl webrtc::P2pServiceWebrtc for NodeService {
         auth: ConnectionAuth,
     ) {
         let encrypted = auth.encrypt(&self.p2p.sec_key, other_pub_key, &mut self.rng);
-        if let Some(peer) = self.peers().get(&peer_id) {
-            let _ = peer
-                .cmd_sender
-                .send(webrtc::PeerCmd::ConnectionAuthorizationSend(encrypted));
-        }
+        Self::auth_send(self, peer_id, other_pub_key, encrypted);
```
tizoc, line 78: what is this change about?

node/src/ledger/ledger_service.rs
```diff
@@ -642,7 +652,7 @@ impl LedgerCtx {
         skip_verification: Option<SkipVerification>,
     ) -> Result<BlockApplyResult, String> {
         openmina_core::info!(openmina_core::log::system_time();
-            kind = "LedgerService::block_apply",
+            kind = "LedgerService::zy",
```
tizoc, line 655: this one seems accidental

node/src/ledger/ledger_service.rs
```diff
@@ -714,12 +724,130 @@ impl LedgerCtx {
...
+                })
+                .partition(|(_, status)| *status == AccessedOrNot::Accessed);
+
+            let mut account_ids_accessed: BTreeSet<_> =
+                accessed.into_iter().map(|(id, _)| id).collect();
+            let mut account_ids_not_accessed: BTreeSet<_> =
+                not_accessed.into_iter().map(|(id, _)| id).collect();
+
+            // Coinbase receiver is included only when the coinbase is not zero
+            // FIXME: Use the coinbase tx as base instead of the coinbase sum
```
tizoc, line 757: Leftover FIXME here. I assume this will work for now because the coinbase reward is not 0, but if possible it would be good to have the proper check.

tizoc:
General comment because it happens at many places in the diff: is passing the `is_archive` boolean around everywhere the best option? Since most of the time the value is `false` I would rather leave the current constructors as-is, and then either have an extra method that takes care of setting up the archive node requirements when called (not sure how viable it is, just an idea)

sebastiencs (changes requested):

ledger/src/database/database_impl.rs
```diff
@@ -102,6 +105,11 @@ impl DatabaseImpl<V2> {
         self.last_location = Some(location.clone());
         self.naccounts += 1;
 
+        if !token_id.is_default() {
+            if let Some(token_to_account) = self.token_to_account.as_mut() {
+                token_to_account.insert(account_id.derive_token_id(), account_id.clone());
```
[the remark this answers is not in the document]
tizoc: I think `token_to_account` will be `None` if the archive node is not set up, so this will never run.

RE: global, I was thinking about that in general to avoid passing `is_archive` around, but that means we cannot control that when running multiple nodes in the simulator.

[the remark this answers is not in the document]
tizoc: https://github.com/openmina/openmina/pull/956#issuecomment-2587007624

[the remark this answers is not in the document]
sebastiencs: > I think token_to_account will be None if the archive node is not set up, so this will never run.

Oh right

ledger/src/database/database_impl.rs
```diff
@@ -20,6 +20,7 @@ pub struct DatabaseImpl<T: TreeVersion> {
     accounts: Vec<Option<T::Account>>,
     pub hashes_matrix: HashesMatrix,
     id_to_addr: HashMap<AccountId, Address>,
+    token_to_account: Option<HashMap<T::TokenId, AccountId>>,
```
sebastiencs, line 23: Can we rename this to `token_owners` ? To match OCaml and `Mask`

ledger/src/mask/mask_impl.rs
```diff
@@ -27,6 +27,7 @@ pub enum MaskImpl {
     Attached {
         parent: Mask,
         owning_account: HashMap<AccountIndex, Account>,
+        token_owners: HashMap<TokenId, AccountId>,
```
sebastiencs, line 30: Can you make it an `Option<HashMap<TokenId, AccountId>>`, and optionally call `derive_token_id` in the diff of this file

sebastiencs (dismissed):

ledger/src/bin/ledger.rs
sebastiencs: Yes I agree, that seems a good solution

producer-dashboard/.sqlx/query-fe6ed440e2556eac41e90fa2942ceed3a481b8e3b0417fd3f1bc9eef1b6ca21b.json
```diff
@@ -0,0 +1,26 @@
+{
```
tizoc, line 1: what are all these files btw? are they needed in the repository?

adonagy: Ah, yeah, they are not needed. They were used by the previous idea of testing that queried the postgres DB and compared the output. Will cleanup this part

mina-p2p-messages/src/v2/manual.rs
adonagy: Not used. It looks like a remnant of some legacy code. Those 2 variants are not used or built anywhere.

node/common/src/service/p2p.rs
adonagy: @binier ^

node/src/service.rs
```diff
@@ -9,6 +9,7 @@ pub use crate::rpc_effectful::RpcService;
 pub use crate::snark::block_verify_effectful::SnarkBlockVerifyService;
 pub use crate::snark::work_verify_effectful::SnarkWorkVerifyService;
 pub use crate::snark_pool::SnarkPoolService;
+use crate::transition_frontier::archive::archive_service::ArchiveService;
```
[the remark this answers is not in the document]
adonagy: Typo

node/src/ledger/ledger_service.rs
tizoc: @adonagy yes, but those are cases in which there is no coinbase at all (which is not the same of there being a coinbase, but its reward being 0 because the protocol constants say so).

But thinking about it a bit more: since this is for creating the list of accounts being accessed, I guess the check you have right now is the correct one? If the coinbase transaction exists, but the sum is 0 I guess the account is not accessed at all either?

node/common/src/service/p2p.rs
binier: Oh. I wonder how this got here. It's just a cosmetic change, so we can keep it or discard the commit.

node/src/ledger/ledger_service.rs
adonagy: @tizoc If there is no coinbase tx, the sum the fn returns is 0, [here](https://github.com/openmina/openmina/blob/feat/archive-compatibility/mina-p2p-messages/src/v2/manual.rs#L1662). 

To be 100% correct, those constants should be set from the constraint/network constants, not hardcoded.

tizoc: @adonagy we are talking about different stuff. My point is, what you actually care about in this check you have here is not that there is a coinbase transaction, but that the sum of the coinbase reward is 0, right? Checking for the coinbase transaction existence in the case that the reward was set to 0, would give you the wrong behavior here, because if the reward were 0, you would not need to access the account, because nothing would be transfered to it. Does that make sense?

So the FIXME comment is wrong, and the current check is correct.

Now I noticed one more thing I am not very sure about: how does `account_ids_not_accessed` work? the coinbase received may not need to be accessed for the coinbase reward, but the account could be accessed because of some transfer to it, or because of earnings from some fee.

docker-compose.archive.devnet.yml
```diff
@@ -0,0 +1,74 @@
...
+    depends_on:
+      postgres-openmina:
+        condition: service_healthy
+
+    ports:
+      - "127.0.0.1:3086:3086"
+
+  node-openmina:
+    # TODO: Update once we have a new release
+    image: openmina/openmina:0.13.0
```
tizoc, line 65: I think it is safe to use `latest` here. We do in most compose files.

tizoc (approved):
There are a bunch of things that I would change and that I think that could be cleaner, also I don't like that `is_archive` argument at all. But that can be solved later, as long as it it works as expected LGTM

tizoc (approved):

PR #1001, opened 2024-12-18
Added block validation before broadcast

0xMimir:


tizoc:
@0xMimir when you think this is ready please expand the summary in the PR with mode details about the changes to ease reviewing (like for example what the new flow looks like compared to what it was before, how cleanup works , etc)

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -200,30 +206,21 @@ impl P2pNetworkPubsubState {
                     }
                 }
 
-                if let Err(error) = Self::broadcast(dispatcher, global_state) {
-                    bug_condition!(
-                        "Failure when trying to broadcast incoming pubsub message: {error}"
-                    );
+                let Some(message_content) = message_content else {
```
tizoc, line 209: Lets add a comment here clarifying that when `None` is returned, that means that the message has been seen before, and that is why we skip processing it. Also to the doc comment of `reduce_incoming_message`

0xMimir: It is okay for flow to be interrupted here, since message has already been seen enough times

node/src/p2p/callbacks/p2p_callbacks_actions.rs
```diff
@@ -46,6 +47,11 @@ pub enum P2pCallbacksAction {
     RpcRespondBestTip {
         peer_id: PeerId,
     },
+    P2pPubsubValidateMessage {
+        message_content: GossipNetMessageV2,
```
tizoc, line 51: Just for consideration, because I am still not sure: do `message_content` and `message` need to be part of the action? that is, instead of being part of the state and the actions just having a reference to them (through some id)

node/testing/src/scenarios/p2p/pubsub.rs
```diff
@@ -1,16 +1,18 @@
...
 
-/// Receive a block via meshsub
+/// Receive a message via meshsub
 /// 1. Create a normal node with default devnet config, with devnet peers as initial peers
 /// 2. Wait for 2 minutes
 /// 3. Create a node with discovery disabled and first node as only peer
-/// 4. Wait for first node to broadcast block to second one
+/// 4. Wait for first node to broadcast message to second one
 #[derive(documented::Documented, Default, Clone, Copy)]
 pub struct P2pReceiveBlock;
```
tizoc, line 17: change to the comment above replaces "block" with "message", but this is still called `Block`, is that correct?

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
tizoc: My point is that I would expect to get a result from the validation callback (which in this case would be just `Ignored`), and this return here breaks that flow. The callback is invoked, but a result never received in this case, which breaks expectations.

tizoc: re: mcache, it looks like it shouldn't be added when it was ignored anyway, right?

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -440,6 +437,126 @@ impl P2pNetworkPubsubState {
                 }
                 Ok(())
             }
+            P2pNetworkPubsubAction::BroadcastValidationCallback {
+                message,
+                message_content,
+                peer_id,
+                result,
+            } => {
+                let message_id = pubsub_state.mcache.put(message.clone());
```
tizoc, line 446: As mentioned in the other thread, here we are adding messages to the cache even in cases where we know we don't need them (rejected or ignored)

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -440,6 +437,126 @@ impl P2pNetworkPubsubState {
...
+                    message_id,
+                    peer_id,
+                    ..
+                } = message;
+
+                let Some(message_id) = message_id else {
+                    return Ok(());
+                };
+
+                let Some(message) = pubsub_state.mcache.map.get(&message_id) else {
```
tizoc, line 530: on which cases is this (and the above `message_id`) expected to be `None` ? Should it be a bug condition? Skipping the broadcasting of a block only because we don't have in the cache doesn't look right. But if there is a situation on which this would be the expected behavior please add a comment explaining that.

p2p/src/network/pubsub/p2p_network_pubsub_state.rs
```diff
@@ -58,6 +52,16 @@ pub struct P2pNetworkPubsubState {
...
     /// `iwant` requests, tracking the number of times peers have expressed interest in specific messages.
     pub iwant: VecDeque<P2pNetworkPubsubIwantRequestCount>,
+
+    /// Block messages currently being processed
+    pub block_messages: BTreeMap<mina_p2p_messages::v2::StateHash, P2pNetworkPubsubBlockMessage>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct P2pNetworkPubsubBlockMessage {
+    pub message_id: Option<Vec<u8>>,
```
tizoc, line 62: When is this `None` ?

0xMimir: I left is `Option` because of types, but I will change it

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
tizoc: I see it can be `None` if `compute_message_id` fails to produce an id, but is that something that should fail silently? shouldn't it be an error instead?

The way it works right now it just fails silently many steps before, and then it silently fails to broadcast the message here, thats not very good.

tizoc: `compute_message_id` should return a `Result` instead. What I don't understand is under which conditions the peer id can be invalid. That doesn't depend on the input right? if that happens, it is because of a bug? (in such case it should be a `bug_condition`, otherwise, it should interrupt the flow and dispatch an error action when it happens.

tizoc: Other possibility is that it is impossible for it to be invalid. If that is the case I would revert `compute_message_id` to use `unwrap`/`expect` like before and add a comment explaining that it cannot fail

node/src/p2p/callbacks/p2p_callbacks_reducer.rs
```diff
@@ -290,6 +296,43 @@ impl crate::State {
...
+                message_content,
+                peer_id,
+            } => {
+                let result = match message_content {
+                    GossipNetMessageV2::NewState(new_best_tip) => {
+                        match BlockWithHash::try_new(new_best_tip.clone()) {
+                            Ok(block) => {
+                                let allow_block_too_late = allow_block_too_late(state, &block);
+                                match state.prevalidate_block(&block, allow_block_too_late) {
+                                    Ok(()) => ValidationResult::Valid,
```
tizoc, line 310: At this point we only know that the block is well-formed, but that doesn't mean that it is valid. We can only consider it valid *after* it has been successfully applied.

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
tizoc: Btw a case in which we are not going to have anything in the cache is when the best tip comes from webrtc.

tizoc: same for the `pubsub_state.block_messages.remove(&hash)` above I think

tizoc: (the webrtc->libp2p aspect is out of scope of this PR unless you think it is easy to solve, but if thats not the case I would add a TODO here)

node/src/transition_frontier/sync/transition_frontier_sync_effects.rs
```diff
@@ -76,6 +76,9 @@ impl TransitionFrontierSyncAction {
                 if let Some(callback) = on_success {
                     store.dispatch_callback(callback.clone(), ());
                 }
+
+                let hash = best_tip.hash.clone();
+                store.dispatch(P2pNetworkPubsubAction::BroadcastAcceptedBlock { hash });
```
tizoc, line 81: this should not happen here right? We just triggered `BlockNextApplyInit` above, we don't know if it is going to succeed or not. Only after `BlocksNextApplySuccess` we know it succeeded. Also if `BlocksNextApplyError` is dispatched, then we must reject this message.

tizoc (changes requested):
I left a bunch of comments, some are about issues that are probably easy to solve, but there are two aspects that may need a bit more discussion:

- When the block message really needs to be re-broadcasted (now it is happening too early)
- That the block validation callback is not going through the full flow (right now it only considers block pre-validation, not block application, it is only after that point that you really know if it is `Valid`)

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -440,6 +437,126 @@ impl P2pNetworkPubsubState {
...
+                                    nonce: nonce.as_u32(),
+                                });
+                            }
+                            _ => {}
+                        }
+
+                        Self::broadcast(dispatcher, global_state)
+                    }
+                    super::ValidationResult::Reject => {
+                        // TODO: add error variants for transactions and snarks
```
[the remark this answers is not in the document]
0xMimir: This is about creating new error variants for on `P2pDisconnectionReason` one for transaction validation and on for snark.

node/src/p2p/callbacks/p2p_callbacks_reducer.rs
```diff
@@ -290,6 +295,38 @@ impl crate::State {
                     best_tip: best_tip.clone(),
                 });
             }
+            P2pCallbacksAction::P2pPubsubValidateMessage {
```
tizoc, line 298: @0xMimir I would expect the block to be rejected here in the case of any error, but that is not happening, how does it work?

node/src/p2p/callbacks/p2p_callbacks_reducer.rs
```diff
@@ -290,6 +296,58 @@ impl crate::State {
...
+                        });
+                    }
+                    PreValidationResult::Reject => {
+                        dispatcher.push(P2pNetworkPubsubAction::RejectMessage {
+                            message_id: p2p::BroadcastMessageId::MessageId {
+                                message_id: *message_id,
+                            },
+                        });
+                    }
+                    PreValidationResult::Ignore => {}
```
tizoc, line 348: @0xMimir It is important that an action signaling that it was ignored is dispatched here so that we have visibility of the fact that it was ignored, even if the action handler doesn't do anything.

p2p/src/network/pubsub/p2p_network_pubsub_actions.rs
```diff
@@ -91,32 +101,65 @@ pub enum P2pNetworkPubsubAction {
...
+    },
+    HandleIncomingMessage {
+        message: pb::Message,
+        message_content: GossipNetMessageV2,
+        peer_id: PeerId,
+    },
+    /// Delete expired messages from state
+    PruneMessages {},
+    RejectMessage {
+        message_id: BroadcastMessageId,
```
tizoc, line 161: I would add a "reason" here (can be a static string). Consider that this will show up in logs and it is useful to know why it happened (for debugging purposes). Same if we are writing tests for specific cases and want to ensure that it was rejected for the right reason.

Same for the `IgnoreMessage` action I mentioned in another comment, the fact that it was ignored must be reflected in an action (even if it does nothing), otherwise it is invisible (both for test-writting and debugging purposes).

tizoc: > even if it does nothing
that will probably not be the case btw, we can use the `Ignore` action to discard the contents and keep just the message id around so that we can recognize that we have seen it already if we get that same message again.

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -173,22 +181,31 @@ impl P2pNetworkPubsubState {
                 message,
                 seen_limit,
             } => {
+                // Check that if we can extract source from message, this is pre check
+                if source_from_message(&message).is_err() {
+                    let dispatcher = state_context.into_dispatcher();
+                    dispatcher.push(P2pDisconnectionAction::Init {
```
tizoc, line 187: Shouldn't here the message be rejected instead? and that will cause the peer to be disconnected

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -200,29 +217,40 @@ impl P2pNetworkPubsubState {
                     }
                 }
 
-                if let Err(error) = Self::broadcast(dispatcher, global_state) {
-                    bug_condition!(
-                        "Failure when trying to broadcast incoming pubsub message: {error}"
-                    );
+                // This happens if message was already seen
+                let Some(message_content) = message_content else {
+                    return Ok(());
```
tizoc, line 222: if it was seen already, this should dispatch the Ignore action for that message

p2p/src/network/pubsub/p2p_network_pubsub_actions.rs
```diff
@@ -91,32 +101,65 @@ pub enum P2pNetworkPubsubAction {
...
     ///
     /// **Fields:**
     /// - `data`: The encoded data to be sent.
-    OutgoingData { data: Data, peer_id: PeerId },
+    OutgoingData {
+        data: Data,
+        peer_id: PeerId,
+    },
+
+    BroadcastValidationCallback {
```
tizoc, line 146: this needs a better name, because the callback is now that we call to validate the message, the callback is what the validator calls to signal that the message was valid, ignored or rejected. Could be `ValidateIncomingMessage` or something like that. And I would put it after `HandleIncomingMessage` because it happens after, not before

p2p/src/network/pubsub/p2p_network_pubsub_actions.rs
```diff
@@ -91,32 +101,65 @@ pub enum P2pNetworkPubsubAction {
...
+        message_id: P2pNetworkPubsubMessageCacheId,
+    },
+
+    BroadcastValidatedMessage {
```
tizoc, line 150: This should also be after `BroadcastValidatedMessage`, it doesn't happen before, it happens after

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -200,29 +217,40 @@ impl P2pNetworkPubsubState {
...
+                        .put(message, message_content, peer_id, time)
+                else {
+                    bug_condition!("Unable to add message to `mcache`");
+                    return Ok(());
+                };
+
+                let (dispatcher, state) = state_context.into_dispatcher_and_state();
+                let p2p_state: &P2pState = state.substate()?;
+
+                if let Some(callback) = p2p_state.callbacks.on_p2p_pubsub_message_received.clone() {
```
tizoc, line 249: can you clarify this? the callback registered there seems to be doing the same thing as the `P2pNetworkPubsubAction::BroadcastValidationCallback` handler? why are both needed? in which case is the `on_p2p_pubsub_message_received` callback not be set?

node/src/p2p/callbacks/p2p_callbacks_reducer.rs
```diff
@@ -290,6 +296,67 @@ impl crate::State {
...
+                            message_id: *message_id,
+                        });
+                    }
+                    PreValidationResult::Reject => {
+                        dispatcher.push(P2pNetworkPubsubAction::RejectMessage {
+                            message_id: Some(p2p::BroadcastMessageId::MessageId {
+                                message_id: *message_id,
+                            }),
+                            peer_id: None,
+                            reason: "Pre validation failed".to_owned(),
```
tizoc, line 347: I would like the reason to be included here. It is contained in the result of `prevalidate_block` but discarded.

node/src/p2p/callbacks/p2p_callbacks_reducer.rs
```diff
@@ -290,6 +296,67 @@ impl crate::State {
                     best_tip: best_tip.clone(),
                 });
             }
+            P2pCallbacksAction::P2pPubsubValidateMessage { message_id } => {
+                let Some(message_content) = state.p2p.ready().and_then(|p2p| {
+                    p2p.network
+                        .scheduler
+                        .broadcast_state
+                        .mcache
+                        .get_message(message_id)
```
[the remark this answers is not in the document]
0xMimir: Yes this should be bug condition

node/src/p2p/callbacks/p2p_callbacks_reducer.rs
```diff
@@ -290,6 +296,72 @@ impl crate::State {
...
+                let pre_validation_result = match message_content {
+                    GossipNetMessageV2::NewState(new_best_tip) => {
+                        match BlockWithHash::try_new(new_best_tip.clone()) {
+                            Ok(block) => {
+                                let allow_block_too_late = allow_block_too_late(state, &block);
+                                match state.prevalidate_block(&block, allow_block_too_late) {
+                                    Ok(()) => PreValidationResult::Continue,
+                                    Err(BlockPrevalidationError::ReceivedTooEarly { .. }) => {
+                                        PreValidationResult::Ignore {
+                                            reason: "Block received too early".to_owned(),
```
tizoc, line 320: lets use `format!("Block prevalidation failed: {:?}", error),` as in the rejection case, so that the contents of the error are present in the message

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -440,6 +472,156 @@ impl P2pNetworkPubsubState {
                 }
                 Ok(())
             }
+            P2pNetworkPubsubAction::ValidateIncomingMessage { message_id } => {
+                let Some(message) = pubsub_state.mcache.map.remove(&message_id) else {
+                    return Ok(());
```
tizoc, line 477: is this a bug_condition ?

0xMimir: Yes this should be a bug condition

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -440,6 +472,156 @@ impl P2pNetworkPubsubState {
...
+
+                dispatcher.push(P2pNetworkPubsubAction::BroadcastValidatedMessage {
+                    message_id: super::BroadcastMessageId::MessageId { message_id },
+                });
+                Ok(())
+            }
+            P2pNetworkPubsubAction::BroadcastValidatedMessage { message_id } => {
+                let Some((message_id, message)) =
+                    pubsub_state.mcache.get_message_id_and_message(message_id)
+                else {
```
tizoc, line 553: same here, is this a bug condition? I assume that by the time these messages are dispatched the message exists. If there are situation in which it will not exist, the enabling condition should probably take care of checking that

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -440,6 +472,156 @@ impl P2pNetworkPubsubState {
...
+
+                Self::broadcast(dispatcher, state)
+            }
+            P2pNetworkPubsubAction::PruneMessages {} => {
+                let messages = pubsub_state
+                    .mcache
+                    .map
+                    .iter()
+                    .filter_map(|(message_id, message)| {
+                        if message.time() + Duration::from_secs(300) > time {
```
tizoc, line 576: lets define this into a properly named constant

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -440,6 +472,156 @@ impl P2pNetworkPubsubState {
...
+            } => {
+                let mut peer_id = peer_id;
+                if let Some(message_id) = message_id {
+                    let Some((_message_id, message)) =
+                        pubsub_state.mcache.get_message_id_and_message(message_id)
+                    else {
+                        return Ok(());
+                    };
+
+                    if peer_id.is_none() {
```
tizoc, line 602: is this correct? under which conditions is `peer_id` none? and can we really take the originator as the actual peer id in such cases?

p2p/src/network/pubsub/p2p_network_pubsub_state.rs
```diff
@@ -200,54 +198,203 @@ impl P2pNetworkPubsubClientState {
...
+        time: Timestamp,
+    },
+    // This is temporary handling for transactions and snark pool
+    PreValidated {
+        message: pb::Message,
+        peer_id: PeerId,
+        time: Timestamp,
+    },
+    Rejected {
+        message: pb::Message,
```
tizoc, line 233: do we need to keep this if it was rejected? it is not going to be broadcasted or used for anything

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -531,17 +669,63 @@ impl P2pNetworkPubsubState {
...
+    /// * `Ok(None)` - Message was a duplicate (already seen)
+    /// * `Err(String)` - Error during processing (invalid message format or deserialization failure)
+    ///
+    #[inline(never)]
+    fn reduce_incoming_message(
+        &mut self,
+        message: &Message,
+        seen_limit: usize,
+    ) -> Result<Option<GossipNetMessageV2>, String> {
+        if let Some(signature) = &message.signature {
```
tizoc, line 706: I think this shouldn't ever be `None`, so it is safe to instead
```suggestion
       let Some(signature) = &message.signature else { bug_condition_and_return }
```

tizoc: I mean, the current version should work as is too, but this would make the intention clearer, now it looks like the message can come without signature, which is not the case

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
tizoc: But when can that be `None`? I am confused about this switch between who sent us the message and the originator

tizoc (approved):
LGTM

@0xMimir not for this PR, but for the future. How hard do you think it would be to write tests for the different situations handled here?

PR #1002, opened 2024-12-19
Use malloc_sizeof

### LINKED ISSUE #954: As a developer, I want to be able to measure the memory consumptions of the state, so that it is easier to spot what to optimize
Task is to implement functions to measure the memory consumption of the different parts of the state that we use in the state machine.

This crate seems useful:

https://docs.rs/graphannis-malloc_size_of/latest/graphannis_malloc_size_of/
https://docs.rs/graphannis-malloc_size_of_derive/latest/graphannis_malloc_size_of_derive/

At the top of each crate this is needed so that the macro works:

```rust
extern crate graphannis_malloc_size_of as malloc_size_of;
extern crate graphannis_malloc_size_of_derive as malloc_size_of_derive;
```

And in `Cargo.toml`:

```toml
[dependencies]
graphannis-malloc_size_of = "2.0.0"
graphannis-malloc_size_of_derive = "2.0.0"
```

Not everything needs to be measured (`P2pConfig` for example is not important to measure), but anything with a dynamic size (networking and peer state in the `p2p` crate for example, the snark pool state, transaction pool state, transition frontier state, etc)

vlad9486:
resolves #954

### DIFF
--- a/node/src/rpc_effectful/rpc_effectful_effects.rs
+++ b/node/src/rpc_effectful/rpc_effectful_effects.rs
@@ -1,4 +1,8 @@
-use std::{collections::BTreeMap, time::Duration};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    ffi::c_void,
+    time::Duration,
+};
 
 use super::{super::rpc, RpcEffectfulAction};
 use crate::{
@@ -807,8 +811,10 @@ fn compute_node_status<S: Service>(store: &mut Store<S>) -> RpcNodeStatus {
         },
         current_block_production_attempt,
         p2p_malloc_size: {
-            let mut ops = MallocSizeOfOps::new(None, None);
-            state.p2p.size_of(&mut ops)
+            let mut set = BTreeSet::new();
+            let fun = move |ptr: *const c_void| !set.insert(ptr.addr());
+            let mut ops = MallocSizeOfOps::new(None, Some(Box::new(fun)));
+            size_of_val(&state.p2p) + state.p2p.size_of(&mut ops)
         },
     };
     status
--- a/node/src/rpc_effectful/rpc_effectful_effects.rs
+++ b/node/src/rpc_effectful/rpc_effectful_effects.rs
@@ -815,7 +815,7 @@ fn compute_node_status<S: Service>(store: &mut Store<S>) -> RpcNodeStatus {
                 let mut set = BTreeSet::new();
                 let fun = move |ptr: *const c_void| !set.insert(ptr.addr());
                 let mut ops = MallocSizeOfOps::new(None, Some(Box::new(fun)));
-                size_of_val(&state.p2p) + state.p2p.size_of(&mut ops)
+                size_of_val(&state.p2p).saturating_add(state.p2p.size_of(&mut ops))
             },
             transition_frontier: state.transition_frontier.resources_usage(),
             snark_pool: state.snark_pool.resources_usage(),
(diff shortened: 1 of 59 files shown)


tizoc:
@vlad9486 just saw this change https://github.com/openmina/redux-rs/commit/9b5b14f35d3f5ecebcd4b539c86678d2eef72e72 cannot fields with that value just be skipped so that we don't need that change in redux-rs?

node/src/rpc_effectful/rpc_effectful_effects.rs
```diff
@@ -804,6 +810,16 @@ fn compute_node_status<S: Service>(store: &mut Store<S>) -> RpcNodeStatus {
             transaction_candidates: state.transaction_pool.candidates.transactions_count(),
         },
         current_block_production_attempt,
+        resources_status: RpcNodeStatusResources {
+            p2p_malloc_size: {
+                let mut set = BTreeSet::new();
+                let fun = move |ptr: *const c_void| !set.insert(ptr.addr());
```
tizoc, line 816: @vlad9486 the compiler is complaining because of this line, error is:

```
error[E0658]: use of unstable library feature 'strict_provenance'
   --> node/src/rpc_effectful/rpc_effectful_effects.rs:816:[69](https://github.com/openmina/openmina/actions/runs/12887689269/job/35930919449?pr=1002#step:6:70)
    |
816 |                 let fun = move |ptr: *const c_void| !set.insert(ptr.addr());
    |                                                                     ^^^^
    |
    = note: see issue #95228 <https://github.com/rust-lang/rust/issues/95228> for more information
 ```
 
 But I see it has been stabilized a few months ago, maybe a more recent release of the compiler already includes that feature?

Comment (tizoc) [reply node/src/rpc_effectful/rpc_effectful_effects.rs:816]: @vlad9486 https://www.infoworld.com/article/3732079/rust-1-84-introduces-strict-provenance-apis.html

1.84 includes it, so we just need to upgrade

Comment (tizoc) [reply node/src/rpc_effectful/rpc_effectful_effects.rs:816]: Making that change in another PR, you can then rebase this PR on top of `develop` after I have merged it

Comment (tizoc) [reply node/src/rpc_effectful/rpc_effectful_effects.rs:816]: #1030 merged, please rebase

File: node/src/rpc_effectful/rpc_effectful_effects.rs [hunk 2]
```diff
@@ -804,6 +810,16 @@ fn compute_node_status<S: Service>(store: &mut Store<S>) -> RpcNodeStatus {
...
+                let mut ops = MallocSizeOfOps::new(None, Some(Box::new(fun)));
+                size_of_val(&state.p2p) + state.p2p.size_of(&mut ops)
```
tizoc, line 818: saturating add

node/src/transition_frontier/transition_frontier_state.rs
```diff
@@ -196,4 +196,19 @@ impl TransitionFrontierState {
...
+        serde_json::json!({
+            "best_chain_size": self.best_chain.len(),
+            "needed_protocol_states_size": self
+                .needed_protocol_states
+                .len(),
+            "blacklist_size": self.blacklist.len(),
+            "diff_tx_size": self
+                .chain_diff
+                .as_ref()
+                .map(|d| d.new_commands.len() + d.removed_commands.len())
```
tizoc, line 210: same here, and I think it will be the case for all these mem usage calculations, saturating ops

PR #1004, opened 2024-12-19
fix panic in ledger_manager when send method return Err

olton:
The error:
```
thread 'main' panicked at node/src/ledger/ledger_manager.rs:478:14:
called `Result::unwrap()` on an `Err` value: SendError { .. }
```
From the context, the specified error occurred in the `LedgerCaller::call` method, which uses the `mpsc` channel indicates that the `send` method called `unwrap`, but it returned `Err`. This means that either there is not enough space in the `mpsc` channel to send, or some other component (such as the receiver) has already completed or closed.

### Potential causes:
1. **Channel already closed**: If the process of receiving messages from the channel (receiver) has completed, then `mpsc::Sender::send` will return `Err`.
2. **Synchronization errors**: Your program may have incorrectly handled the lifecycle of components that use the channel.
3. **Incorrect channel usage**: For example, after creating a channel with `sync_channel`, the program may have forgotten to call the appropriate methods to receive messages.

#### Редагування методу `LedgerCaller::call` and `LedgerCaller::call_sync`

Instead of using `unwrap()`:

```rust
pub fn call(&self, request: LedgerRequest) {
    if let Err(err) = self.0.send(LedgerRequestWithChan {
        request,
        responder: None,
    }) {
        eprintln!("Failed to send request: {}", err);
    }
}
```


tizoc:
Thanks @olton, but this doesn't fix the real issue. We want this to panic, because if the node reaches this point it means that the ledger thread crashed, so the node will not recover from this. Whenever you see this error you have to look back in the logs to find the real error.

olton:
Understood

PR #1008, opened 2024-12-26
feat: add docs china translation

lispking:


tizoc:
@lispking thank you. Give me a few days to discuss this internally (part of the team is taking some rest).

What I was thinking is that it may be a better idea to just add an external link to the translated document (which you can maintain in another repo). The reason is that we cannot keep these documents up to date ourselves (both because of the language and because of time).

lispking:
Regarding this part, there is also a feasible solution which is to use mdbook to write the documentation. It is worth mentioning that mdbook itself has the i18n function, which can provide support for other international languages.

PR #1021, opened 2025-01-10
Added temporary fix for data handling in yamux

0xMimir:
Added temporary fix in yamux so that rust node is able to sync from rust node, in this PR proper handling of outgoing data with respect to window size is removed


tizoc (approved):
LGTM. Please add a description to the summary as for why this was needed to keep a record. Then open a new issue to create a task for the actual fix, referencing this PR

PR #1022, opened 2025-01-10
Minor refactors in yamux layer

0xMimir:
* Removed frame cloning in `P2pNetworkYamuxAction::IncomingFrame`
* Changed types in `YamuxFrameInner` from `i32` to `u32`

### DIFF
--- a/p2p/src/network/yamux/p2p_network_yamux_reducer.rs
+++ b/p2p/src/network/yamux/p2p_network_yamux_reducer.rs
@@ -89,7 +89,7 @@ impl P2pNetworkYamuxState {
                                 }
                             }
                             1 => {
-                                let difference = i32::from_be_bytes(b);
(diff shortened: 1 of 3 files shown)


p2p/src/network/yamux/p2p_network_yamux_reducer.rs
```diff
@@ -477,22 +474,13 @@ impl P2pNetworkYamuxState {
...
-                *window = 0;
-            } else {
-                *window = (*window).wrapping_sub(decreasing);
-            }
-        } else {
-            let increasing = difference as u32;
-            *window = (*window).wrapping_add(increasing);
-        }
+
+        *window = (*window).wrapping_add(difference);
```
tizoc, line 484: lets use saturating here, I don't think wrapping is what we want

PR #1024, opened 2025-01-15
Fix transaction pool `revalidate`

sebastiencs:
Apply changes from MinaProtocol/mina#16401

### DIFF
--- a/ledger/src/proofs/transaction.rs
+++ b/ledger/src/proofs/transaction.rs
@@ -4755,7 +4755,12 @@ pub(super) mod tests {
         }
 
         let rsa_private_key = {
-            let Ok(string) = std::fs::read_to_string("~/.openmina/debug/rsa.priv") else {
+            let Ok(home) = std::env::var("HOME") else {
+                eprintln!("$HOME not set");
+                return;
+            };
+            let Ok(string) = std::fs::read_to_string(format!("{home}/.openmina/debug/rsa.priv"))
+            else {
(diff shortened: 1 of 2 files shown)


ledger/src/proofs/transaction.rs
```diff
@@ -4755,7 +4755,12 @@ pub(super) mod tests {
         }
 
         let rsa_private_key = {
-            let Ok(string) = std::fs::read_to_string("~/.openmina/debug/rsa.priv") else {
+            let Ok(home) = std::env::var("HOME") else {
+                eprintln!("$HOME not set");
+                return;
+            };
+            let Ok(string) = std::fs::read_to_string(format!("{home}/.openmina/debug/rsa.priv"))
```
tizoc, line 4762: @sebastiencs we have a workdir variable for this stuff, search for `get_debug_dir` and `get_work_dir`

sebastiencs: Yes I tried it, but during tests it is not set:
```
running 1 test
thread 'proofs::transaction::tests::add_private_key_to_block_proof_input' panicked at core/src/lib.rs:45:24:
Work dir is not set
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test proofs::transaction::tests::add_private_key_to_block_proof_input ... FAILED
```

tizoc: oh duh, didn't see the context.

PR #1025, opened 2025-01-16
transaction logic: fix missing account updates when prevalidating tra…

dkuehr:
This PR fixes a bug in the pre-validation function used by the transaction pool.
We were checking top-level account updates and missing the call lists.

Note: the bug was found with differential fuzzing against the OCaml implementation.


sebastiencs (approved):
Good catch

tizoc (approved):
LGTM

tizoc:
@sebastiencs can you think of any other cases in which the call forest abstraction was being bypassed? (for this specific check I only found two, which are exactly the ones @dkuehr fixed in this PR)

sebastiencs:
Yes I didn't find any other cases than those 2

PR #1028, opened 2025-01-20
Refactor window size update to prevent underflow in stream

0xMimir:
Refactor window size update to use `saturating_sub` instead of `wrapping_sub`


p2p/src/network/yamux/p2p_network_yamux_reducer.rs
```diff
@@ -216,7 +216,8 @@ impl P2pNetworkYamuxState {
                         if let Some(stream) = yamux_state.streams.get_mut(&frame.stream_id) {
                             // must not underflow
                             // TODO: check it and disconnect peer that violates flow rules
-                            stream.window_ours = stream.window_ours.wrapping_sub(data.len() as u32);
+                            stream.window_ours =
```
tizoc, line 219: @0xMimir why would a peer send use more data than what fits in our window? I assume that peer is another rust node and we just don't respect that at the moment?

tizoc (approved):

PR #1032, opened 2025-01-21
feat(yamux): Reuse the recv buffer

tizoc:


0xMimir (approved):
LGTM, thanks for the update

PR #1040, opened 2025-01-27
Fixed the issue with pending queue in yamux

0xMimir:
* Changed window size logic so window size can increase more than 256kb
* Changed logic with outgoing data to send packets with size 0
* Changed logic to send window update after half our window is consumed instead of only 64kb being left

### DIFF
--- a/p2p/src/network/yamux/p2p_network_yamux_reducer.rs
+++ b/p2p/src/network/yamux/p2p_network_yamux_reducer.rs
@@ -136,12 +136,12 @@ impl P2pNetworkYamuxState {
                 }
 
                 match &frame.inner {
-                    YamuxFrameInner::Data(data) => {
+                    YamuxFrameInner::Data(_) => {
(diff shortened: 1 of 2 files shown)


p2p/src/network/yamux/p2p_network_yamux_reducer.rs
```diff
@@ -286,47 +286,38 @@ impl P2pNetworkYamuxState {
...
-                        //     {
-                        //         let dispatcher = state_context.into_dispatcher();
-                        //         let error = P2pNetworkConnectionError::YamuxOverflow(stream_id);
-                        //         dispatcher.push(P2pNetworkSchedulerAction::Error { addr, error });
-                        //     }
-
-                        //     return Ok(());
-                        // }
+                        if let Some(new_window) =
+                            stream.window_theirs.checked_sub(data.len() as u32)
```
tizoc, line 290: @0xMimir unless it is too much of a PITA, it would be better to avoid these truncating casts (here and above for `frame.len()`) when processing external input

PR #1042, opened 2025-01-27
WebRTC sniffer

### LINKED ISSUE #1041: As a developer I want to create a tool to capture and analyze WebRTC traffic, so developers can debug webnode p2p communication.
- [x] Capture UDP datagrams on the specified interface, filtered by the port. Store the traffic persistently. Use `pcap`.
- [x] Parse Ethernet, IP, and UDP Transport headers, use `etherparse`.
- [x] Setup testing environment where nodes communicate WebRTC without browser.
- [ ] Decipher DTLS encryption using keys provided by the node.
- [ ] Reconstruct SCTP state machine. Parse raw data in WebRTC data channels. Store the data persistently in RocksDB.
- [ ] Dissect openmina specific protocols (...).
- [ ] Provide http interface for fetch the data sorted, paged and filtered.
- [ ] Implement tests.

vlad9486:
resolves #1041


node/testing/src/scenarios/mod.rs
```diff
@@ -131,7 +131,7 @@ impl Scenarios {
             Self::SimulationSmall(_) => true,
             Self::SimulationSmallForeverRealTime(_) => true,
             Self::MultiNodePubsubPropagateBlock(_) => true, // in progress
-            Self::P2pSignaling(_) => cfg!(feature = "p2p-webrtc"),
+            Self::P2pSignaling(_) => !cfg!(feature = "p2p-webrtc"),
```
binier, line 134: @vlad9486  please revert this.

cli/src/commands/node/mod.rs
```diff
@@ -216,7 +219,27 @@ impl Node {
...
+                        node::core::log::system_time();
+                        summary = "bad rng seed",
+                        err = err.to_string(),
+                    );
+                    return Err(err);
+                }
+            },
+        };
+        let mut node_builder: NodeBuilder =
+            NodeBuilder::new(custom_rng_seed, daemon_conf, genesis_conf);
```
binier, line 242: Only things that are required should be passed via `NodeBuilder::new`. Since custom rng is optional, please remove it and use similar pattern instead:
```
if let Some(sec_key) = self.p2p_secret_key {
            node_builder.p2p_sec_key(sec_key);
}
```

PR #1054, opened 2025-02-02
Fix: ledger stores more staged ledgers than it needs

binier:
The issue is caused by an empty block, which has the same merkle root hash as it's predecessor.
When cleaning up staged ledgers, we are using merkle_hash, so staged ledger for the replaced block (after fork) stays in there when it shouldn't.


tizoc (approved):
ignoring the CI failures, looks good to me

PR #1065, opened 2025-02-05
Added broadcast delay for snarks and transactions until they are fully vertified

### LINKED ISSUE #1055: As a developer, I want to broadcast snark pool and transactions only after they have been fully validated, to avoid broadcasting invalid ones
Continuation of #952, transactions and snark pool messages from pubsub currently get broadcasted without being validated first, this is not correct behavior and may result in the nodes being banned.

0xMimir:
Resolves #1055, continuation of #1001

### DIFF
--- a/core/src/p2p.rs
+++ b/core/src/p2p.rs
@@ -1,0 +1,16 @@
+/// TODO: These types and methods should be moved to `p2p` crate, they are here because they are used in `snark` crates callbacks
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Copy)]
+pub struct P2pNetworkPubsubMessageCacheId {
+    pub source: libp2p_identity::PeerId,
+    pub seqno: u64,
+}
+
+impl P2pNetworkPubsubMessageCacheId {
+    pub fn to_raw_bytes(&self) -> Vec<u8> {
+        let mut message_id = self.source.to_base58();
+        message_id.push_str(&self.seqno.to_string());
+        message_id.into_bytes()
+    }
+}
--- a/core/src/transaction/mod.rs
+++ b/core/src/transaction/mod.rs
@@ -5,3 +5,36 @@ mod transaction_with_hash;
 pub use transaction_with_hash::*;
 
 pub use mina_p2p_messages::v2::{MinaBaseUserCommandStableV2 as Transaction, TransactionHash};
+
+use crate::{p2p::P2pNetworkPubsubMessageCacheId, requests::RpcId};
+
+/// TODO: Types and methods bellow, should be moved to `node` crate, they are here because they are used in `snark` crates callbacks
+#[derive(serde::Serialize, serde::Deserialize, Debug, Clone, Copy, Default)]
+pub enum TransactionPoolMessageSource {
+    Rpc {
+        id: RpcId,
+    },
+    Pubsub {
+        id: P2pNetworkPubsubMessageCacheId,
+    },
+    #[default]
+    None,
+}
+
+impl TransactionPoolMessageSource {
+    pub fn rpc(id: RpcId) -> Self {
+        Self::Rpc { id }
+    }
+
+    pub fn pubsub(id: P2pNetworkPubsubMessageCacheId) -> Self {
+        Self::Pubsub { id }
+    }
+
+    pub fn is_sender_local(&self) -> bool {
+        matches!(self, Self::Rpc { .. })
+    }
+
+    pub fn is_libp2p(&self) -> bool {
+        matches!(self, Self::Pubsub { .. })
+    }
+}
--- a/node/src/snark_pool/candidate/snark_pool_candidate_actions.rs
+++ b/node/src/snark_pool/candidate/snark_pool_candidate_actions.rs
@@ -49,7 +49,7 @@ pub enum SnarkPoolCandidateAction {
     WorkVerifyError {
         peer_id: PeerId,
         verify_id: SnarkWorkVerifyId,
-        batch: Vec<Snark>,
+        batch: Vec<SnarkJobId>,
     },
     WorkVerifySuccess {
         peer_id: PeerId,
--- a/node/src/snark_pool/candidate/snark_pool_candidate_actions.rs
+++ b/node/src/snark_pool/candidate/snark_pool_candidate_actions.rs
@@ -49,6 +49,7 @@ pub enum SnarkPoolCandidateAction {
     WorkVerifyError {
         peer_id: PeerId,
         verify_id: SnarkWorkVerifyId,
+        batch: Vec<Snark>,
     },
     WorkVerifySuccess {
         peer_id: PeerId,
--- a/node/src/snark_pool/candidate/snark_pool_candidate_reducer.rs
+++ b/node/src/snark_pool/candidate/snark_pool_candidate_reducer.rs
@@ -160,7 +160,12 @@ impl SnarkPoolCandidatesState {
                     reason: P2pDisconnectionReason::SnarkPoolVerifyError,
                 });
 
-                // TODO: this is incorrect, only one snark in batch could be invalid but we would reject that whole batch
+                // TODO: This is not correct. We are rejecting all snark messages, but the fact that the batch
+                // failed to verify means that there is at least one invalid snark in the batch, not that all of them
+                // are invalid.
+                // Instead, what should happen here is that we split the batch in two and try to verify the two batches
+                // again. Repeating until batches don't fail to verify anymore, or each batch is of size 1.
+                // It may also be worth capping the batch sizes to 10.
                 for snark_job_id in batch {
                     dispatcher.push(P2pNetworkPubsubAction::RejectMessage {
                         message_id: Some(BroadcastMessageId::Snark {
--- a/node/src/snark_pool/candidate/snark_pool_candidate_reducer.rs
+++ b/node/src/snark_pool/candidate/snark_pool_candidate_reducer.rs
@@ -1,7 +1,7 @@
 use std::collections::BTreeMap;
 
 use crate::{p2p_ready, SnarkPoolAction};
-use openmina_core::snark::Snark;
+use openmina_core::snark::{Snark, SnarkJobId};
 use p2p::{
     channels::rpc::{P2pChannelsRpcAction, P2pRpcId, P2pRpcRequest},
     disconnection::{P2pDisconnectionAction, P2pDisconnectionReason},
@@ -124,7 +124,7 @@ impl SnarkPoolCandidatesState {
                             }
                         }),
                     on_error: redux::callback!(
-                        on_snark_pool_candidate_work_verify_error((req_id: SnarkWorkVerifyId, sender: String, batch: Vec<Snark>)) -> crate::Action {
+                        on_snark_pool_candidate_work_verify_error((req_id: SnarkWorkVerifyId, sender: String, batch: Vec<SnarkJobId>)) -> crate::Action {
                             SnarkPoolCandidateAction::WorkVerifyError {
                                 peer_id: sender.parse().unwrap(),
                                 verify_id: req_id,
@@ -160,10 +160,10 @@ impl SnarkPoolCandidatesState {
                     reason: P2pDisconnectionReason::SnarkPoolVerifyError,
                 });
 
-                for snark in batch {
+                for snark_job_id in batch {
                     dispatcher.push(P2pNetworkPubsubAction::RejectMessage {
                         message_id: Some(BroadcastMessageId::Snark {
-                            job_id: snark.job_id(),
+                            job_id: snark_job_id.clone(),
                         }),
                         peer_id: None,
                         reason: "Snark work verification failed".to_string(),
--- a/node/src/snark_pool/candidate/snark_pool_candidate_reducer.rs
+++ b/node/src/snark_pool/candidate/snark_pool_candidate_reducer.rs
@@ -5,7 +5,7 @@ use openmina_core::snark::Snark;
 use p2p::{
     channels::rpc::{P2pChannelsRpcAction, P2pRpcId, P2pRpcRequest},
     disconnection::{P2pDisconnectionAction, P2pDisconnectionReason},
-    PeerId,
+    BroadcastMessageId, P2pNetworkPubsubAction, PeerId,
 };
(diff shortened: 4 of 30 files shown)


node/src/snark_pool/candidate/snark_pool_candidate_actions.rs
```diff
@@ -49,6 +49,7 @@ pub enum SnarkPoolCandidateAction {
     WorkVerifyError {
         peer_id: PeerId,
         verify_id: SnarkWorkVerifyId,
+        batch: Vec<Snark>,
```
tizoc, line 52: @0xMimir what about just passing the job ids here? Seems the whole snark is not actually needed

p2p/src/network/pubsub/p2p_network_pubsub_state.rs
```diff
@@ -368,10 +385,33 @@ impl P2pNetworkPubsubMessageCache {
...
+                self.map
+                    .iter_mut()
+                    .find_map(|(message_id, message)| match message {
+                        P2pNetworkPubsubMessageCacheMessage::PreValidatedSnark {
+                            job_id, ..
+                        } if job_id == snark_job_id => Some((*message_id, message)),
+                        _ => None,
+                    })
+            }
+            BroadcastMessageId::Transaction { tx } => {
```
tizoc, line 404: what happens if I receive two different diffs containing the same transaction? I don't think you can map to the original message by just using a single transaction hash

0xMimir: Yes, I will rewrite, to update all transaction pending messages in pubsub that contain that transaction id, and then if all messages that have all transactions vertified to broadcast them.

node/src/transaction_pool/transaction_pool_reducer.rs
```diff
@@ -300,9 +319,22 @@ impl TransactionPoolState {
                 if let Some(rpc_action) = rpc_action {
                     dispatcher.push(rpc_action);
                 }
-                // TODO: libp2p logic already broadcasts everything right now and doesn't
-                // wait for validation, thad needs to be fixed. See #952
+
                 if was_accepted {
```
tizoc, line 323: something should be done too if it wasn't accepted

node/src/transaction_pool/transaction_pool_reducer.rs
```diff
@@ -300,14 +319,35 @@ impl TransactionPoolState {
...
+                            },
+                        });
+                    }
                     dispatcher.push(TransactionPoolAction::Rebroadcast {
                         accepted,
                         rejected,
                         is_local: is_sender_local,
                     });
+                } else {
+                    for (tx, _) in rejected {
```
tizoc, line 344: @0xMimir this is not correct, because the full diff may have been rejected because of one particular transaction, but here we are reacting to every rejected transaction.

I have to think about it more, but I am unsure about the approach in general. This probably should work on the diff level, not per transaction.

tizoc: To add a bit more context, diffs will only be rejected if one of the transactions is rejected because of one of these reasons:

https://github.com/openmina/openmina/blob/b152745fce013d2cee65f78f9a51ae1e776d0726/ledger/src/transaction_pool.rs#L239

0xMimir: I could filter by error if tx error is `grounds_for_diff_rejection`, reject it otherwise continue

tizoc: The "diff" is lost here:

https://github.com/openmina/openmina/blob/develop/p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs#L577-L586

I guess it was implemented that way to make it more compatible with the way the webrtc layer works, but that makes it harder to work at the diff level. It also means that the transactions will be verified one by one instead of batched (in webrtc they are batched [here](https://github.com/openmina/openmina/blob/44ba15084f97b13f360ba4b88145ba07a78e638c/node/src/transaction_pool/candidate/transaction_pool_candidate_reducer.rs#L98, but that will not help the libp2p part).

You can also see here that the action dispatched to verify the transactions also supports multiple transactions (even if the callback always constructs the call using a single transaction):

https://github.com/openmina/openmina/blob/44ba15084f97b13f360ba4b88145ba07a78e638c/node/src/state.rs#L535-L538

I don't know if in practice we receive more than one transaction per diff from libp2p's gossip btw (you could add some logging and run the node for some time to check).

So:
- `P2pChannelsTransactionAction::Libp2pReceived` should contain all the transactions, not just one
- `on_p2p_channels_transaction_libp2p_received` callback should receive the full list of transactions, not just one. And then dispatch `StartVerify` with all the transactions at once.
- `StartVerify` and the related actions should probably contain the message id (as an `Option` similar to how it contains a reference to the rpc from where the transaction may have come from if it was sent by the user directly using the API)
- Then at the end of the sequence when handling `ApplyVerifiedDiffWithAccounts` you will have access to the actual message id (if any) from which those transactions came from.

Btw those actions probably need to be split in more, but lets leave that for later

tizoc: Another idea is that instead of having `from_rpc` and `from_gossip` separately (because you cannot have both), there could be a single `from_origin` that is either `FromRpc(RpcId) | FromGossip(MessageId) | Other` so that you can react accordingly depending on the origin.

tizoc: One last thing, and this should be left for later. But transactions should first go into the  `TransactionPoolState::candidates` and with `TransactionPoolCandidatesState` extended to support storing the libp2p received messages grouped by message id. The logic should also be extended to handle libp2p gossip as an origin. This will make it much easier to later fix the issue of us trying to validate the pool before the node is even synced.

With that approach we would still receive and store the gossip messages but wait until we are synced to validate them.

Anyway, this is for later, first thing to solve is what I mentioned in the above comments.

0xMimir: I have updated logic in pubsub and transaction pool, to use message id instead of transaction hashes, also I have been running node for almost a day now and every time transaction diff, had only one transaction.

tizoc: Ok, thats useful to know (still we need to consider multiple transactions because the protocol allows it)

node/src/transaction_pool/transaction_pool_reducer.rs
```diff
@@ -315,22 +307,37 @@ impl TransactionPoolState {
...
+                        dispatcher.push(P2pNetworkPubsubAction::RejectMessage {
+                            message_id: Some(BroadcastMessageId::MessageId { message_id: id }),
+                            peer_id: None,
+                            reason: "Rejected transaction diff".to_owned(),
+                        });
+                    }
+                    (_, TransactionPoolMessageSource::None) => {}
                 }
 
                 if was_accepted {
```
tizoc, line 340: should this be a match above for `(true, TransactionPoolMessageSource::None)` instead? we don't want rebroadcast if we pushed any of the `P2pNetworkPubsubAction` actions above right?

tizoc: Actually, it should happen for the RPC source too. Just not for the case in which the source is the libp2p gossip

node/src/transaction_pool/transaction_pool_reducer.rs
```diff
@@ -297,17 +302,48 @@ impl TransactionPoolState {
...
-                // TODO: libp2p logic already broadcasts everything right now and doesn't
-                // wait for validation, thad needs to be fixed. See #952
+
                 if was_accepted {
-                    dispatcher.push(TransactionPoolAction::Rebroadcast {
-                        accepted,
-                        rejected,
-                        is_local: is_sender_local,
-                    });
+                    match from_source {
```
tizoc, line 336: what about an `is_libp2p` method similar to `is_sender_local` for this check to avoid the need to match here? (contents are not used anyway).

Btw I will check later today, but overall the current version looks good

core/src/p2p.rs
```diff
@@ -0,0 +1,15 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Copy)]
+pub struct P2pNetworkPubsubMessageCacheId {
+    pub source: libp2p_identity::PeerId,
+    pub seqno: u64,
+}
+
+impl P2pNetworkPubsubMessageCacheId {
```
tizoc, line 9: @0xMimir does this belong here? I seems it belongs to the `p2p` crate, since this is p2p specific, and not part of `core`. I would move this there and get rid of the `libp2p-identity` dependency from `core`.

core/src/transaction/mod.rs
```diff
@@ -5,3 +5,35 @@ mod transaction_with_hash;
 pub use transaction_with_hash::*;
 
 pub use mina_p2p_messages::v2::{MinaBaseUserCommandStableV2 as Transaction, TransactionHash};
+
+use crate::{p2p::P2pNetworkPubsubMessageCacheId, requests::RpcId};
+
+#[derive(serde::Serialize, serde::Deserialize, Debug, Clone, Copy, Default)]
+pub enum TransactionPoolMessageSource {
```
tizoc, line 12: Same here, doesn't seem this belongs to `core`, this is specific to the transaction pool stuff in `node`, I would move it there.

node/src/snark_pool/snark_pool_reducer.rs
```diff
@@ -202,13 +205,17 @@ impl SnarkPoolState {
...
 
-                // TODO: libp2p logic already broadcasts everything right now and doesn't
-                // wait for validation, thad needs to be fixed. See #952
                 dispatcher.push(P2pChannelsSnarkAction::Libp2pBroadcast {
                     snark: snark.clone(),
                     nonce: 0,
                     is_local: *is_sender_local,
                 });
+
+                dispatcher.push(P2pNetworkPubsubAction::BroadcastValidatedMessage {
```
tizoc, line 214: This doesn't feel right, how come we have both this and `Libp2pBroadcast` above ? Please revise the flow, because `Libp2pBroadcast` eventually ends dispatching `BroadcastValidatedMessage` too.

tizoc (changes requested):

core/src/transaction/mod.rs
0xMimir: Both `P2pNetworkPubsubMessageCacheId` and `TransactionPoolMessageSource` should be in `p2p` and `node` crate, but `TransactionPoolMessageSource` is required in `snark` crate for callbacks

tizoc: ok I see. There are still ways around that, but for now just add a `TODO` comment explaining this.

node/src/snark_pool/snark_pool_reducer.rs
```diff
@@ -202,13 +206,20 @@ impl SnarkPoolState {
...
-                    snark: snark.clone(),
-                    nonce: 0,
-                    is_local: *is_sender_local,
-                });
+                let broadcast_action = P2pNetworkPubsubAction::BroadcastValidatedMessage {
+                    message_id: BroadcastMessageId::Snark {
+                        job_id: snark.job_id(),
+                    },
+                };
+                if broadcast_action.is_enabled(global_state, meta.time()) {
```
tizoc, line 214: This logic here is not clear to me. Either it needs a comment explaining well what is going on and why (same for every case where a non-standard patterns shows up), or something is off and it needs to be implemented some other way.

0xMimir: I have added a comment, logic is that if `P2pNetworkPubsubAction::BroadcastValidatedMessage` gets dispatched it means we have received that snark from pubsub, otherwise we have either received it from webrtc or we have produced it.

node/src/snark_pool/candidate/snark_pool_candidate_reducer.rs
```diff
@@ -154,6 +159,17 @@ impl SnarkPoolCandidatesState {
                     peer_id,
                     reason: P2pDisconnectionReason::SnarkPoolVerifyError,
                 });
+
+                // TODO: this is incorrect, only one snark in batch could be invalid but we would reject that whole batch
```
tizoc, line 163:
```suggestion
                // TODO: This is not correct. We are rejecting all snark messages, but the fact that the batch
                // failed to verify means that there is at least one invalid snark in the batch, not that all of them
                // are invalid.
                // Instead, what should happen here is that we split the batch in two and try to verify the two batches
                // again. Repeating until batches don't fail to verify anymore, or each batch is of size 1.
                // It may also be worth capping the batch sizes to 10.
```

p2p/src/network/pubsub/p2p_network_pubsub_actions.rs
```diff
@@ -182,6 +182,10 @@ pub enum P2pNetworkPubsubAction {
     BroadcastValidatedMessage {
         message_id: BroadcastMessageId,
     },
+
+    BroadcastMessage {
```
tizoc, line 186: Would be great to have a comment on this one, it is not clear how it is different from `BroadcastValidatedMessage`

p2p/src/network/pubsub/p2p_network_pubsub_reducer.rs
```diff
@@ -687,6 +672,31 @@ impl P2pNetworkPubsubState {
...
+            P2pNetworkPubsubAction::BroadcastMessage { message_id } => {
+                let Some(message) = pubsub_state.mcache.map.get(&message_id) else {
+                    bug_condition!("Message with id: {:?} not found", message_id);
+                    return Ok(());
+                };
+
+                let raw_message = message.message().clone();
+                let peer_id = *message.peer_id();
+
+                pubsub_state.reduce_incoming_validated_message(message_id, peer_id, &raw_message);
```
tizoc, line 684: `fn reduce_incoming_validated_message` would benefit from a comment explaining on a high level what it does (similar to `reduce_incoming_message` but doesn't need to be as extense). Could be this:

```
/// Queues a validated message for propagation to other peers in the pubsub network.
/// For peers that are "on mesh" for the message's topic, queues the full message.
/// For other peers, queues an IHAVE control message to notify about message availability.
/// The original sender is excluded from propagation.
```

tizoc (approved):
LGTM (left some comments re: comments to add, please add them before merging)

p2p/src/network/pubsub/p2p_network_pubsub_actions.rs
0xMimir: I have removed this action as it was no longer needed

Issue #1067, opened 2025-02-07
As a developer, I want to change transaction validation flow, in order to not reject transactions during nodes bootstrap


0xMimir:
Transactions should first go into the `TransactionPoolState::candidates` and with `TransactionPoolCandidatesState` extended to support storing the libp2p received messages grouped by message id. The logic should also be extended to handle libp2p gossip as an origin. This will make it much easier to later fix the issue of us trying to validate the pool before the node is even synced.

With that approach we would still receive and store the gossip messages but wait until we are synced to validate them.

https://github.com/openmina/openmina/pull/1065#discussion_r1943744306

PR #1079, opened 2025-02-17
Changed action flow for transaction in pubsub

0xMimir:
Solves: #1067

* Updated `TransactionPoolCandidatesState` to have transactions grouped by pubsub message id
* Changed flow for transactions received in pubsub to be stored in candidate state before before being processed

### DIFF
--- a/node/src/transaction_pool/candidate/transaction_pool_candidate_actions.rs
+++ b/node/src/transaction_pool/candidate/transaction_pool_candidate_actions.rs
@@ -1,4 +1,6 @@
-use openmina_core::transaction::{TransactionHash, TransactionInfo, TransactionWithHash};
+use openmina_core::transaction::{
+    TransactionHash, TransactionInfo, TransactionPoolMessageSource, TransactionWithHash,
+};
 use openmina_core::ActionEvent;
 use p2p::P2pNetworkPubsubMessageCacheId;
 use serde::{Deserialize, Serialize};
@@ -50,7 +52,7 @@ pub enum TransactionPoolCandidateAction {
         peer_id: PeerId,
         transaction_hashes: Vec<TransactionHash>,
         verify_id: (),
-        from_source: Option<P2pNetworkPubsubMessageCacheId>,
+        from_source: TransactionPoolMessageSource,
     },
     VerifyError {
         peer_id: PeerId,
@@ -59,7 +61,7 @@ pub enum TransactionPoolCandidateAction {
     VerifySuccess {
         peer_id: PeerId,
         verify_id: (),
-        from_source: Option<P2pNetworkPubsubMessageCacheId>,
+        from_source: TransactionPoolMessageSource,
     },
     PeerPrune {
         peer_id: PeerId,
--- a/node/src/transaction_pool/candidate/transaction_pool_candidate_actions.rs
+++ b/node/src/transaction_pool/candidate/transaction_pool_candidate_actions.rs
@@ -111,6 +111,7 @@ impl redux::EnablingCondition<crate::State> for TransactionPoolCandidateAction {
                 .is_some(),
             TransactionPoolCandidateAction::Libp2pTransactionsReceived { .. } => true,
             TransactionPoolCandidateAction::VerifyNext => {
+                // TODO: if a block is being applied or produced, skip this action too
                 state.transition_frontier.sync.is_synced()
             }
             TransactionPoolCandidateAction::VerifyPending {
--- a/node/src/transaction_pool/candidate/transaction_pool_candidate_actions.rs
+++ b/node/src/transaction_pool/candidate/transaction_pool_candidate_actions.rs
@@ -1,5 +1,6 @@
 use openmina_core::transaction::{TransactionHash, TransactionInfo, TransactionWithHash};
 use openmina_core::ActionEvent;
+use p2p::P2pNetworkPubsubMessageCacheId;
 use serde::{Deserialize, Serialize};
 
 use crate::p2p::channels::rpc::P2pRpcId;
@@ -37,12 +38,19 @@ pub enum TransactionPoolCandidateAction {
         peer_id: PeerId,
         transaction: TransactionWithHash,
     },
+    /// Callback for transactions received over pubsub
+    Libp2pTransactionsReceived {
+        peer_id: PeerId,
+        transactions: Vec<TransactionWithHash>,
+        message_id: P2pNetworkPubsubMessageCacheId,
+    },
     #[action_event(level = trace)]
     VerifyNext,
     VerifyPending {
         peer_id: PeerId,
         transaction_hashes: Vec<TransactionHash>,
         verify_id: (),
+        from_source: Option<P2pNetworkPubsubMessageCacheId>,
     },
     VerifyError {
         peer_id: PeerId,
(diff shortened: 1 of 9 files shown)


node/src/transaction_pool/candidate/transaction_pool_candidate_actions.rs
```diff
@@ -37,12 +38,18 @@ pub enum TransactionPoolCandidateAction {
         peer_id: PeerId,
         transaction: TransactionWithHash,
     },
+    TransactionsReceived {
```
tizoc, line 41: @0xMimir I think this should have some mention of libp2p in the name, or at least a comment making it explicit that this is libp2p specific

node/src/transaction_pool/candidate/transaction_pool_candidate_state.rs
```diff
@@ -203,10 +209,59 @@ impl TransactionPoolCandidatesState {
...
+
+        self.by_message_id
+            .insert(message_id, (peer_id, transaction_hashes));
+
+        transactions.into_iter().for_each(|transaction| {
+            self.transaction_received(time, peer_id, transaction);
+        })
+    }
+
+    pub fn get_batch_to_verify(
```
tizoc, line 233: This needs more comments, or for the inner blocks to be better abstracted into their own methods so that it is very clear what is going on. Right now the logic is quite complicated and not obvious.

node/src/transaction_pool/candidate/transaction_pool_candidate_actions.rs
```diff
@@ -37,12 +38,19 @@ pub enum TransactionPoolCandidateAction {
...
+        transactions: Vec<TransactionWithHash>,
+        message_id: P2pNetworkPubsubMessageCacheId,
+    },
     #[action_event(level = trace)]
     VerifyNext,
     VerifyPending {
         peer_id: PeerId,
         transaction_hashes: Vec<TransactionHash>,
         verify_id: (),
+        from_source: Option<P2pNetworkPubsubMessageCacheId>,
```
tizoc, line 53: Somethig is off, here and on success the field is called `from_source` but only pubsub is valid, it is not a `TransactionPoolMessageSource` like in other places.

tizoc: `get_batch_to_verify` should return the source

node/src/transaction_pool/candidate/transaction_pool_candidate_state.rs
```diff
@@ -15,10 +16,17 @@ use crate::p2p::PeerId;
 static EMPTY_PEER_TX_CANDIDATES: BTreeMap<TransactionHash, TransactionPoolCandidateState> =
     BTreeMap::new();
 
+type NextBatch = Option<(
+    PeerId,
+    Vec<TransactionWithHash>,
+    Option<P2pNetworkPubsubMessageCacheId>,
```
tizoc, line 22: As mentioned above, this should have the actual source. Also I would not make this an `Option` alias, lets define a proper struct here and have `get_batch_to_verify` return `Option`

node/src/transaction_pool/candidate/transaction_pool_candidate_state.rs
```diff
@@ -219,17 +264,43 @@ impl TransactionPoolCandidatesState {
...
+            })
+            .cloned()
+            .collect();
+
+        Some((
+            *peer_id,
+            transactions,
+            TransactionPoolMessageSource::pubsub(*message_id),
+        ))
+    }
```
tizoc, line 303: annoying nitpick on my side, but lets add a newline between these two methods

node/src/transaction_pool/candidate/transaction_pool_candidate_state.rs
```diff
@@ -203,10 +216,42 @@ impl TransactionPoolCandidatesState {
...
+        transactions.into_iter().for_each(|transaction| {
+            self.transaction_received(time, peer_id, transaction);
+        })
+    }
+
+    /// Get next batch of transactions to verify,
+    /// first checks if there are any transactions to verify from pubsub
+    /// after that checks for transactions from peers
+    pub fn get_batch_to_verify(&self) -> Option<NextBatch> {
+        if let Some(batch) = self.next_batch_from_pubsub() {
```
tizoc, line 244: what about something like this instead?

```rust
       self.next_batch_from_pubsub().or_else(|| self.next_batch_from_peers())
```

node/src/transaction_pool/candidate/transaction_pool_candidate_state.rs
```diff
@@ -344,6 +429,20 @@ impl TransactionPoolCandidatesState {
             !peers.is_empty()
         })
     }
+
+    pub fn remove_pubsub_transactions(&mut self, message_id: &P2pNetworkPubsubMessageCacheId) {
```
tizoc, line 433: This is defined but never called anywhere?

tizoc: Seems to be handled by `verify_result` above already.

node/src/transaction_pool/candidate/transaction_pool_candidate_state.rs
```diff
@@ -259,17 +330,31 @@ impl TransactionPoolCandidatesState {
...
-        if let Some(peer_transactions) = self.by_peer.get_mut(peer_id) {
-            let txs_to_remove = peer_transactions
-                .iter()
-                .filter(|(_, job_state)| job_state.pending_verify_id() == Some(verify_id))
-                .map(|(hash, _)| hash.clone())
-                .collect::<Vec<_>>();
-
-            for hash in txs_to_remove {
-                self.transaction_remove(&hash);
+        match from_source {
```
tizoc, line 336: Btw why is this called `verify_result` ? It is not verifying anything, it is just cleaning up state.

tizoc: I guess it is trying to follow the same API as the snark candidates state machine, but it is confusing.

node/src/transaction_pool/candidate/transaction_pool_candidate_reducer.rs
```diff
@@ -90,38 +89,53 @@ impl TransactionPoolCandidatesState {
...
             }
             TransactionPoolCandidateAction::VerifyPending {
                 peer_id,
                 transaction_hashes,
                 verify_id,
+                from_source,
             } => {
                 state.verify_pending(meta.time(), peer_id, *verify_id, transaction_hashes);
                 let dispatcher = state_context.into_dispatcher();
                 dispatcher.push(TransactionPoolCandidateAction::VerifySuccess {
```
tizoc, line 135: Something else I am confused about is that we set the verification state to pending then immediately dispatch a success.

tizoc: Overall the flow looks right, because these stop being candidates, and we want to cleanup that state. It is just that the naming and the way it is modeled here is confusing. Something to improve later.

tizoc:
@0xMimir btw have you tested this by running the node locally? does it behave as expected? (no transactions are verified until the node is in sync, but once in sync those get verified)

node/src/transaction_pool/candidate/transaction_pool_candidate_state.rs
0xMimir: I am not 100% sure but it either seems that it is unfinished, because action `TransactionPoolCandidateAction::VerifyError` is never called and just has `todo!()`, or following the same API like you mentioned

tizoc: Indeed. We can take care of that later.

PR #1090, opened 2025-02-26
Transition Frontier: do cheap consensus first then expensive verification

binier:
After receiving a new block from peer, check consensus first, if it's better than the current best tip, only then proceed to do a more expensive proof verification.

re: #998

### DIFF
--- a/node/src/transition_frontier/candidate/transition_frontier_candidate_actions.rs
+++ b/node/src/transition_frontier/candidate/transition_frontier_candidate_actions.rs
@@ -1,16 +1,12 @@
-use std::sync::Arc;
-
-use mina_p2p_messages::v2::{MinaBlockBlockStableV2, StateHash};
+use mina_p2p_messages::v2::StateHash;
 use openmina_core::block::prevalidate::BlockPrevalidationError;
-use openmina_core::block::{ArcBlockWithHash, BlockWithHash};
+use openmina_core::block::ArcBlockWithHash;
 use openmina_core::consensus::consensus_take;
 use openmina_core::{action_event, ActionEvent};
 use serde::{Deserialize, Serialize};
 use snark::block_verify::SnarkBlockVerifyError;
 
 use crate::snark::block_verify::SnarkBlockVerifyId;
-
-use super::TransitionFrontierCandidateStatus;
 
 pub type TransitionFrontierCandidateActionWithMeta =
     redux::ActionWithMeta<TransitionFrontierCandidateAction>;
@@ -22,10 +18,11 @@ pub type TransitionFrontierCandidateActionWithMetaRef<'a> =
 #[derive(Serialize, Deserialize, Debug, Clone, ActionEvent)]
 #[action_event(level = debug, fields(debug(hash), debug(error)))]
 pub enum TransitionFrontierCandidateAction {
-    #[action_event(level = info)]
+    P2pBestTipUpdate {
+        best_tip: ArcBlockWithHash,
+    },
     BlockReceived {
-        hash: StateHash,
-        block: Arc<MinaBlockBlockStableV2>,
+        block: ArcBlockWithHash,
         chain_proof: Option<(Vec<StateHash>, ArcBlockWithHash)>,
     },
     BlockPrevalidateSuccess {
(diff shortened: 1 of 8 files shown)


node/src/transition_frontier/candidate/transition_frontier_candidate_actions.rs
```diff
@@ -22,10 +20,12 @@ pub type TransitionFrontierCandidateActionWithMetaRef<'a> =
...
 #[action_event(level = debug, fields(debug(hash), debug(error)))]
 pub enum TransitionFrontierCandidateAction {
+    P2pBestTipUpdate {
+        best_tip: BlockWithHash<Arc<MinaBlockBlockStableV2>>,
+    },
     #[action_event(level = info)]
     BlockReceived {
-        hash: StateHash,
-        block: Arc<MinaBlockBlockStableV2>,
+        block: ArcBlockWithHash,
```
tizoc, line 28: @binier can this be made to print the block hash? (with some custom block printer I guess)

one nice property of the old version was that the hash was printed in the logs

tizoc:
```
2025-02-26T19:36:19.399621Z  INFO openmina_core::log::action: time="1740598579338852958" kind="TransitionFrontierCandidateBlockReceived" hash=3NKQ9j3CkMhCXH4Sdsf25DMUJhi6iyqTEhpya54XBzTkxuTAHEm4
```

node/src/transition_frontier/candidate/transition_frontier_candidate_actions.rs
```diff
@@ -52,131 +52,50 @@ pub enum TransitionFrontierCandidateAction {
...
-        hash: StateHash,
-    },
-    LongRangeForkResolve {
-        hash: StateHash,
-    },
     #[action_event(level = info)]
-    BestTipUpdate {
-        hash: StateHash,
-    },
     TransitionFrontierSyncTargetUpdate,
```
tizoc, line 56: should `TransitionFrontierSyncTargetUpdate` contain level=info? seems accidental

node/src/transition_frontier/candidate/transition_frontier_candidate_actions.rs
binier: Moved to logger_effects: https://github.com/openmina/openmina/blob/3305941aa2668b30d8fdd1dda27a97d6eaf0e56b/node/src/logger/logger_effects.rs#L123

node/src/transition_frontier/candidate/transition_frontier_candidate_actions.rs
binier: fixed

node/src/transition_frontier/candidate/transition_frontier_candidate_state.rs
```diff
@@ -89,113 +71,189 @@ impl TransitionFrontierCandidateStatus {
...
 
-    pub fn best_tip_block_with_hash(&self) -> Option<BlockWithHash<Arc<MinaBlockBlockStableV2>>> {
-        let hash = self.best_tip.as_ref()?;
-        let block = self.blocks.get(hash)?;
-        Some(BlockWithHash {
-            hash: hash.clone(),
-            block: block.block.clone(),
-        })
+    pub fn contains(&self, hash: &StateHash) -> bool {
+        self.invalid.contains_key(hash) || self.get(hash).is_none()
```
tizoc, line 136: why `is_none` ? shouldn't be `is_some` instead?

binier: fixed

node/src/transition_frontier/candidate/transition_frontier_candidate_state.rs
```diff
@@ -89,113 +71,189 @@ impl TransitionFrontierCandidateStatus {
...
+
+                has_reached_best_candidate = true;
+            }
+
+            has_reached_best_candidate
+        });
+    }
+
+    pub(super) fn best(&self) -> Option<&TransitionFrontierCandidateState> {
+        self.ordered.last()
```
tizoc, line 225: Just a random comment: It makes perfect sense for the better blocks to be `>` the inferior ones, but for some reason I hate that the best one ends up being last in the collection

node/src/transition_frontier/candidate/transition_frontier_candidate_actions.rs
```diff
@@ -22,10 +20,11 @@ pub type TransitionFrontierCandidateActionWithMetaRef<'a> =
 #[derive(Serialize, Deserialize, Debug, Clone, ActionEvent)]
 #[action_event(level = debug, fields(debug(hash), debug(error)))]
 pub enum TransitionFrontierCandidateAction {
-    #[action_event(level = info)]
+    P2pBestTipUpdate {
+        best_tip: BlockWithHash<Arc<MinaBlockBlockStableV2>>,
```
tizoc, line 24: @binierbtw isn't this the same as ArcBlockWithHash?

node/src/transition_frontier/candidate/transition_frontier_candidate_state.rs
```diff
@@ -89,113 +71,189 @@ impl TransitionFrontierCandidateStatus {
...
 pub struct TransitionFrontierCandidatesState {
-    pub blocks: BTreeMap<StateHash, TransitionFrontierCandidateState>,
-    // TODO(binier): rename to best candidate. Best tip will be in transition_frontier state.
-    pub best_tip: Option<StateHash>,
-    pub best_tip_chain_proof: Option<(Vec<StateHash>, ArcBlockWithHash)>,
+    /// Maintains an ordered list of transition frontier Candidates,
+    /// ordered using consensus rules worst to best.
+    ordered: BTreeSet<TransitionFrontierCandidateState>,
+    /// Invalid candidates.
+    invalid: BTreeMap<StateHash, u32>,
```
tizoc, line 127: would it be worth it to add the reason of the invalidation?

tizoc (approved):
Left some comments, but LGTM, and cleaner.

node/src/transition_frontier/candidate/transition_frontier_candidate_actions.rs
binier: yeah

node/src/transition_frontier/candidate/transition_frontier_candidate_state.rs
binier: those that fail validations. I'll expand on it

node/src/transition_frontier/candidate/transition_frontier_candidate_state.rs
binier: It's consistent with how blocks are stored in transition frontier as well (best_tip is last), so it should be fine.

node/src/transition_frontier/candidate/transition_frontier_candidate_state.rs
tizoc: Oh I didn't mean the comment, but to keep information about what was the reason for the failure (block too early/late? invalid proof? invalid protocol version? etc)

tizoc: Btw, from the prevalidation errors there are two that are temporary and recoverable: `GenesisNotReady` and `ReceivedTooEarly`

PR #1092, opened 2025-02-27
Fix(transition_frontier/candidate): don't forever invalidate early received blocks

binier:
As pointed out by @tizoc , this pr: #1090 introduced a bug which would cause too early received blocks to be invalidated forever.
This pr fixes that.


tizoc (approved):
LGTM

PR #1101, opened 2025-02-28
feat(pool-candidate): Don't proceed to verify transactions snarks when producing or applying a block

tizoc:


binier (commented):
I'd remove is_synced part, because we might be temporarily stuck in sync if e.g. we cant fetch predecessor in case of a fork.

tizoc:
@binier but we want that, we don't want to process transactions if we are not at the best tip already, thats why that condition was added there (but it is not part of the deployed version of the webnode, thats why we still see that delay)

tizoc:
The other reason is that during the initial sync we don't want to process transactions either. Other than them probably having been created on top of a more recent state, we will need to re-process them all anyway after the best tip changes

binier:
> @binier but we want that, we don't want to process transactions if we are not at the best tip already, thats why that condition was added there (but it is not part of the deployed version of the webnode, thats why we still see that delay)

Real problem is that it can be abused by malicious block producer to freeze the mempool.

binier:
> The other reason is that during the initial sync we don't want to process transactions either. Other than them probably having been created on top of a more recent state, we will need to re-process them all anyway after the best tip changes

For this we can instead change the check to `is_synced || !best_tip.is_genesis()`.

tizoc:
> Real problem is that it can be abused by malicious block producer to freeze the mempool.

Can you elaborate on that?

binier:
> Can you elaborate on that?

E.g. he can produce a block which forks best tip on 2 heights. So that block makes it into the sync state, but we never manage to fetch it's parent (because it doesn't send us) to actually apply it.

tizoc:
@binier I may be misunderstanding what you are describing. The idea is that for a best tip at height 10 he would produce two blocks at 11, broadcast only one and keep the other "hidden", and then produce block 12 on top of the hidden 11 and never make the now-required 11 available to others?

He would need to have an enormous amount of stake to do that, right? (must be able to produce two winning blocks for two consecutive heights) in which case blocking other node's transaction pool for a few minutes is the lesser of the issues. And in general it would be costly (you are sacrificing a block) and not that effective.

binier:
@tizoc no he doesnt need to produce 2 blocks just 1. Best tip forks are very common so he can simply choose to build on top of a block which isnt current best tip. And since we dont maintain all forks, that predecessor block wont be available in the network.

tizoc:
ok I see, so something like this?

```
Main Chain: [B8 (available)] -- [B9 (available)] (current best tip)
             |
Fork:        +----------------- [B9a (unavailable)] -- [B10a (broadcasted)] (produced block)
```

`B9` and `B9a` were both candidates at height 9, with `B9` being the superior block. But the block producer decided to build on top of `B9a` (which he saved) without making it available.

But `B10a` must be better than other `B10` alternatives (of for there to be none). And none of the peers (or peers of its peers) must have `B9a` saved (even if not applied, candidates are still kept around for some time right?), because if they do they can just apply it and continue.

I think it is not very easy to pull, the issue it causes is not that serious (it is not a freeze but a delay, and only to immediate peers) and is very costly for the producer (he must sacrifice a block) **BUT** we can still avoid it with a logic that is a bit more complicated right? let me think about it.

binier:
@tizoc Yes exactly that case. I don't think it's hard to pull off (I'm assuming only Rust nodes in the network as it should be self-sustainable). Block producer just needs to time it right.

> it is not a freeze but a delay

Yes it's a delay you are right.

> and only to immediate peers

True, but those immediate peers could be the whole network unless network gets big enough.

> is very costly for the producer (he must sacrifice a block)

Not really, he can still broadcast it just with a huge delay. But that does become harder to pull off. Needs an intricate knowledge of how our transition frontier works.

tizoc:
> This solution should be enough, no?
> https://github.com/openmina/openmina/pull/1101#issuecomment-2691144064

Sounds good except `is_synced` is not really necessary (for a brief moment at the beginning the node will consider himself in sync?). Then an extra check to make sure no block is being applied at the moment.

node/src/transaction_pool/candidate/transaction_pool_candidate_actions.rs
```diff
@@ -113,8 +113,14 @@ impl redux::EnablingCondition<crate::State> for TransactionPoolCandidateAction {
...
-                // TODO: if a block is being applied or produced, skip this action too
-                state.transition_frontier.sync.is_synced()
+                // Don't continue if we are producing a block, or we never synced yet
+                // or if the ledger service is busy.
+                !state.block_producer.is_producing()
+                    && state
+                        .transition_frontier
+                        .best_tip()
+                        .is_some_and(|b| !b.is_genesis())
+                    && !state.ledger.write.is_busy()
```
tizoc, line 123: @binier what do you think of this version? maybe I should add a similar check for reads? (I remember there was some prioritization going on but I don't remember the details, not even sure if we have any heavy reads that would delay anything)

binier: > Sounds good except is_synced is not really necessary (for a brief moment at the beginning the node will consider himself in sync?)

At the beginning node will not be synced. Then it will reconstruct genesis ledger/block, make it our best tip and go to synced state.

with above solution, it will not include any txs after genesis block (given that we are starting up a new network or doing hard fork). That might be a reasonable tradeoff though, so maybe it's fine.

Anyways, I think above solution is decent enough.

> maybe I should add a similar check for reads?

No, we could have reads all the time. E.g. peer requesting parts of the ledger.

binier (approved):

PR #1130, opened 2025-03-25
Added 'currentSnarkWorker' graphql endpoint

0xMimir:
Added 'currentSnarkWorker' graphql endpoint

### DIFF
--- a/node/native/src/graphql/mod.rs
+++ b/node/native/src/graphql/mod.rs
@@ -43,7 +43,8 @@ pub mod transaction;
 pub mod user_command;
 pub mod zkapp;
 
-pub type PublicKey = String;
+/// Base58 encoded public key
+pub type GraphQLPublicKey = String;
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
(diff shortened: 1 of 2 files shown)


node/native/src/graphql/mod.rs
```diff
@@ -35,6 +36,8 @@ pub mod transaction;
 pub mod user_command;
 pub mod zkapp;
 
+pub type PublicKey = String;
```
tizoc, line 39: `PublicKeyBase58` would be a better name since this needs to be decoded anyway to become a real public key.

0xMimir: Yes, but I wanted to keep same name as OCaml

node/native/src/graphql/mod.rs
```diff
@@ -433,6 +437,37 @@ impl Query {
...
+            .0
+            .oneshot_request(RpcRequest::SnarkerConfig)
+            .await
+            .ok_or(Error::StateMachineEmptyResponse)?;
+
+        let Some(config) = config else {
+            return Ok(None);
+        };
+
+        let account = context
```
adonagy, line 455: Please, use `context.load_account()`

adonagy (changes requested):
Sorry for late reply, I wanted to merge the account changes before any other grapqhl PRs. I expanded the context with a few methods. `load_account` will now take care of any account loading and per request caching. 

This adds a few merge conflicts I can help resolve if needed.

PR #1150, opened 2025-07-16
Add Makefile

dannywillems:
We want to have a centralized reference for all the commands the CI runs, so the user can replicate exactly the CI behavior on their machine.


Makefile
```diff
@@ -0,0 +1,57 @@
...
+help: ## Ask for help!
+	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+.PHONY: build
+build: ## Build the project in debug mode
+	cargo build
+
+.PHONY: build-release
+build-release: ## Build the project in release mode
+	cargo build --release --bin openmina
```
0xMimir, line 13: a small nitpick here, but `build` builds the entire project, while `build-release` builds only `openmina` executable

Makefile
```diff
@@ -0,0 +1,57 @@
...
+
+.PHONY: build-tests
+build-tests: ## Build test binaries
+	cargo build --release --tests --package=openmina-node-testing --package=cli
+
+.PHONY: build-wasm
+build-wasm: ## Build WebAssembly node
+	cd node/web && cargo +nightly build --release --target wasm32-unknown-unknown
```
0xMimir, line 21: This will build the `openmina-node-web` library but will not produce code needed to run the webnode for that following command should be used 

```sh
wasm-pack build --target web --out-dir pkg
```

0xMimir (approved):
LGTM, but check comments

Makefile
dannywillems: Good point, lemme fix this.

PR #1190, opened 2025-07-16
Docker: fix openmina dockerfile and slightly improve it

dannywillems:
Always build the docker images in CI, but do not publish it except on main, develop or a tag, as before.


0xMimir (approved):
LGTM

PR #1221, opened 2025-07-21
Add Makefile targets for block producer operations

### LINKED ISSUE #1198: Makefile: add a target to run a block producer
The target should to run on:
- devnet
- mainnet (in the future)

The documentation in `docs` should be updated accordingly. There is already a document called [Run Block Producing Node on Devnet] and this one should be modified accordingly.
The target should be named `run-block-producer` to follow the naming convention `run-archive` which has the pattern `run-[components]`.
The producer key should be handle with an environment variable.
All environment variables should be supported.

An example target to run the archive node is available [here](https://github.com/o1-labs/openmina/blob/develop/Makefile#L238) and the parameters used to run the block producer can be found [here](https://github.com/o1-labs/openmina/blob/develop/docker-compose.block-producer.yml). This should define the basic arguments the target should be using.

If using Claude, the file CLAUDE.md must be updated accordingly to include instructions for Claude.

dannywillems:
- Add run-block-producer target with devnet/mainnet support
- Add generate-block-producer-key target for key generation
- Include CI workflow to test documentation commands
- Update docs and CLAUDE.md with new target instructions

Resolves #1198

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,8 +40,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#1247](https://github.com/o1-labs/openmina/pull/1234)).
 - **Development tools**: Add shellcheck to the CI
   ([#1255](https://github.com/o1-labs/openmina/pull/1255))
-
-
+- **Makefile**: Add `run-block-producer` target with devnet/mainnet support and
+  `generate-block-producer-key` target for key generation
+  ([#1221](https://github.com/o1-labs/openmina/pull/1221)).
 
 ### Changed
 


.github/workflows/doc-commands.yml
```diff
@@ -0,0 +1,95 @@
...
+            exit 1
+          fi
+
+          echo "✅ Producer key file generated successfully with correct permissions"
+
+      - name: Test key file can be read by run-block-producer
+        run: |
+          echo "Testing that generated key can be used by run-block-producer target..."
+          # Run with --help to avoid actually starting the producer but verify key validation passes
+          timeout 30s make run-block-producer-devnet COINBASE_RECEIVER="test" || {
```
volhovm, line 76: Will this `make` call not trigger compilation process? If it does then 30 seconds will not be enough. Or do I misunderstand what this job is doing? 

On my machine the time to do the following is about 3 minutes:
```
її> make run-block-producer-devnet COINBASE_RECEIVER="test"
make run-block-producer NETWORK=devnet
make[1]: Entering directory '/home/volhovm/code/openmina'
cargo build --release --package=cli --bin openmina
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/volhovm/code/openmina/ledger/Cargo.toml
workspace: /home/volhovm/code/openmina/Cargo.toml
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/volhovm/code/openmina/tools/fuzzing/Cargo.toml
workspace: /home/volhovm/code/openmina/Cargo.toml
   Compiling mina-tree v0.16.0 (/home/volhovm/code/openmina/ledger)
   Compiling snark v0.16.0 (/home/volhovm/code/openmina/snark)
   Compiling vrf v0.16.0 (/home/volhovm/code/openmina/vrf)
   Compiling node v0.16.0 (/home/volhovm/code/openmina/node)
   Compiling openmina-node-common v0.16.0 (/home/volhovm/code/openmina/node/common)
   Compiling openmina-node-native v0.16.0 (/home/volhovm/code/openmina/node/native)
   Compiling cli v0.16.0 (/home/volhovm/code/openmina/cli)
    Finished `release` profile [optimized] target(s) in 2m 52s
MINA_PRIVKEY_PASS="" \
        cargo run --bin openmina \
        --release -- \
        node \
        --producer-key ./openmina-workdir/producer-key \
        --coinbase-receiver test \
         \
        --libp2p-port 8302 \
        --network devnet
```

volhovm (approved):
Approving, but please check my comment about 30 seconds timeout.  

I checked `generate-block-producer-key`, didn't run the other targets but they look reasonable from reading the code. I only have some reservations regarding CI role.

.github/workflows/doc-commands.yml
dannywillems: It's because the code is recompiled because of action_kind.rs being changed. Gotta fix that first. Good point.

dannywillems: FWIW: you can always execute a CI job locally by using [act](https://github.com/nektos/act). For instance, you can test locally the doc job using `act -j test-doc-commands`.

PR #1241, opened 2025-07-23
Add nix flake

volhovm:
This adds a nix flake that is necessary to compile the project on nixos.


volhovm:
@dannywillems I added it, rereview?

dannywillems:
> @dannywillems I added it, rereview?

Can you please add a test in the CI, as we do for macOS in #1247? Claude can be very helpful, I recommend you to use it to automate.
I would like us to be strict on new build systems added to the repository, and be sure we keep maintaining it. Adding a CI job testing it is the best way.

dannywillems:
You can rebase on top of `develop`. MacOS and Ubuntu setups have been fixed.

PR #1255, opened 2025-07-25
CI: add shellcheck

dannywillems:
Fixing the lint issues now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,6 +162,16 @@ format-md: ## Format all markdown and MDX files to wrap at 80 characters
 lint: ## Run linter (clippy)
 	cargo clippy --all-targets -- -D warnings --allow clippy::mutable_key_type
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -162,6 +162,16 @@ format-md: ## Format all markdown and MDX files to wrap at 80 characters
 lint: ## Run linter (clippy)
 	cargo clippy --all-targets -- -D warnings --allow clippy::mutable_key_type
 
+.PHONY: lint-bash
+lint-bash: ## Check all shell scripts using shellcheck
```
volhovm, line 166: This works for me. 

Should we also add `shellcheck` into `install-system-deps.sh`? I'd definitely add it to the `flake.nix` since it's something that developers might want to run locally.

volhovm (approved):

PR #1285, opened 2025-08-01
Bump up OCaml node to release 3.2.0-beta2

dannywillems:


amc-ie (approved):
Merci!

dannywillems:
Force merge as I only squashed commit and the previous build passes.

PR #1286, opened 2025-08-04
Fix test-update-ocaml-node-script

volhovm:
The `test-update-ocaml-node-script` runner times out on develop after 24h of doing nothing.
- https://github.com/o1-labs/openmina/actions/runs/16700395771/job/47270419437
- https://github.com/o1-labs/openmina/actions/runs/16689017012

This PR attempts to debug/fix this.


dannywillems:
The ubuntu-latest one works now. I will merge if this passes.

PR #1301, opened 2025-08-13
Website: remove outdated files

dannywillems:
They are related to an old k8s configuration or old network (berkeleynet).


0xMimir (approved):
LGTM

PR #1309, opened 2025-08-14
CI: add timeout-minutes at different places

dannywillems:
See commit messages.


.github/workflows/ci.yaml
```diff
@@ -94,6 +96,7 @@ jobs:
         run: make test-vrf
 
   p2p-tests:
+    timeout-minutes: 6
```
0xMimir, line 99: This seems very short, can we make it 10minutes

0xMimir (commented):
LGTM

PR #1335, opened 2025-08-22
Docker images: publish and update docs accordingly 

dannywillems:


dannywillems:
Merging for testing publishing.

PR #1367, opened 2025-09-02
Ledger + VRF: use public fields instead of methods from proof-systems

dannywillems:


dannywillems:
Closing as https://github.com/o1-labs/mina-rust/pull/1383

PR #1379, opened 2025-09-04
Vendor num-bigint into num-bigint-generics

dannywillems:
The library is from https://github.com/openmina/num-bigint/tree/1996d0807c31124315b93d009457716b61c6faab.
We want to get rid of forks, like `ark-ff`, `ark-ec`, `num-bigint`, etc in https://github.com/o1-labs/mina-rust/pull/1367. The fork of num-bigint includes a parametrized version of `BigInt`, which is not included in the original library. To avoid keeping multiple repositories over time, and have to go through different GitHub repositories, we suggest to vendor the library, and name it appropriately, coined `num-bigint-generic`.

### DIFF
--- a/vendor/num-bigint-generic/src/lib.rs
+++ b/vendor/num-bigint-generic/src/lib.rs
@@ -211,7 +211,7 @@ impl<T> fmt::Display for TryFromBigIntError<T> {
     }
 }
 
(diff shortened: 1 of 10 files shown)


dannywillems:
Closing as https://github.com/o1-labs/mina-rust/pull/1383

PR #1383, opened 2025-09-08
Remove forks of arkworks and use o1labs/proof-systems

dannywillems:
See also https://github.com/o1-labs/proof-systems/pull/3335


0xMimir (approved):
LGTM

PR #1432, opened 2025-09-19
CI: fix test-docs-infrastructure

dannywillems:


.github/workflows/test-docs-infrastructure.yaml
```diff
@@ -327,10 +329,10 @@ jobs:
               continue
             fi
 
-            echo "Testing script: $script_file"
+            echo "Testing script: $script_file with the endpoint $test_endpoint"
 
             # Execute the script with test endpoint and capture output
-            if output=$(timeout 30 bash "$script_file" "$test_endpoint" 2>&1); then
```
dannywillems, line 333: `timeout` does not exist on macos

PR #1447, opened 2025-09-23
CI: add a workflow for the project assignment

dannywillems:
We use GitHub projects, in particular the dashboard "Rust node" to track our work. We would like to add a check that an iteration and a project is added to each PR. It would help tracking our work per iteration.


dannywillems:
Closing. Old. Would never be revamped.

dannywillems:
We will check manually.

PR #1479, opened 2025-09-29
CI/build: reuse a cache when dependencies did not change

dannywillems:


dannywillems:
The CI changed a lot since then, and we use a cache. Closing.

PR #1493, opened 2025-10-02
Website: add documentation guidelines + code ref in doc

dannywillems:
In the future, we can add a tab with the OCaml code on one side, and the Rust code on the other side.

Here is an example of the output when using the plugin:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Website**: add "Join devnet" page with instructions for community members to
   join the devnet program and test the Rust node implementation
   ([#1425](https://github.com/o1-labs/mina-rust/issues/1425))
+- **Website**: add documentation guidelines
+  ([#1493](https://github.com/o1-labs/mina-rust/pull/1493))
 
 ### Changed
 


.github/scripts/verify-code-references.sh
```diff
@@ -0,0 +1,150 @@
...
+            # Trim whitespace from file_path
+            file_path=$(echo "$file_path" | xargs)
+
+            TOTAL_REFS=$((TOTAL_REFS + 1))
+
+            # Check if the source file exists
+            source_file="${REPO_ROOT}/${file_path}"
+            if [[ ! -f "$source_file" ]]; then
+                echo -e "${RED}✗${NC} Invalid reference in ${doc_file}:${line_num}"
+                echo "  File not found: ${file_path}"
```
yamimaio, line 48: Does this case cover the file existing but not having permissions?

yamimaio: And by the way, this works only for the related repo?  As in, our doc website will. check only references from the rust node repo or can it check any arbitrary repository as long as you configure it?

website/docs/developers/documentation-guidelines.md
```diff
@@ -0,0 +1,124 @@
...
+sidebar_position: 10
+title: Documentation Guidelines
+description: Best practices for writing and maintaining documentation
+slug: /developers/documentation-guidelines
+---
+
+# Documentation Guidelines
+
+This guide explains how to write and maintain documentation for the Mina Rust
+project, including how to reference code from the codebase. Referencing code
```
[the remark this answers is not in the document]
dannywillems: The line breaks at 80 characters, and it is fine as in markdown, it adds a new line only when you have two new lines.
```
`\n` -> no new line when rendered
`\n\n` -> a new line when rendered
```

website/docs/developers/documentation-guidelines.md
```diff
@@ -0,0 +1,124 @@
...
+- ✓ GitHub URLs match CODE_REFERENCE comments
+- ✓ Code blocks have corresponding references
+- ✓ Local code matches what's deployed on GitHub
+
+### When code changes
+
+If code is added or removed and line numbers shift:
+
+1. The verification script will detect the mismatch in CI
+2. Update the `CODE_REFERENCE` comment with new line numbers
```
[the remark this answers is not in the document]
dannywillems: It does not happen automatically. It is manual.

.github/scripts/verify-code-references.sh
dannywillems: > Does this case cover the file existing but not having permissions?

`-f filename` checks if the file exists. We should not care about the permissions as it would be on code that is supposed to have at least the `read` permissions (code, toml, yaml, etc).

PR #1562, opened 2025-10-20
Run unit tests in CI as a separate job

richardpringle:
This PR starts running unit-tests, just for `mina-tree` for now.

We can start commenting out the "exclude" flags one by one as we enable more and more tests.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -271,16 +271,11 @@ jobs:
       - name: Git checkout
         uses: actions/checkout@v5
 
+      - name: Load versions
+        uses: ./.github/actions/load-versions
+
       - name: Setup build dependencies
         uses: ./.github/actions/setup-build-deps
-
-      - name: Install libpcap
-        run: sudo apt-get update && sudo apt-get install -y libpcap-dev
-
-      - name: Use shared OCaml setting up steps
-        uses: ./.github/actions/setup-ocaml
-        with:
-          ocaml_version: ${{ env.OCAML_VERSION }}
 
       - name: Setup Rust
         uses: ./.github/actions/setup-rust
@@ -291,6 +286,7 @@ jobs:
       - name: Install cargo-nextest
         uses: taiki-e/install-action@nextest
 
+      # TODO: add this to the makefile
       - name: Run unit tests
         run: |
           cargo nextest run --workspace --lib \


.github/workflows/frontend-ci.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: ${{ github.event.inputs.os && fromJSON(format('["{0}"]', github.event.inputs.os)) || fromJSON('["ubuntu-22.04", "ubuntu-24.04-arm", "macos-13", "macos-14", "macos-15", "macos-latest"]') }}
```
richardpringle: Same comment as above

.github/workflows/build-ci.yaml
```diff
@@ -1,32 +1,19 @@
-name: Reusable Build Workflow
+name: Build CI
```
dannywillems: The idea of having separated files is to have different badges to add in the README: https://github.com/o1-labs/mina-rust?tab=readme-ov-file#supported-platforms. We can therefore see which platform is supported and the ones that are not.

richardpringle: Actually, we can just use shields.io, it's free
https://img.shields.io/github/check-runs/o1-labs/mina-rust/rp/test-everything?nameFilter=deploy


I haven't quite figured out the naming (`deploy` works, but I couldn't figure out the others), but I really don't think having a bunch of separate workflows is worth it just for the pretty badges

.github/workflows/tests.yaml
```diff
@@ -250,6 +250,65 @@ jobs:
...
+        with:
+          toolchain: ${{ env.RUST_STABLE_VERSION }}
+          cache-prefix: unit-tests-${{ env.CACHE_VERSION }}
+
+      - name: Install cargo-nextest
+        uses: taiki-e/install-action@nextest
+
+      - name: Run unit tests
+        run: |
+          cargo nextest run --workspace --lib \
```
dannywillems: Could we have a target in the Makefile for this instead, please? We try to have all the commands run by the CI in the Makefile.

.github/workflows/tests.yaml
```diff
@@ -265,6 +265,61 @@ jobs:
...
+        uses: ./.github/actions/setup-rust
+        with:
+          toolchain: ${{ env.RUST_STABLE_VERSION }}
+          cache-prefix: unit-tests-${{ env.CACHE_VERSION }}
+
+      - name: Install cargo-nextest
+        uses: taiki-e/install-action@nextest
+
+      # TODO: add this to the makefile
+      - name: Run unit tests
```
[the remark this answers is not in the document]
dannywillems: Note that we run already the tests of this crate. It is called `ledger-tests`. I would accept for now, as probably you'll do in a follow-up an improvement of it.

PR #1563, opened 2025-10-21
Fixed typo in `p2p_request_transactions_if_needed`

0xMimir:


dannywillems:
@0xMimir can you add a CHANGELOG entry please?

dannywillems:
Even for this kind of small patch, it would be nice to have unit/integration tests. We need to increase our code coverage. I accepted and merged, but for future PRs, we would enforce having additional tests to test the behaviors.

PR #1580, opened 2025-10-30
Move salsa-simple from tools/ to vendor/ and alphabetize workspace

dannywillems:
salsa-simple is not a tool but a vendored reimplementation of XSalsa20 with serde support, which the external salsa20 crate lacks. Moving it to vendor/ better reflects its purpose as a vendored dependency.

Also alphabetized workspace members in Cargo.toml for better organization and removed duplicate ledger entry.

### DIFF
--- a/frontend/angular.json
+++ b/frontend/angular.json
@@ -72,7 +72,15 @@
                   "with": "src/environments/environment.prod.ts"
                 }
               ],
-              "outputHashing": "all"
+              "outputHashing": "all",
+              "optimization": {
+                "scripts": true,
+                "styles": {
+                  "minify": true,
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,41 +1,36 @@
...
+  "node/web",
+  "p2p",
+  "p2p/libp2p-rpc-behaviour",
+  "p2p/testing",
   "poseidon",
-
-  "tools/transport",
+  "producer-dashboard",
+  "snark",
+  "tools/archive-breadcrumb-compare",
```
richardpringle: FYI, taplo has a setting to do this, I'm pretty sure

PR #1585, opened 2025-10-31
Website: add documentation about fee transfers

dannywillems:


dannywillems:
Closing as we will move the documentation in the code.

PR #1595, opened 2025-11-04
CI/docker: fix the image selection when testing env

dannywillems:
See failing test https://github.com/o1-labs/mina-rust/actions/runs/19072666396/job/54481606584


directcuteo (approved):

dannywillems:
Force merge as it will only impact develop.

PR #1620, opened 2025-11-10
Replayer node - add documentation

dannywillems:


0xMimir (dismissed):
Small comment for optional improvement, other than that LGTM

Issue #1235, opened 2025-07-22
Remove fork of ocaml-interop from sebastiencs


dannywillems:
We use the branch closure-values from https://github.com/sebastiencs/ocaml-interop.git. We should use the upstream version.

PR #1646, opened 2025-11-14
Remove OCaml related code from mina-tree

dannywillems:
Fix https://github.com/o1-labs/mina-rust/issues/1235

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Dependency**: use tag instead of references of o1-labs/proof-systems, fix
   [[#1674](https://github.com/o1-labs/mina-rust/issues/1674)]
   ([#1673](https://github.com/o1-labs/mina-rust/pull/1673))
+- Remove ocaml-interop dependency, fix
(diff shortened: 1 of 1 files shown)


ledger/src/account/legacy.rs
```diff
@@ -1,13 +1,10 @@
-use std::borrow::Cow;
-
+use super::common::*;
```
richardpringle, line 1: Gross... we need a lint preventing `*` imports.

ledger/src/lib.rs
```diff
@@ -129,9 +129,6 @@ mod wasm {
 #[macro_use]
 mod cache;
 
-#[cfg(all(not(target_family = "wasm"), feature = "ocaml-interop"))]
-mod ffi;
```
richardpringle, line 133: 🙌

ledger/src/proofs/mod.rs
richardpringle, line 1: I think we like code to look the same way 🏆

richardpringle (approved):
😍

ledger/src/account/legacy.rs
dannywillems: Yes!!

PR #1663, opened 2025-11-25
Deps: install taplo using nightly

dannywillems:
We're getting:
```
error: failed to compile `taplo-cli v0.9.3`, intermediate artifacts can be found at `/tmp/cargo-installcKWfPA`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  failed to parse manifest at `/home/fok/.cargo/registry/src/index.crates.io-6f17d22bba15001f/globset-0.4.18/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.84.1 (66221abde 2024-11-19)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
```
otherwise.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -296,7 +296,7 @@ setup-taplo: ## Install taplo TOML formatter
 	@if taplo --version 2>/dev/null | grep -q ${TAPLO_CLI_VERSION}; then \
 		echo "taplo ${TAPLO_CLI_VERSION} already installed"; \
 	else \
-		cargo install taplo-cli --version ${TAPLO_CLI_VERSION} --force; \
+		cargo +nightly install taplo-cli --version ${TAPLO_CLI_VERSION} --force; \
 	fi
 
 .PHONY: setup


Makefile
```diff
@@ -296,7 +296,7 @@ setup-taplo: ## Install taplo TOML formatter
 	@if taplo --version 2>/dev/null | grep -q ${TAPLO_CLI_VERSION}; then \
 		echo "taplo ${TAPLO_CLI_VERSION} already installed"; \
 	else \
-		cargo install taplo-cli --version ${TAPLO_CLI_VERSION} --force; \
+		cargo +nightly install taplo-cli --version ${TAPLO_CLI_VERSION} --force; \
```
richardpringle, line 299: I think this is fine given the pattern that we have. But we should really give people the option to get dependencies in whatever way works best for them. 

For instance, if I'm using `cargo install`, I'm going to do `cargo +stable install` because I usually have the lastest cargo version no matter the version that this workspace depends on. 

However, I probably wouldn't use `cargo install`, but instead use `cargo-binstall` which is a lot faster than compiling from source.

richardpringle (approved):

Makefile
richardpringle: So what I would love to see us do in the future is check for dependencies in the Makefile and print an error if the dependency isn't there.

PR #1671, opened 2025-11-26
Add the remaining commit 

richardpringle:
I think the top commit here has already been added
- **Add new version and rearrange steps for the release documentation**
- **Prepare release v0.18.1**


richardpringle:
The changelog is messed up, but we can fix that afterwards

PR #1673, opened 2025-11-26
Deps: use tag for proof-systems instead of rev

### LINKED ISSUE #1674: Update mina-rust to use the tag 0.2.0 of proof-systems

dannywillems:
We start enforcing the Caml and Rust node to use the same reference, and in particular we want to enforce releases only

Fix https://github.com/o1-labs/mina-rust/issues/1674

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1688,7 +1688,7 @@ checksum = "98cc8fbded0c607b7ba9dd60cd98df59af97e84d24e49c8557331cfc26d301ce"
 
 [[package]]
 name = "cli"
-version = "0.18.0"
+version = "0.18.1"
 dependencies = [
  "anyhow",
  "backtrace",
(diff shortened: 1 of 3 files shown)


richardpringle (approved):

Cargo.lock
```diff
@@ -3282,7 +3282,7 @@ dependencies = [
 [[package]]
 name = "groupmap"
 version = "0.1.0"
-source = "git+https://github.com/o1-labs/proof-systems?rev=282faf5#282faf58a97a0397304ca02010de7e43d3d3ec20"
+source = "git+https://github.com/o1-labs/proof-systems?tag=0.2.0#e8e02799802b42caf727d69c2dba7fce9548cb5b"
```
richardpringle, line 3285: @dannywillems, you're aware that this is a commit change too, right?

PR #1684, opened 2025-12-01
CI: VRF unit tests run as part of unit-tests

dannywillems:
There are no integrations tests (following the Rust naming convention described [here](https://doc.rust-lang.org/book/ch11-03-test-organization.html)), only unit tests.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -150,36 +150,6 @@ jobs:
   #     - name: Run ledger tests
   #       run: make test-ledger
 
-  vrf-tests:
-    timeout-minutes: 20
-    runs-on: ubuntu-24.04
-    steps:
-      - name: Git checkout
-        uses: actions/checkout@v6
-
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -302,7 +272,6 @@ jobs:
               --exclude libp2p-rpc-behaviour \
               --exclude node \
               --exclude mina-node-account \
-              --exclude vrf \
```
richardpringle, line 305: Can you move this down to a comment at the bottom? Just helps us keep track of what's explicitly being run. Once we stop excluding most tests, we can delete the comment

richardpringle (changes requested):

.github/workflows/tests.yaml
dannywillems: Done, and pushed forced.

richardpringle (approved):

PR #1685, opened 2025-12-01
Website: update release branch name for hotfix

### LINKED ISSUE #1686: Website: update appendix to request to create a tag for minor releases to include bugfixes
- When we push a first version x.y, we must create the tag x.y and also x.y.0
- When a hotfix is created, the branch x.y must be updated by making a PR from x.y.z to x.y.z.

dannywillems:


website/docs/appendix/release-process.mdx
```diff
@@ -427,7 +452,7 @@ For urgent patches to released versions:
 
    ```bash
    git checkout v1.5.0
-   git checkout -b hotfix/v1.5.1
+   git checkout -b release/v1.5.1
```
richardpringle, line 455: At the end of the day, I wouldn't want to be too pedantic about this, but the way I would do it given our current branching pattern is to `git fetch && git checkout -b hotfix/v1.5.1` then `git reset --hard origin/releases/v1.5`. In fact, I would say that the branch for the hotfix really doesn't matter and can be named whatever. The important part is that the new commits are merged into `release/v1.5`, then the release process is performed directly on the branch. 

This would mean that there's a brief moment where the tip isn't an actual tagged version, but that should only last as long as the actual release process takes after the hotfix is merged.

website/docs/appendix/release-process.mdx
```diff
@@ -440,12 +465,49 @@ For urgent patches to released versions:
    - Verify Docker builds
    - Create GitHub release
 
+4. **Update the minor release branch**
+
+   After the hotfix is released, update the minor release branch
+   (`release/vX.Y`) to include the hotfix changes. This ensures the branch
+   always points to the latest patch version.
```
richardpringle, line 472: So with my comment above, I don't see the point of releasing then updating the branch instead of just releasing directly on the branch. The important thing to question here is, "What is the purpose of the release branch and how if at all can it be misused?". 

My answer to this question is that the release branch's sole purpose is for creating releases. If the version is `v1.5.z`, the branch should contain every patch from 0 onward. If you want to know what the latest version of the code is, you should be looking at the releases page, not the branch.

website/docs/appendix/release-process.mdx
```diff
@@ -440,12 +465,49 @@ For urgent patches to released versions:
...
+   the latest `v1.5.1` release.
+
+<!-- prettier-ignore-start -->
+
+:::important Keep minor release branches current
+
+The minor release branch (`release/vX.Y`) must always point to the latest patch
+release in that series. After every hotfix release (`vX.Y.Z` where `Z > 0`),
+create a PR to merge the hotfix branch into the corresponding `release/vX.Y`
+branch.
```
richardpringle, line 491: I'm repeating myself here, but I just want to make sure the point is clear. The releases should happen on the branch and nowhere else. If that's the case, we don't have to enforce the "must always point to the latest patch
release," it just will be the case.

website/docs/appendix/release-process.mdx
```diff
@@ -440,12 +465,49 @@ For urgent patches to released versions:
...
+
+For example:
+
+- After releasing `v1.5.1`, merge `release/v1.5.1` into `release/v1.5`
+- After releasing `v1.5.2`, merge `release/v1.5.2` into `release/v1.5`
```
richardpringle, line 496: You'll have to explain the purpose of `release/v1.5.1` (and 1.5.2) as branches. I don't see the point of having a specific branch naming strategy for a single released version, a tag does everything you need for a particular version. No branches (especially protected branches) serve a purpose if their progress should ever cease.

richardpringle (commented):
I left some comments... I think this is fine as is, but there's a bunch of process here where I don't see any immediate benefits. Branches are a mechanism for tracking something that's dynamic and tags are a way to track something that's static. We seem to be tracking a static thing with a branch then redundantly using tags. 

I'm not suggesting we're doing anything wrong or that there aren't benefits to the way we're doing things, the benefits just aren't immediately clear to me and it seems there's extra process where I can't see a benefit.

richardpringle (approved):

richardpringle:
I meant to leave my "comment" review as an approval

dannywillems:
Forcing merged, even if the test has been cancelled. We should probably increase the time later if important.

dannywillems:
The changes are not related to the test, so the fact it fails only gives an information, but unrelated to this patch.

dannywillems:
@richardpringle I'm addressing all your concerns in a new PR. I agree with all of them.

PR #1691, opened 2025-12-01
Makefile: fix fix-trailing-whitespace on MacOS

dannywillems:
It should be compatible now with BSD sed


richardpringle (approved):

dannywillems:
Force merging as it is useless to wait for the CI jobs running e2e and integration tests.

PR #1705, opened 2025-12-03
Node dashboard: remove old nodes and use o1Labs infrastructure

### LINKED ISSUE #1695: Node dashboard: remove old nodes and BP from configuration and use o1Labs
There are still some configurations set-up with old BP and old nodes, like the one used when we run `make start-dev-mobile`.
We are now running at o1Labs some nodes, documented [here](https://o1-labs.github.io/mina-rust/docs/node-operators/infrastructure/seed-nodes/), [here](https://o1-labs.github.io/mina-rust/docs/node-operators/infrastructure/plain-nodes/) and [here](https://o1-labs.github.io/mina-rust/docs/node-operators/infrastructure/block-producers/). We should use them instead.

We should also deploy a new version of the frontend with these new nodes.

directcuteo:
## Summary

- Cleaned up development configuration by removing old devnet nodes
- Updated configuration to use o1Labs infrastructure (Plain, BP, and Seed nodes)
- Updated GraphQL endpoint to use o1labs devnet o1test.net

### DIFF
--- a/frontend/src/environments/environment.ts
+++ b/frontend/src/environments/environment.ts
@@ -1,3 +1,7 @@
+/**
+ * Development environment configuration
+ */
+
 import { MinaEnv } from '@shared/types/core/environment/mina-env.type';
 
 export const environment: Readonly<MinaEnv> = {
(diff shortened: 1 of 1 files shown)


frontend/src/environments/environment.ts
```diff
@@ -28,138 +32,44 @@ export const environment: Readonly<MinaEnv> = {
...
     {
-      name: 'staging-devnet-bp-1',
-      url: 'https://staging-devnet-openmina-bp-1.minaprotocol.network',
+      name: 'o1Labs Seed Node 2',
+      url: 'https://mina-rust-seed-2.gcp.o1test.net/',
     },
     {
-      name: 'staging-devnet-bp-2',
-      url: 'https://staging-devnet-openmina-bp-2.minaprotocol.network',
+      name: 'o1Labs Seed Node 3',
```
dannywillems, line 67: We should also add them in the production environment as it is the one we used to deploy in prod.

dannywillems: Follow-up https://github.com/o1-labs/mina-rust/issues/1735

dannywillems (approved):
Merging to unblock, but I left a follow-up request.
Note that the command to run to get that default environment is `make start-dev-mobile`.

PR #1875, opened 2025-12-12
CI: add beta channel lint workflow

### LINKED ISSUE #1777: CI: add a check that the code can be compiled with the latest beta channel
The job could fail, and most of the time it is mostly the job regarding`clippy` that fails and we should simply fix it.

### LINKED ISSUE #1877: Clippy: investigate usage of mutable_key_type
image

Seems it's just in `ledger/tx_pool`, either way, I think a separate issue to either fix the root cause or add the add allow in `ledger` crate is the way to go.

_Originally posted by @iostat in https://github.com/o1-labs/mina-rust/pull/1875#discussion_r2614813705_

iostat:
### Description

Add a new CI workflow to check compilation and linting with the Rust beta
channel. This helps catch build issues early before they appear in stable releases.

The make targets explicitly use `cargo +beta` to override `rust-toolchain.toml`,
ensuring the beta toolchain is used for checking and linting.

### Related Issue(s)

Closes #1777

### Type of Change

- [x] Build-related changes

### Testing

The workflow will run automatically on this PR to verify it works correctly.

- [x] Added make targets `check-beta` and `lint-beta`
- [x] Workflow configured with `continue-on-error: true`
- [x] Workflow runs successfully in CI (will verify on this PR)
See... the workflow failed, but that's actually a success! :)
image

### Changelog Entry

**Changelog Summary:** CI: add beta channel lint workflow to catch
build issues early

### DIFF
--- a/.github/workflows/lint-beta.yaml
+++ b/.github/workflows/lint-beta.yaml
@@ -1,0 +1,47 @@
+name: Lint (Beta Channel)
+
+on:
+  push:
+    branches: [ main, develop ]
+  pull_request:
+  schedule:
+    # Run weekly on Mondays at 00:00 UTC
+    - cron: '0 0 * * 1'
+
+jobs:
+  lint-beta:
+    timeout-minutes: 15
+    name: Lint (Beta) - ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    # Allow this job to fail without failing the entire workflow
+    continue-on-error: true
+    strategy:
+      matrix:
+        os: [ubuntu-24.04]
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Load versions
+        uses: ./.github/actions/load-versions
+
+      - name: Setup build dependencies
+        uses: ./.github/actions/setup-build-deps
+
+      - name: Use shared OCaml setting up steps
+        uses: ./.github/actions/setup-ocaml
+        with:
+          ocaml_version: ${{ env.OCAML_VERSION }}
+
+      - name: Setup Rust (Beta)
+        uses: ./.github/actions/setup-rust
+        with:
+          toolchain: ${{ env.RUST_BETA_VERSION }}
+          components: clippy, rustfmt
+          cache-prefix: lint-beta-${{ env.RUST_BETA_VERSION }}-v0
+
+      - name: Run make check-beta
+        run: make check-beta
+
+      - name: Run make lint-beta
+        run: make lint-beta
+
--- a/Makefile
+++ b/Makefile
@@ -151,6 +151,10 @@ bench-database: ## Run ledger database benchmark
 check: ## Check code for compilation errors
 	cargo check --all-targets
 
+.PHONY: check-beta
+check-beta: ## Check code for compilation errors using beta Rust
+	cargo +beta check --all-targets
+
 .PHONY: check-tx-fuzzing
 check-tx-fuzzing: ## Check the transaction fuzzing tools, requires nightly Rust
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -244,6 +248,10 @@ format-md: ## Format all markdown and MDX files to wrap at 80 characters
 lint: ## Run linter (clippy)
 	cargo clippy --all-targets -- -D warnings --allow clippy::mutable_key_type
 
+.PHONY: lint-beta
+lint-beta: ## Run linter (clippy) using beta Rust
+	cargo +beta clippy --all-targets -- -D warnings --allow clippy::mutable_key_type
```
richardpringle, line 253: Wait... Why do we need this allow? This should be an allow only in the modules where needed.

(This shouldn't hold up the PR though, it can be fixed after)

.github/workflows/lint-beta.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Setup Rust (Beta)
+        uses: ./.github/actions/setup-rust
+        with:
+          toolchain: ${{ env.RUST_BETA_VERSION }}
+          components: clippy, rustfmt
+          cache-prefix: lint-beta-${{ env.RUST_BETA_VERSION }}-v0
+
+      - name: Run make check-beta
+        run: make check-beta
```
richardpringle, line 43: Can you leave a comment here (I think this is the better place rather than the makefile) that we aren't checking all feature flags? 

I don't have an immediate solution for this as we really want to perform checks with and without feature flags. I think there's a tool for this that checks every permutation of feature flags, but I think that's overkill for this job right now.

richardpringle (approved):
Leaving a couple comments but since this isn't meant to block PRs, I think it's best to merge this one and do a follow up PR to address my comments. 

However, @iostat if you want to address those comments in this PR so you don't have to open up another one, I think that's fine too, your call!

Makefile
iostat: This is exactly the `lint` target right above it with `+beta` added. But yeah, that's probably worth a separate issue to clean up :)

dannywillems: > Wait... Why do we need this allow? This should be an allow only in the modules where needed.

Agreed with @richardpringle

iostat: image

Seems it's just in `ledger/tx_pool`, either way, I think a separate issue to either fix the root cause or add the allow in `ledger` crate is the way to go.

dannywillems: Opened https://github.com/o1-labs/mina-rust/issues/1877

.github/workflows/lint-beta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Load versions
+        uses: ./.github/actions/load-versions
+
+      - name: Setup build dependencies
+        uses: ./.github/actions/setup-build-deps
+
+      - name: Use shared OCaml setting up steps
```
dannywillems, line 30: This should not be useful anymore.

.github/workflows/lint-beta.yaml
iostat: Created https://github.com/o1-labs/mina-rust/issues/1878 so we can spec this out better and address it sooner than later

richardpringle (approved):

dannywillems:
Merging after https://github.com/o1-labs/mina-rust/pull/1872

dannywillems:
Force merging, and @iostat will fix it soon by upgrading to Rust 1.92

PR #1923, opened 2025-12-16
Frontend: update package-lock.json

dannywillems:
Running `make install-deps` gives me this `package-lock.json`.


dannywillems:
Added in https://github.com/o1-labs/mina-rust/pull/2096

PR #1957, opened 2025-12-18
Ensure transaction hash is parsed as a string

### LINKED ISSUE #1331: Text overlapping in mempool table
Check this:

Image
It may be a bug in the UI, or something changed in the data. We should look first where does this change come from.

directcuteo:
### Description

Fixing bug on text overlapping in mempool table.

### Related Issue(s)

https://github.com/orgs/o1-labs/projects/24/views/3?sliceBy%5Bvalue%5D=directcuteo&pane=issue&itemId=125294112&issue=o1-labs%7Cmina-rust%7C1331

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

### Changelog Entry

- **Fixed** a bug where text was overlapping in the view.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Frontend**: Fixed a bug where text was overlapping in the view.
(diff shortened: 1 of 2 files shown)


yamimaio (approved):
Approving but just wondering if there is any testing anywhere to prevent failures due to wrong inputs.

I'd expect a test whenever there's a bug fix

Issue #1893, opened 2025-12-12
Clippy: Address unused_assignments warnings


iostat:
## Summary

After upgrading to Rust 1.92 (PR #1894), clippy reports approximately **700 instances** of unused assignments, primarily in Redux-style action enums where fields are destructured but individual values aren't used. We've actually always had a large amount of these unused assigns (it might even be the same number) but it came to my attention during the upgrade work and it's worth addressing this sooner than later.

## Context

These warnings are typically in Redux action patterns where action fields are destructured but not all fields are used in the reducer. The fix is to prefix unused fields with `_` (e.g., `_rpc_id`).

## Priority

These warnings are **not blocking** - they don't prevent compilation and don't indicate actual bugs. This can be addressed in a follow-up PR after the main 1.92 upgrade is merged.

## Related

Issue #1560 
PR #1894

iostat:
So looking at this deeper, there are some really strange false positives going on here. When I first looked at this a few days ago, I saw places where some fields were genuinely unused, and assumed they were being kept around for logging or serialization compatibility etc. There were other instances where some fields were just cloned and passed on to another action (which makes sense that the compiler might see it as dead code, and the ownership should move somewhere else -- but this doesn't fit neatly with the Redux-style model)

[...]

PR #1968, opened 2025-12-18
Fix more 1.92 Clippies

### LINKED ISSUE #1895: Clippy: Address large struct/enum variants
As part of PR #1894, we added `--allow` for `clippy::large_enum_variant` and `clippy::result-large-err` (`--allow unused` is #1893). We want to remove this lint exception, but some existing large structs might be architecturally necessary. This is a follow-up issue to investigate what we can `Box` and what we need to `#[allow]`/`#[expect]`

iostat:
### Description

Addresses the `clippy::{large_enum_variants,result_large_err}` lints introduced by the 1.92 upgrade. Some enum variants were boxed where it made sense, others we're allowed as the size difference between the largest/second-largest wasn't too big to warrant a larger refactor. 

Removes the `--allow clippy::{large_enum_variants,result_large_err}` from `make lint[-beta]`

Additionally, the `--allow unused` was changed to `--allow unused_assignments`, which is the more specific lint that was triggering #1893. Seeing as that's likely a compiler bug, we're keeping that issue open. However, `--allow unused` did allow some `dead_code` to slip through the cracks, so that's been addressed here too.

### Related Issue(s)

Closes #1895 

### Type of Change

- [x] Performance improvement (potentially! some multi-kb structures were moved from stack to heap)

### Testing

Code should compile and CI should pass. There are no functional changes in this PR.

### Changelog Entry

- **Fixed**: **CI**: Remove clippy exceptions for large enum/err variants added as part of the Rust 1.92 upgrade

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -262,16 +262,12 @@ format-md: ## Format all markdown and MDX files to wrap at 80 characters
 .PHONY: lint
 lint: ## Run linter (clippy)
 	cargo clippy --all-targets -- -D warnings \
-	  --allow clippy::large_enum_variant \
-	  --allow clippy::result-large-err \
-	  --allow unused
+	  --allow unused_assignments
 
 .PHONY: lint-beta
 lint-beta: ## Run linter (clippy) using beta Rust
 	cargo +beta clippy --all-targets -- -D warnings \
-	  --allow clippy::large_enum_variant \
-	  --allow clippy::result-large-err \
-	  --allow unused
+	  --allow unused_assignments
 
 .PHONY: lint-bash
 lint-bash: ## Check all shell scripts using shellcheck
--- a/crates/ledger/src/proofs/transaction.rs
+++ b/crates/ledger/src/proofs/transaction.rs
@@ -4397,6 +4397,7 @@ pub(super) mod tests {
 
     /// External worker input.
     #[derive(Debug, BinProtRead, BinProtWrite)]
+    #[expect(clippy::large_enum_variant, reason = "This enum is only used in tests")]
     pub enum ExternalSnarkWorkerRequest {
         /// Queries worker for readiness, expected reply is `true`.
         AwaitReadiness,
--- a/crates/ledger/src/scan_state/scan_state.rs
+++ b/crates/ledger/src/scan_state/scan_state.rs
@@ -2142,7 +2142,10 @@ impl ScanState {
                     }
                 };
 
-                Ok(snark_work::spec::Work::Transition((statement, witness)))
+                Ok(snark_work::spec::Work::Transition((
+                    statement,
+                    Box::new(witness),
+                )))
             }
             Extracted::Second(s) => {
                 let merged = s.0.statement().merge(&s.1.statement())?;
@@ -2196,7 +2199,10 @@ impl ScanState {
                     }
                 };
 
-                Some(snark_work::spec::Work::Transition((statement, witness)))
+                Some(snark_work::spec::Work::Transition((
+                    statement,
+                    Box::new(witness),
+                )))
             }
             Extracted::Second(s) => {
                 let merged = s.0.statement().merge(&s.1.statement()).unwrap();
--- a/crates/ledger/src/scan_state/snark_work.rs
+++ b/crates/ledger/src/scan_state/snark_work.rs
@@ -5,7 +5,7 @@ pub mod spec {
     };
 
     pub enum Work {
-        Transition((Box<Statement<()>>, TransactionWitness)),
+        Transition((Box<Statement<()>>, Box<TransactionWitness>)),
         Merge(Box<(Statement<()>, Box<(LedgerProof, LedgerProof)>)>),
     }
 }
--- a/crates/ledger/src/scan_state/snark_work.rs
+++ b/crates/ledger/src/scan_state/snark_work.rs
@@ -30,6 +30,7 @@ mod tests {
 
     /// External worker input.
     #[derive(Debug, BinProtRead, BinProtWrite)]
+    #[expect(clippy::large_enum_variant, reason = "This enum is only used in tests")]
     pub enum ExternalSnarkWorkerRequest {
         /// Queries worker for readiness, expected reply is `true`.
         AwaitReadiness,
--- a/crates/node/src/event_source/event_source_actions.rs
+++ b/crates/node/src/event_source/event_source_actions.rs
@@ -3,6 +3,7 @@ use serde::{Deserialize, Serialize};
 pub type EventSourceActionWithMeta = redux::ActionWithMeta<EventSourceAction>;
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
+#[expect(clippy::large_enum_variant, reason = "This enum wraps super::Event")]
 pub enum EventSourceAction {
     /// Notify state machine that the new events might be received/available,
     /// so trigger processing of those events.
--- a/crates/node/src/event_source/event_source_effects.rs
+++ b/crates/node/src/event_source/event_source_effects.rs
@@ -489,7 +489,9 @@ pub fn event_source_effects<S: Service>(store: &mut Store<S>, action: EventSourc
             Event::GenesisLoad(res) => match res {
                 Err(err) => todo!("error while trying to load genesis config/ledger. - {err}"),
                 Ok(data) => {
-                    store.dispatch(TransitionFrontierGenesisAction::LedgerLoadSuccess { data });
+                    store.dispatch(TransitionFrontierGenesisAction::LedgerLoadSuccess {
+                        data: Box::new(data),
+                    });
                 }
             },
         },
--- a/crates/node/src/event_source/event_source_effects.rs
+++ b/crates/node/src/event_source/event_source_effects.rs
@@ -261,10 +261,7 @@ pub fn event_source_effects<S: Service>(store: &mut Store<S>, action: EventSourc
                             store.dispatch(P2pDisconnectionAction::Init { peer_id, reason });
                         }
                         Ok(message) => {
-                            store.dispatch(P2pChannelsMessageReceivedAction {
-                                peer_id,
-                                message: Box::new(message),
-                            });
+                            store.dispatch(P2pChannelsMessageReceivedAction { peer_id, message });
                         }
                     },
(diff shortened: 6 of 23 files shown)


Makefile
```diff
@@ -262,16 +262,12 @@ format-md: ## Format all markdown and MDX files to wrap at 80 characters
 .PHONY: lint
 lint: ## Run linter (clippy)
 	cargo clippy --all-targets -- -D warnings \
-	  --allow clippy::large_enum_variant \
-	  --allow clippy::result-large-err \
-	  --allow unused
+	  --allow unused_assignments
```
richardpringle, line 265: Oh no... for the whole thing!?

richardpringle: (I actually don't know the difference between `unused_assignments` vs `unused`)

crates/ledger/src/proofs/transaction.rs
```diff
@@ -4397,6 +4397,7 @@ pub(super) mod tests {
 
     /// External worker input.
     #[derive(Debug, BinProtRead, BinProtWrite)]
+    #[expect(clippy::large_enum_variant, reason = "This enum is only used in tests")]
```
richardpringle, line 4400: We should put this behind a feature flag if it's only used in tests

richardpringle: Don't need to change in this PR, but can you create an issue before resolving the comment?

crates/ledger/src/scan_state/scan_state.rs
```diff
@@ -2142,7 +2142,10 @@ impl ScanState {
                     }
                 };
 
-                Ok(snark_work::spec::Work::Transition((statement, witness)))
+                Ok(snark_work::spec::Work::Transition((
+                    statement,
+                    Box::new(witness),
```
richardpringle, line 2147: Curious to hear your opinion on this, but when it comes to wrapper types like this, I prefer to use `witness.into()`. That way, changes are minimized if we decide to have any other wrapper or anything like that. 

(not a change for this PR, just interested in your thoughts)

crates/ledger/src/scan_state/snark_work.rs
```diff
@@ -30,6 +30,7 @@ mod tests {
 
     /// External worker input.
     #[derive(Debug, BinProtRead, BinProtWrite)]
+    #[expect(clippy::large_enum_variant, reason = "This enum is only used in tests")]
```
richardpringle, line 33: Same comment

crates/node/src/event_source/event_source_actions.rs
```diff
@@ -3,6 +3,7 @@ use serde::{Deserialize, Serialize};
 pub type EventSourceActionWithMeta = redux::ActionWithMeta<EventSourceAction>;
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
+#[expect(clippy::large_enum_variant, reason = "This enum extends super::Event")]
```
richardpringle: 🤔 I'm not sure if I agree with this one. This should almost certainly be boxed, no?

Also... a nit... but... I would say "wraps," not "extends"

crates/node/src/event_source/event_source_effects.rs
```diff
@@ -489,7 +486,9 @@ pub fn event_source_effects<S: Service>(store: &mut Store<S>, action: EventSourc
             Event::GenesisLoad(res) => match res {
                 Err(err) => todo!("error while trying to load genesis config/ledger. - {err}"),
                 Ok(data) => {
-                    store.dispatch(TransitionFrontierGenesisAction::LedgerLoadSuccess { data });
+                    store.dispatch(TransitionFrontierGenesisAction::LedgerLoadSuccess {
+                        data: Box::new(data),
```
richardpringle, line 490: Again, this is one of those places where I would use `Into::into`, just because it means this code doesn't change if the size of `data` changes

crates/node/src/ledger/ledger_manager.rs
```diff
@@ -22,38 +22,48 @@ use std::collections::BTreeMap;
 /// can't be expressed in the Rust type system at the moment. For this
 /// reason this type is private while functions wrapping the whole call
 /// to the service are exposed as the service's methods.
-#[allow(dead_code)] // TODO
```
richardpringle, line 25: 😳

crates/node/src/ledger/ledger_manager.rs
```diff
@@ -22,38 +22,48 @@ use std::collections::BTreeMap;
...
+#[expect(
```
richardpringle, line 25: `expect`?

crates/node/src/ledger/ledger_service.rs
```diff
@@ -1271,7 +1271,7 @@ impl LedgerCtx {
             .view()
             .map(|jobs| {
                 let jobs = jobs.collect::<Vec<JobValueWithIndex<'_>>>();
-                let mut iter = jobs.iter().peekable();
+                let iter = jobs.iter().peekable();
```
richardpringle, line 1274: 😎

crates/node/src/transition_frontier/genesis/transition_frontier_genesis_reducer.rs
```diff
@@ -45,7 +45,7 @@ impl TransitionFrontierGenesisState {
             TransitionFrontierGenesisAction::LedgerLoadSuccess { data } => {
                 *state = Self::LedgerLoadSuccess {
                     time: meta.time(),
-                    data: data.clone(),
+                    data: *data.clone(),
```
richardpringle, line 47: Really!? `Clone::clone` should always return an owned type, no?

richardpringle: Ah nvm... `data` is `&Box`

tools/testing/src/scenario/event_details.rs
```diff
@@ -8,14 +8,13 @@ use node::{
...
-        let rpc_state = &state.p2p.get_ready_peer(peer_id)?.channels.rpc;
-        if *req_id == rpc_state.pending_local_rpc_id()? {
-            return Some(format!("Request: {}", rpc_state.pending_local_rpc()?));
+    // this could be a let-chain but we're on rust 2021 instead of 2024 >:(
+    if let Event::P2p(P2pEvent::Channel(P2pChannelEvent::Received(peer_id, Ok(msg)))) = event {
+        if let ChannelMsg::Rpc(RpcChannelMsg::Response(req_id, _)) = &**msg {
+            let rpc_state = &state.p2p.get_ready_peer(peer_id)?.channels.rpc;
+            if *req_id == rpc_state.pending_local_rpc_id()? {
+                return Some(format!("Request: {}", rpc_state.pending_local_rpc()?));
+            }
```
richardpringle, line 17: This is gross... but it was gross before... I just couldn't look at it without commenting

crates/p2p/src/service_impl/webrtc/mod.rs
richardpringle, line 1: yikes

richardpringle (dismissed):

Makefile
iostat: `unused` is every kind of unused, including `_assignments`, `_variables`, `_muts`, `_imports`, and I believe `dead_code`. `unused_assignments` is just assignments

crates/ledger/src/proofs/transaction.rs
iostat: It's already inside `#[cfg(test)] pub(super) mod tests`...

crates/ledger/src/scan_state/scan_state.rs
iostat: I prefer to be explicit... Into can occasionally confuse the compiler when it's trying to resolve a lot of potentially applicable instances (happens when you love your generics). Moreover, being explicit allows you to see the surface area with the errors you're going to have to address when you change the container, forcing you to be more considerate of your change

crates/ledger/src/scan_state/snark_work.rs
iostat: Similarly, already within `#[cfg(test)] mod tests`.

crates/node/src/event_source/event_source_actions.rs
iostat: The large_enum_variant lint gets triggered not when your structure itself is large, but when there's a large size difference between the largest and second-largest variants of the enum, implying that in many cases, there's a lot of wasted space allocated to variants that don't need it. In this case, I'm considering how the data is actually being used. In many ways, this enum is like a "spicy `Option<Event>`" -- i.e., it has a reason for being `None` (one of the three variants besides `NewEvent`), but if there **is** `Some(Event)`, you probably want it already on the stack to be processed, and the heap indirection would be inefficient

crates/node/src/ledger/ledger_manager.rs
iostat: It's like `#[allow]`, but if the warning **isn't** triggered it throws a warning/error. So if you've addressed the issue that led to the `#[allow]` in the first place, you're forced to remove it. 

https://users.rust-lang.org/t/dont-allow-expect-and-be-reasonable/136078

iostat: In fact, that `#[allow(dead_code)]` I removed in the line right above is a prime example of why you should prefer `#[expect]` :)

crates/p2p/src/service_impl/webrtc/mod.rs
iostat: I'm guessing you mean the ton of changes below -- it's effectively an indentation change but looks dumb in the diff. I had to add a deref to the Box in External, but the matches before had several arms for different variants of the PeerCmd inside External.

tools/testing/src/scenario/event_details.rs
iostat: image

(agreed)

crates/node/src/event_source/event_source_actions.rs
iostat: Agree on "wraps" -> "extends"

Makefile
iostat: Ideally we take it out when the compiler issue in #1893 is addressed ;)

dannywillems (approved):
Bumping @richardpringle approval

PR #2052, opened 2026-01-09
Deprecate WebRTC-Multiaddrish signaling format in favor of standard Multiaddr

### LINKED ISSUE #2036: Unify WebRTC signaling address format with standard Multiaddr
This can be done independent of the epic, but melds well with the overall goal of polishing up the WebRTC experience.

We'd like to replace the custom Multiaddr-like format, lovingly dubbed WebRTC-Multiaddrish, with actual Multiaddr parsing.
The current `FromStr for P2pConnectionOutgoingInitOpts` in `crates/p2p/src/connection/outgoing/mod.rs` uses a custom format: (`/{peer_id}/{signal-protocol}/{host}/{port}`).

**Current behavior** (around lines 350-400):

Checks if the address starts with `/ip` or `/dns`, if yes, treates it as a Multiaddr for a LibP2P connection, otherwise parses it as signaling information

**Proposed behavior**:

- Parse standard multiaddr format
- Check protocol stack for `/webrtc/http`, `/webrtc/https`, `/webrtc/http(s)/http-path` (for `Http(s)Proxy`). These are [standard Multiaddr protocols](https://github.com/multiformats/multiaddr/blob/master/protocols.csv)!
- Extract peer ID and signaling info from Multiaddr components if it's a WebRTC signal address (return `Self::WebRTC`) or a normal `Self::LibP2p` peer otherwise

**Example transformation**:

```
# Current format
/12D3KooW.../https/signal.example.com/443

# Standard multiaddr format
/dns4/signal.example.com/tcp/443/webrtc-https/12D3KooW...
```

**Benefits**:
- Simplified parsing logic
- Clearer semantics
- Standard tooling compatibility
- Eliminates a ton of awkward documentation, and the peer address clearly identifies itself as WebRTC-capable

iostat:
### Description

Unifies the WebRTC signaling address format with standard multiaddr conventions. This introduces a "new" address format using the `/webrtc` protocol as a distinguisher from standard libp2p multiaddrs, while maintaining backward compatibility with the legacy format through deprecation warnings.

**New multiaddr format:**
- HTTP(S): `/<dns|dns4|dns6|ip4|ip6>/{host}/tcp/{port}/webrtc/<http|https>/p2p/{peer_id}`
- Proxied: `/<dns|dns4|dns6|ip4|ip6>/{host}/tcp/{port}/webrtc/<http|https>/http-path/{path}/p2p/{peer_id}`
- P2P Relay: `/p2p/{relay_peer_id}/webrtc/p2p-circuit/p2p/{target_peer_id}`

**Legacy format (deprecated):**
- `/{peer_id}/{http|https|https_proxy|p2p}/{host}/{port}`

### Related Issue(s)

Closes #2036

### Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

### Testing

- [x] All existing signaling_method tests pass (68 tests)
- [x] All existing connection::outgoing tests pass (15 tests)
- [x] New multiaddr parsing/conversion tests added and pass
- [x] Legacy format parsing still works with deprecation warning
- [x] Verified doc changes manually
- [x] Verified deprecation warning appears when running webnode
- [x] Verified web node is able to connect to local Rust node with legacy and standard formats

```bash
cargo test -p mina-p2p signaling_method::tests
cargo test -p mina-p2p connection::outgoing::tests
```

### Changelog

**Changes**
- **Web Node**: Standardized WebRTC peer address to Multiaddr and deprecated
  old "WebRTC-Multiaddrish" format [#2052](https://github.com/o1-labs/mina-rust/pull/2052)

### DIFF
--- a/crates/p2p/src/connection/outgoing/mod.rs
+++ b/crates/p2p/src/connection/outgoing/mod.rs
@@ -323,6 +323,80 @@ impl P2pConnectionOutgoingInitOpts {
     pub fn from_libp2p_socket_addr(peer_id: PeerId, addr: SocketAddr) -> Self {
         P2pConnectionOutgoingInitOpts::LibP2P((peer_id, addr).into())
     }
+
+    fn parse_p2p_relay_webrtc_multiaddr(
+        maddr: &multiaddr::Multiaddr,
+    ) -> Result<Self, P2pConnectionOutgoingInitOptsParseError> {
+        let mut iter = maddr.iter();
+
+        let Some(Protocol::P2p(relay_peer_id_hash)) = iter.next() else {
+            return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                "expected p2p protocol for relay".to_string(),
+            ));
+        };
+        let relay_peer_id = libp2p_identity::PeerId::from_multihash(relay_peer_id_hash.into())
+            .map_err(|_| {
+                P2pConnectionOutgoingInitOptsParseError::Other(
+                    "invalid relay peer_id multihash".to_string(),
+                )
+            })?
+            .try_into()
+            .map_err(|_| {
+                P2pConnectionOutgoingInitOptsParseError::Other(
+                    "unexpected error converting relay PeerId".to_string(),
+                )
+            })?;
+
+        // Expect /webrtc
+        if iter.next() != Some(Protocol::WebRTC) {
+            return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                "expected webrtc protocol".to_string(),
+            ));
+        };
+
+        // Expect /p2p-circuit
+        if iter.next() != Some(Protocol::P2pCircuit) {
+            return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                "expected p2p-circuit protocol".to_string(),
+            ));
+        };
+
+        // Get target peer_id
+        let peer_id = Self::parse_p2p_peer_id(iter.next(), "target")?;
+
+        Ok(Self::WebRTC {
+            peer_id,
+            signaling: webrtc::SignalingMethod::P2p { relay_peer_id },
+        })
+    }
+
+    fn parse_p2p_peer_id(
+        protocol: Option<multiaddr::Protocol>,
+        label: &'static str,
+    ) -> Result<PeerId, P2pConnectionOutgoingInitOptsParseError> {
+        match protocol {
+            Some(Protocol::P2p(peer_id_hash)) => {
+                libp2p_identity::PeerId::from_multihash(peer_id_hash.into())
+                    .map_err(|_| {
+                        P2pConnectionOutgoingInitOptsParseError::Other(format!(
+                            "invalid {label} peer_id multihash"
+                        ))
+                    })?
+                    .try_into()
+                    .map_err(|_| {
+                        P2pConnectionOutgoingInitOptsParseError::Other(format!(
+                            "unexpected error converting {label} PeerId"
+                        ))
+                    })
+            }
+            Some(other_protocol) => Err(P2pConnectionOutgoingInitOptsParseError::Other(format!(
+                "expected p2p protocol for {label} peer id, got {other_protocol:?}"
+            ))),
+            None => Err(P2pConnectionOutgoingInitOptsParseError::Other(format!(
+                "missing {label} peer id"
+            ))),
+        }
+    }
 }
 
 impl P2pConnectionOutgoingInitLibp2pOpts {
@@ -451,171 +525,102 @@ impl TryFrom<&multiaddr::Multiaddr> for P2pConnectionOutgoingInitOpts {
         // Check if this is a WebRTC multiaddr
         let is_webrtc = maddr.iter().any(|p| p == Protocol::WebRTC);
 
-        if is_webrtc {
-            let mut iter = maddr.iter().peekable();
+        // Standard libp2p multiaddr (no /webrtc)
+        if !is_webrtc {
+            cfg_if::cfg_if! {
+                if #[cfg(target_arch = "wasm32")] {
+                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                        "libp2p not supported in wasm".to_owned(),
+                    ))
+                } else {
+                    return Ok(Self::LibP2P(maddr.try_into()?))
+                }
+            };
+        }
 
-            // Check for P2P relay format: /p2p/{relay}/webrtc/p2p-circuit/p2p/{target}
-            if matches!(iter.peek(), Some(Protocol::P2p(_))) {
-                let relay_peer_id = match iter.next() {
-                    Some(Protocol::P2p(hash)) => {
-                        libp2p_identity::PeerId::from_multihash(hash.into())
-                            .map_err(|_| {
-                                P2pConnectionOutgoingInitOptsParseError::Other(
-                                    "invalid relay peer_id multihash".to_string(),
-                                )
-                            })?
-                            .try_into()
-                            .map_err(|_| {
-                                P2pConnectionOutgoingInitOptsParseError::Other(
-                                    "unexpected error converting relay PeerId".to_string(),
-                                )
-                            })?
+        let mut iter = maddr.iter();
+
+        // Check for P2P relay format: /p2p/{relay}/webrtc/p2p-circuit/p2p/{target}
+        // and go to parse_p2p_relay_webrtc_multiaddr.
+        // Otherwise, parse one of the HTTP-based variants:
+        // /dns|dns4|dns6|ip4|ip6/{host}/tcp/{port}/webrtc/http|https/[http-proxy/{proxy_path}/]p2p/{peer_id}
+        match iter.next() {
+            Some(Protocol::P2p(_)) => Self::parse_p2p_relay_webrtc_multiaddr(maddr),
+            other_transport_protocol => {
+                // Extract /dns|dns4|dns6|ip4|ip6/{host}
+                let host = match other_transport_protocol {
+                    Some(Protocol::Ip4(v)) => Host::Ipv4(v),
+                    Some(Protocol::Ip6(v)) => Host::Ipv6(v),
+                    Some(Protocol::Dns(v) | Protocol::Dns4(v) | Protocol::Dns6(v)) => {
+                        Host::Domain(v.to_string())
                     }
                     _ => {
                         return Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                            "expected p2p protocol for relay".to_string(),
+                            "expected host (dns/dns4/dns6/ip4/ip6) in webrtc multiaddr".to_string(),
                         ))
                     }
+                };
+
+                // Extract /tcp/{port}
+                let Some(Protocol::Tcp(port)) = iter.next() else {
+                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                        "expected tcp port in webrtc multiaddr".to_string(),
+                    ));
                 };
 
                 // Skip /webrtc
-                match iter.next() {
-                    Some(Protocol::WebRTC) => {}
+                if iter.next() != Some(Protocol::WebRTC) {
+                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                        "expected webrtc protocol".to_string(),
+                    ));
+                };
+
+                // Determine signaling method: http, https, or proxy with http-path
+                let signaling_info = HttpSignalingInfo { host, port };
+                let scheme = match iter.next() {
+                    Some(Protocol::Http) => webrtc::ProxyScheme::Http,
+                    Some(Protocol::Https) => webrtc::ProxyScheme::Https,
                     _ => {
                         return Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                            "expected webrtc protocol".to_string(),
+                            "expected http or https protocol after webrtc".to_string(),
                         ))
                     }
-                }
-
-                // Expect /p2p-circuit
-                match iter.next() {
-                    Some(Protocol::P2pCircuit) => {}
-                    _ => {
-                        return Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                            "expected p2p-circuit protocol".to_string(),
-                        ))
+                };
+                let (signaling, peer_id) = match iter.next() {
+                    Some(Protocol::HttpPath(path)) => {
+                        let signaling = webrtc::SignalingMethod::Proxied(
+                            scheme,
+                            webrtc::PathPrefix(path.to_string()),
+                            signaling_info,
+                        );
+                        let peer_id = Self::parse_p2p_peer_id(iter.next(), "webrtc")?;
+                        (signaling, peer_id)
                     }
-                }
-
-                // Get target peer_id
-                let peer_id = match iter.next() {
-                    Some(Protocol::P2p(hash)) => {
-                        libp2p_identity::PeerId::from_multihash(hash.into())
-                            .map_err(|_| {
-                                P2pConnectionOutgoingInitOptsParseError::Other(
-                                    "invalid target peer_id multihash".to_string(),
-                                )
-                            })?
-                            .try_into()
-                            .map_err(|_| {
-                                P2pConnectionOutgoingInitOptsParseError::Other(
-                                    "unexpected error converting target PeerId".to_string(),
-                                )
-                            })?
+                    p2p @ Some(Protocol::P2p(_)) => {
+                        let signaling = match scheme {
+                            webrtc::ProxyScheme::Http => {
+                                webrtc::SignalingMethod::Http(signaling_info)
+                            }
+                            webrtc::ProxyScheme::Https => {
+                                webrtc::SignalingMethod::Https(signaling_info)
+                            }
+                        };
+                        let peer_id = Self::parse_p2p_peer_id(p2p, "webrtc")?;
+                        (signaling, peer_id)
                     }
-                    _ => {
+                    Some(other_protocol) => {
                         return Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                            "expected p2p protocol for target peer".to_string(),
+                            format!("expected /p2p/peer_id or /http-path/encoded_path, got {other_protocol:?}"
+                        )))
+                    },
+                    None => {
+                        return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                            "expected p2p protocol with peer_id".to_string(),
                         ))
                     }
                 };
 
-                return Ok(Self::WebRTC {
-                    peer_id,
-                    signaling: webrtc::SignalingMethod::P2p { relay_peer_id },
-                });
-            }
-
-            // HTTP(S) signaling format: /dns4|ip4/{host}/tcp/{port}/webrtc/http|https/p2p/{peer_id}
-            let host = match iter.next() {
-                Some(Protocol::Ip4(v)) => Host::Ipv4(v),
-                Some(Protocol::Ip6(v)) => Host::Ipv6(v),
-                Some(Protocol::Dns(v) | Protocol::Dns4(v) | Protocol::Dns6(v)) => {
-                    Host::Domain(v.to_string())
-                }
-                _ => {
-                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                        "expected host (ip4/ip6/dns) in webrtc multiaddr".to_string(),
-                    ))
-                }
-            };
-
-            let port = match iter.next() {
-                Some(Protocol::Tcp(p)) => p,
-                _ => {
-                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                        "expected tcp port in webrtc multiaddr".to_string(),
-                    ))
-                }
-            };
-
-            // Skip /webrtc
-            match iter.next() {
-                Some(Protocol::WebRTC) => {}
-                _ => {
-                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                        "expected webrtc protocol".to_string(),
-                    ))
-                }
-            }
-
-            // Determine signaling method: http, https, or proxy with http-path
-            let signaling_info = HttpSignalingInfo { host, port };
-            let scheme = match iter.next() {
-                Some(Protocol::Http) => webrtc::ProxyScheme::Http,
-                Some(Protocol::Https) => webrtc::ProxyScheme::Https,
-                _ => {
-                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                        "expected http or https protocol after webrtc".to_string(),
-                    ))
-                }
-            };
-            let signaling = if let Some(Protocol::HttpPath(path)) = iter.peek() {
-                let path = path.to_string();
-                iter.next(); // consume the peeked item
-                webrtc::SignalingMethod::Proxied(scheme, webrtc::PathPrefix(path), signaling_info)
-            } else {
-                match scheme {
-                    webrtc::ProxyScheme::Http => webrtc::SignalingMethod::Http(signaling_info),
-                    webrtc::ProxyScheme::Https => webrtc::SignalingMethod::Https(signaling_info),
-                }
-            };
-
-            // Get peer_id
-            let peer_id = match iter.next() {
-                Some(Protocol::P2p(hash)) => libp2p_identity::PeerId::from_multihash(hash.into())
-                    .map_err(|_| {
-                        P2pConnectionOutgoingInitOptsParseError::Other(
-                            "invalid peer_id multihash".to_string(),
-                        )
-                    })?
-                    .try_into()
-                    .map_err(|_| {
-                        P2pConnectionOutgoingInitOptsParseError::Other(
-                            "unexpected error converting PeerId".to_string(),
-                        )
-                    })?,
-                _ => {
-                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                        "expected p2p protocol with peer_id".to_string(),
-                    ))
-                }
-            };
-
-            Ok(Self::WebRTC { peer_id, signaling })
-        } else {
-            // Standard libp2p multiaddr (no /webrtc)
-            #[cfg(target_arch = "wasm32")]
-            {
-                Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                    "libp2p not supported in wasm".to_owned(),
-                ))
-            }
-
-            #[cfg(not(target_arch = "wasm32"))]
-            {
-                Ok(Self::LibP2P(maddr.try_into()?))
+                Ok(Self::WebRTC { peer_id, signaling })
             }
         }
     }
@@ -644,6 +649,13 @@ impl From<&P2pConnectionOutgoingInitOpts> for Multiaddr {
             P2pConnectionOutgoingInitOpts::WebRTC { peer_id, signaling } => {
                 use webrtc::SignalingMethod;
 
+                // expect() safety: by the time we have a P2pConnectionOutgoingInitOpts
+                // peer_id was already validated. This is a quirk of the libp2p_identity vs mina-p2p types
+                // validation logics in:
+                // 1. P2pConnectionOutgoingInitLibp2pOpts::try_from_mina_rpc()
+                // 2. impl TryFrom<&multiaddr::Multiaddr> for P2pConnectionOutgoingInitLibp2pOpts
+                // P2pConnectionOutgoingInitOpts will never come in over the wire or from a rogue peer,
+                // possibly only bad CLI flags.
                 let peer_id_proto = Protocol::P2p(
                     libp2p_identity::PeerId::try_from(*peer_id).expect("valid peer_id"),
                 );
@@ -710,6 +722,7 @@ impl From<&P2pConnectionOutgoingInitOpts> for Multiaddr {
                             .with(peer_id_proto)
                     }
                     SignalingMethod::P2p { relay_peer_id } => {
+                        // same expect() safety as peer_id_proto
                         let relay_id_proto = Protocol::P2p(
                             libp2p_identity::PeerId::try_from(*relay_peer_id)
                                 .expect("valid relay_peer_id"),
@@ -722,6 +735,7 @@ impl From<&P2pConnectionOutgoingInitOpts> for Multiaddr {
                     }
                 }
             }
+            // same expect() safety rationale as the others. it's all from peer_id >__>
             P2pConnectionOutgoingInitOpts::LibP2P(v) => v.to_maddr().expect("valid libp2p opts"),
         }
     }
@@ -746,23 +760,23 @@ impl TryFrom<&multiaddr::Multiaddr> for P2pConnectionOutgoingInitLibp2pOpts {
                 Some(Protocol::Dns(v) | Protocol::Dns4(v) | Protocol::Dns6(v)) => {
                     Host::Domain(v.to_string())
                 }
-                Some(_) => {
+                Some(other_host) => {
                     return Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                        "unexpected part in multiaddr! expected host".to_string(),
+                        format!("unexpected transport in multiaddr! expected /dns|dns4|dns6|ip4|ip6/<host>, got {other_host:?}!")
                     ));
                 }
                 None => {
                     return Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                        "missing host part from multiaddr".to_string(),
+                        "missing /dns|dns4|dns6|ip4|ip6/host from multiaddr".to_string(),
                     ));
                 }
             },
             port: match iter.next() {
                 Some(Protocol::Tcp(port)) => port,
-                Some(_) => {
-                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                        "unexpected part in multiaddr! expected port".to_string(),
-                    ));
+                Some(other_port) => {
+                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(format!(
+                        "unexpected part in multiaddr! expected /tcp/<port>, got {other_port:?}"
+                    )));
                 }
                 None => {
                     return Err(P2pConnectionOutgoingInitOptsParseError::Other(
@@ -814,13 +828,6 @@ mod measurement {
 
 #[cfg(test)]
 mod tests {
-    //! Unit tests for P2pConnectionOutgoingInitOpts multiaddr parsing
-    //!
-    //! Run these tests with:
-    //! ```bash
-    //! cargo test -p mina-p2p connection::outgoing::tests
-    //! ```
-
     use super::*;
     use std::net::Ipv4Addr;
 
@@ -842,9 +849,9 @@ mod tests {
                     assert_eq!(info.host, Host::Domain("signal.example.com".to_string()));
                     assert_eq!(info.port, 8080);
                 }
-                _ => panic!("Expected Http signaling method"),
+                x => panic!("Expected Http signaling method, got {x:?}"),
             },
-            _ => panic!("Expected WebRTC variant"),
+            x => panic!("Expected WebRTC variant, got {x:?}"),
         }
     }
 
@@ -862,9 +869,9 @@ mod tests {
                     assert_eq!(info.host, Host::Ipv4(Ipv4Addr::new(192, 168, 1, 100)));
                     assert_eq!(info.port, 8080);
                 }
-                _ => panic!("Expected Http signaling method"),
+                x => panic!("Expected Http signaling method, got {x:?}"),
             },
-            _ => panic!("Expected WebRTC variant"),
+            x => panic!("Expected WebRTC variant, got {x:?}"),
         }
     }
 
@@ -882,9 +889,9 @@ mod tests {
                     assert_eq!(info.host, Host::Domain("signal.example.com".to_string()));
                     assert_eq!(info.port, 443);
                 }
-                _ => panic!("Expected Https signaling method"),
+                x => panic!("Expected Https signaling method, got {x:?}"),
             },
-            _ => panic!("Expected WebRTC variant"),
+            x => panic!("Expected WebRTC variant, got {x:?}"),
         }
     }
 
@@ -904,9 +911,9 @@ mod tests {
                     assert_eq!(info.host, Host::Domain("proxy.example.com".to_string()));
                     assert_eq!(info.port, 443);
                 }
-                _ => panic!("Expected Proxied signaling method"),
+                x => panic!("Expected Proxied signaling method, got {x:?}"),
             },
-            _ => panic!("Expected WebRTC variant"),
+            x => panic!("Expected WebRTC variant, got {x:?}"),
         }
     }
 
@@ -923,9 +930,9 @@ mod tests {
                 webrtc::SignalingMethod::P2p { .. } => {
                     // Successfully parsed as P2P relay
                 }
-                _ => panic!("Expected P2p signaling method"),
+                x => panic!("Expected P2p signaling method, got {x:?}"),
             },
-            _ => panic!("Expected WebRTC variant"),
+            x => panic!("Expected WebRTC variant, got {x:?}"),
         }
     }
 
@@ -945,7 +952,7 @@ mod tests {
                 );
                 assert_eq!(libp2p_opts.port, 10003);
             }
-            _ => panic!("Expected LibP2P variant"),
+            x => panic!("Expected LibP2P variant, got {x:?}"),
         }
     }
 
@@ -960,11 +967,11 @@ mod tests {
         let opts1: P2pConnectionOutgoingInitOpts = maddr_str.parse().unwrap();
 
         // Encode: convert to Multiaddr, then to string
-        let maddr: Multiaddr = (&opts1).into();
+        let maddr = Multiaddr::from(&opts1);
         let encoded_str = maddr.to_string();
 
         // Second decode: parse the encoded string
-        let opts2: P2pConnectionOutgoingInitOpts = encoded_str.parse().unwrap();
+        let opts2 = P2pConnectionOutgoingInitOpts::from_str(&encoded_str).unwrap();
 
         // Both decodes should match structurally
         assert_eq!(opts1, opts2);
@@ -1032,13 +1039,16 @@ mod tests {
         let maddr: Multiaddr = maddr_str.parse().unwrap();
         let opts: P2pConnectionOutgoingInitOpts = (&maddr).try_into().unwrap();
 
-        assert!(matches!(
-            opts,
-            P2pConnectionOutgoingInitOpts::WebRTC {
-                signaling: webrtc::SignalingMethod::Http(_),
-                ..
-            }
-        ));
+        assert!(
+            matches!(
+                opts,
+                P2pConnectionOutgoingInitOpts::WebRTC {
+                    signaling: webrtc::SignalingMethod::Http(_),
+                    ..
+                }
+            ),
+            "expected WebRTC with Http signaling, got {opts:?}"
+        );
     }
 
     #[test]
@@ -1106,9 +1116,12 @@ mod tests {
 
         match opts {
             P2pConnectionOutgoingInitOpts::WebRTC { signaling, .. } => {
-                assert!(matches!(signaling, webrtc::SignalingMethod::Http(_)));
+                assert!(
+                    matches!(signaling, webrtc::SignalingMethod::Http(_)),
+                    "expected Http signaling, got {signaling:?}"
+                );
             }
-            _ => panic!("Expected WebRTC variant"),
+            x => panic!("Expected WebRTC variant, got {x:?}"),
         }
     }
 }
--- a/crates/p2p/src/connection/outgoing/mod.rs
+++ b/crates/p2p/src/connection/outgoing/mod.rs
@@ -590,7 +590,7 @@ impl TryFrom<&multiaddr::Multiaddr> for P2pConnectionOutgoingInitOpts {
                     Some(Protocol::HttpPath(path)) => {
                         let signaling = webrtc::SignalingMethod::Proxied(
                             scheme,
-                            webrtc::PathPrefix(path.to_string()),
+                            path.into(),
                             signaling_info,
                         );
                         let peer_id = Self::parse_p2p_peer_id(iter.next(), "webrtc")?;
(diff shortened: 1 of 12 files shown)


crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -303,6 +303,17 @@ impl P2pConnectionOutgoingInitOpts {
...
+                    Some(v2::NetworkPeerPeerStableV1 {
+                        host: format!("{scheme}://{}{}", info.host, path_prefix)
+                            .as_bytes()
+                            .into(),
+                        libp2p_port: (info.port as u64).into(),
+                        peer_id: v2::NetworkPeerPeerIdStableV1(
+                            (*peer_id).to_string().into_bytes().into(),
+                        ),
+                    })
+                }
```
richardpringle, line 316: 👍

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -322,19 +333,8 @@ impl P2pConnectionOutgoingInitLibp2pOpts {
...
-            Self::LibP2P(v) => {
-                if let Some(maddr) = v.to_maddr() {
-                    write!(f, "{}", maddr)
-                } else {
-                    write!(f, "*INVALID MULTIADDRESS*")
-                }
-            }
-        }
+        let maddr: Multiaddr = self.into();
+        write!(f, "{}", maddr)
```
richardpringle, line 411: 👍

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -358,30 +358,20 @@ impl FromStr for P2pConnectionOutgoingInitOpts {
...
-            return Ok(Self::LibP2P(opts));
-        }
-        #[cfg(target_arch = "wasm32")]
-        if is_libp2p_maddr {
-            return Err(P2pConnectionOutgoingInitOptsParseError::Other(
-                "libp2p not supported in wasm".to_owned(),
-            ));
+        // Try parsing as multiaddr first (the preferred format)
+        if let Ok(maddr) = Multiaddr::from_str(s) {
+            return Self::try_from(&maddr);
```
richardpringle, line 437: 👍

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -358,30 +358,20 @@ impl FromStr for P2pConnectionOutgoingInitOpts {
...
         }
 
+        // Fallback: try legacy WebRTC format (/{peer_id}/{signaling_method}/...)
+        // This format is deprecated; prefer multiaddr format instead.
         let id_end_index = s[1..]
             .find('/')
             .map(|i| i + 1)
             .filter(|i| s.len() > *i)
             .ok_or(P2pConnectionOutgoingInitOptsParseError::NotEnoughArgs)?;
```
richardpringle, line 446: This is... fragile. Think it's worth improving while we're editing this code? These `FromStr` implementations (at least this and `webrtc::SignalingMethod` implementations) rely on the leading `/` instead of using the `/` as a delimiter and parsing the actual stuff between the `/`. 

Personally, I would be in favour of changing this now so that it doesn't come to bite us at some other point. I realize that would scope creep on this PR though... so, you're call.

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -390,7 +380,17 @@ impl FromStr for P2pConnectionOutgoingInitOpts {
...
+
+        // Emit deprecation warning with the suggested multiaddr format
+        let suggested_maddr: Multiaddr = (&opts).into();
+        tracing::warn!(
+            message = "Deprecated address format detected. Please use multiaddr format instead.",
+            legacy_format = %s,
+            suggested_format = %suggested_maddr,
+        );
+
+        Ok(opts)
```
richardpringle, line 467: 👍

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+            {
+                Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                    "libp2p not supported in wasm".to_owned(),
+                ))
+            }
+
+            #[cfg(not(target_arch = "wasm32"))]
+            {
+                Ok(Self::LibP2P(maddr.try_into()?))
+            }
```
richardpringle: We should use [`cfg-if`](https://docs.rs/cfg-if/latest/cfg_if/) here. Can do in a follow up

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+    /// - Proxied: `/<dns|dns4|dns6|ip4|ip6>/{host}/tcp/{port}/webrtc/<http|https>/http-path/{path}/p2p/{peer_id}`
+    /// - P2P Relay: `/p2p/{relay_peer_id}/webrtc/p2p-circuit/p2p/{target_peer_id}`
+    ///
+    /// **LibP2P format**:
+    /// - `/<dns|dns4|dns6|ip4|ip6>/{host}/tcp/{port}/p2p/{peer_id}`
+    fn try_from(maddr: &multiaddr::Multiaddr) -> Result<Self, Self::Error> {
+        // Check if this is a WebRTC multiaddr
+        let is_webrtc = maddr.iter().any(|p| p == Protocol::WebRTC);
+
+        if is_webrtc {
```
richardpringle: This is one of those times where I think there's enough code here to say this isn't a nit
```rust
if !maddr.iter().any(|p| p == Protocol::WebRTC) {
  return // stuff in the else block
}

// everything that wasn't in the else block
```

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+            let mut iter = maddr.iter().peekable();
+
+            // Check for P2P relay format: /p2p/{relay}/webrtc/p2p-circuit/p2p/{target}
+            if matches!(iter.peek(), Some(Protocol::P2p(_))) {
```
richardpringle: I'm being a bit of a pedant here but it would be a lot nicer if we did something like:
```rust
// I think we have to do this to keep `iter` working as expected in the match body, but I could be wrong.
let first = iter.next();

match first {
    Some(Protocol::P2p(hash)) => handle_p2p_parsing(&mut iter, hash),
    host => // ... (could be None here, but you would handle that in the body as the error is the same
}
```

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+                                    "unexpected error converting relay PeerId".to_string(),
+                                )
+                            })?
+                    }
+                    _ => {
+                        return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                            "expected p2p protocol for relay".to_string(),
+                        ))
+                    }
+                };
```
richardpringle: I'm being a bit of a pedant, but my suggestions are going to flatten out the code making it a lot easier to follow

```suggestion
                let Some(Protocol::P2p(hash)) = iter.next() else {
                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
                        "expected p2p protocol for relay".to_string(),
                    ));
                };

                let relay_peer_id = libp2p_identity::PeerId::from_multihash(hash.into())
                    .map_err(|_| {
                        P2pConnectionOutgoingInitOptsParseError::Other(
                            "invalid relay peer_id multihash".to_string(),
                        )
                    })?
                    .try_into()
                    .map_err(|_| {
                        P2pConnectionOutgoingInitOptsParseError::Other(
                            "unexpected error converting relay PeerId".to_string(),
                        )
                    })?;
```

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+
+                // Skip /webrtc
+                match iter.next() {
+                    Some(Protocol::WebRTC) => {}
+                    _ => {
+                        return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                            "expected webrtc protocol".to_string(),
+                        ))
+                    }
+                }
```
richardpringle:
```suggestion
                if iter.next() != Some(Protocol::WebRTC) {
                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
                        "expected webrtc protocol".to_string(),
                    ));
                }
```

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+
+                // Expect /p2p-circuit
+                match iter.next() {
+                    Some(Protocol::P2pCircuit) => {}
+                    _ => {
+                        return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                            "expected p2p-circuit protocol".to_string(),
+                        ))
+                    }
+                }
```
richardpringle:
```suggestion
                if iter.next() != Some(Protocol::WebRTC) {
                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
                        "expected webrtc protocol".to_string(),
                    ));
                }
```

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+                                    "unexpected error converting target PeerId".to_string(),
+                                )
+                            })?
+                    }
+                    _ => {
+                        return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                            "expected p2p protocol for target peer".to_string(),
+                        ))
+                    }
+                };
```
richardpringle:
```suggestion
                let Some(Protocol::P2p(hash)) = iter.next() else {
                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
                        "expected p2p protocol for target peer".to_string(),
                    ));
                };

                let peer_id = libp2p_identity::PeerId::from_multihash(hash.into())
                    .map_err(|_| {
                        P2pConnectionOutgoingInitOptsParseError::Other(
                            "invalid target peer_id multihash".to_string(),
                        )
                    })?
                    .try_into()
                    .map_err(|_| {
                        P2pConnectionOutgoingInitOptsParseError::Other(
                            "unexpected error converting target PeerId".to_string(),
                        )
                    })?;
```

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+                Some(Protocol::Ip6(v)) => Host::Ipv6(v),
+                Some(Protocol::Dns(v) | Protocol::Dns4(v) | Protocol::Dns6(v)) => {
+                    Host::Domain(v.to_string())
+                }
+                _ => {
+                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                        "expected host (ip4/ip6/dns) in webrtc multiaddr".to_string(),
+                    ))
+                }
+            };
```
richardpringle: 👍

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+
+            let port = match iter.next() {
+                Some(Protocol::Tcp(p)) => p,
+                _ => {
+                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                        "expected tcp port in webrtc multiaddr".to_string(),
+                    ))
+                }
+            };
```
richardpringle:
```suggestion
            let Some(Protocol::Tcp(port)) = iter.next() else {
                return Err(P2pConnectionOutgoingInitOptsParseError::Other(
                    "expected tcp port in webrtc multiaddr".to_string(),
                ));
            };
```

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+
+            // Skip /webrtc
+            match iter.next() {
+                Some(Protocol::WebRTC) => {}
+                _ => {
+                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                        "expected webrtc protocol".to_string(),
+                    ))
+                }
+            }
```
richardpringle:
```suggestion
            if iter.next() != Some(Protocol::WebRTC) {
                return Err(P2pConnectionOutgoingInitOptsParseError::Other(
                    "expected webrtc protocol".to_string(),
                ));
            }
```

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+                Some(Protocol::Https) => webrtc::ProxyScheme::Https,
+                _ => {
+                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                        "expected http or https protocol after webrtc".to_string(),
+                    ))
+                }
+            };
+            let signaling = if let Some(Protocol::HttpPath(path)) = iter.peek() {
+                let path = path.to_string();
+                iter.next(); // consume the peeked item
```
richardpringle: 😳 

There's an optional path here that only exists for proxied signaling? yikes...

crates/p2p/src/connection/outgoing/mod.rs
richardpringle: If you do the above along with one other change, I don't think you need to make the iterator peekable

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+                        P2pConnectionOutgoingInitOptsParseError::Other(
+                            "unexpected error converting PeerId".to_string(),
+                        )
+                    })?,
+                _ => {
+                    return Err(P2pConnectionOutgoingInitOptsParseError::Other(
+                        "expected p2p protocol with peer_id".to_string(),
+                    ))
+                }
+            };
```
richardpringle: If you do the following, along with a comment above, you don't have to make the iterator peekable. In general, peekable iterators are a bit of a code-smell IMO. There are definitely times when you need them, but I think it's pretty rare. That optional path parameter for the proxy is the thing making this code ugly, though. 

```suggestion
            let (signaling, peer_id) = match iter.next() {
                Some(Protocol::HttpPath(path)) => {
                    let signaling = webrtc::SignalingMethod::Proxied(
                        scheme,
                        webrtc::PathPrefix(path.to_string()),
                        signaling_info,
                    );

                    (signaling, iter.next())
                }
                peer_id => {
                    let signaling = match scheme {
                        webrtc::ProxyScheme::Http => webrtc::SignalingMethod::Http(signaling_info),
                        webrtc::ProxyScheme::Https => {
                            webrtc::SignalingMethod::Https(signaling_info)
                        }
                    };

                    (signaling, peer_id)
                }
            };

            // Get peer_id
            let Some(Protocol::P2p(hash)) = peer_id else {
                return Err(P2pConnectionOutgoingInitOptsParseError::Other(
                    "expected p2p protocol with peer_id".to_string(),
                ));
            };

            let peer_id = libp2p_identity::PeerId::from_multihash(hash.into())
                .map_err(|_| {
                    P2pConnectionOutgoingInitOptsParseError::Other(
                        "invalid peer_id multihash".to_string(),
                    )
                })?
                .try_into()
                .map_err(|_| {
                    P2pConnectionOutgoingInitOptsParseError::Other(
                        "unexpected error converting PeerId".to_string(),
                    )
                })?;
```

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+                            .with(Protocol::Tcp(info.port))
+                            .with(Protocol::WebRTC)
+                            .with(scheme_proto)
+                            .with(Protocol::HttpPath(path.into()))
+                            .with(peer_id_proto)
+                    }
+                    SignalingMethod::P2p { relay_peer_id } => {
+                        let relay_id_proto = Protocol::P2p(
+                            libp2p_identity::PeerId::try_from(*relay_peer_id)
+                                .expect("valid relay_peer_id"),
```
richardpringle, line 728: I think anywhere in the code that can panic should have a big comment as to why it's expected that this code wouldn't panic and why we can't handle the error in the case where it does cause a panic. 

I would think that nothing in this crate should ever cause the node to crash...

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
+                        );
+                        Multiaddr::empty()
+                            .with(relay_id_proto)
+                            .with(Protocol::WebRTC)
+                            .with(Protocol::P2pCircuit)
+                            .with(peer_id_proto)
+                    }
+                }
+            }
+            P2pConnectionOutgoingInitOpts::LibP2P(v) => v.to_maddr().expect("valid libp2p opts"),
```
richardpringle, line 739: same comment as above

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -527,3 +811,304 @@ mod measurement {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    //! Unit tests for P2pConnectionOutgoingInitOpts multiaddr parsing
```
richardpringle: This is a module comment. This module `connection::outgoing::tests` should be for testing *anything* in its parent, not just this one type.

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -527,3 +811,304 @@ mod measurement {
...
+    //!
+    //! Run these tests with:
+    //! ```bash
+    //! cargo test -p mina-p2p connection::outgoing::tests
+    //! ```
```
richardpringle: I think it's okay to expect anyone reading this code to know how to test a specific module.

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -527,3 +811,304 @@ mod measurement {
...
+        );
+        let opts: P2pConnectionOutgoingInitOpts = maddr_str.parse().unwrap();
+
+        match opts {
+            P2pConnectionOutgoingInitOpts::WebRTC { signaling, .. } => match signaling {
+                webrtc::SignalingMethod::Http(info) => {
+                    assert_eq!(info.host, Host::Domain("signal.example.com".to_string()));
+                    assert_eq!(info.port, 8080);
+                }
+                _ => panic!("Expected Http signaling method"),
```
richardpringle: Is the signaling-method `Debug`? I would print it out with the `panic` mesage

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -527,3 +811,304 @@ mod measurement {
...
+            },
+            _ => panic!("Expected WebRTC variant"),
```
richardpringle: I would print out what we got here in the panic message (if it's `Debug`)

crates/p2p/src/connection/outgoing/mod.rs
richardpringle: confirmed, it is `Debug`

crates/p2p/src/connection/outgoing/mod.rs
richardpringle: confirmed it is `Debug`

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -527,3 +811,304 @@ mod measurement {
...
+
+        match opts {
+            P2pConnectionOutgoingInitOpts::WebRTC { signaling, .. } => match signaling {
+                webrtc::SignalingMethod::Http(info) => {
+                    assert_eq!(info.host, Host::Ipv4(Ipv4Addr::new(192, 168, 1, 100)));
+                    assert_eq!(info.port, 8080);
+                }
+                _ => panic!("Expected Http signaling method"),
+            },
+            _ => panic!("Expected WebRTC variant"),
```
richardpringle: Same comment

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -527,3 +811,304 @@ mod measurement {
...
+
+        match opts {
+            P2pConnectionOutgoingInitOpts::WebRTC { signaling, .. } => match signaling {
+                webrtc::SignalingMethod::Https(info) => {
+                    assert_eq!(info.host, Host::Domain("signal.example.com".to_string()));
+                    assert_eq!(info.port, 443);
+                }
+                _ => panic!("Expected Https signaling method"),
+            },
+            _ => panic!("Expected WebRTC variant"),
```
richardpringle: Same comment

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -527,3 +811,304 @@ mod measurement {
...
+            P2pConnectionOutgoingInitOpts::WebRTC { signaling, .. } => match signaling {
+                webrtc::SignalingMethod::Proxied(scheme, webrtc::PathPrefix(path), info) => {
+                    assert_eq!(scheme, webrtc::ProxyScheme::Https);
+                    assert_eq!(path, "cluster/123/mina/webrtc/signal");
+                    assert_eq!(info.host, Host::Domain("proxy.example.com".to_string()));
+                    assert_eq!(info.port, 443);
+                }
+                _ => panic!("Expected Proxied signaling method"),
+            },
+            _ => panic!("Expected WebRTC variant"),
```
richardpringle: Same comment

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -527,3 +811,304 @@ mod measurement {
...
+        let opts: P2pConnectionOutgoingInitOpts = maddr_str.parse().unwrap();
+
+        match opts {
+            P2pConnectionOutgoingInitOpts::WebRTC { signaling, .. } => match signaling {
+                webrtc::SignalingMethod::P2p { .. } => {
+                    // Successfully parsed as P2P relay
+                }
+                _ => panic!("Expected P2p signaling method"),
+            },
+            _ => panic!("Expected WebRTC variant"),
```
richardpringle: Same comment

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -527,3 +811,304 @@ mod measurement {
...
+
+        match opts {
+            P2pConnectionOutgoingInitOpts::LibP2P(libp2p_opts) => {
+                assert_eq!(
+                    libp2p_opts.host,
+                    Host::Domain("seed-1.devnet.gcp.o1test.net".to_string())
+                );
+                assert_eq!(libp2p_opts.port, 10003);
+            }
+            _ => panic!("Expected LibP2P variant"),
```
richardpringle: Same comment

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -436,16 +436,300 @@ impl TryFrom<P2pConnectionOutgoingInitLibp2pOpts> for multiaddr::Multiaddr {
...
 impl TryFrom<multiaddr::Multiaddr> for P2pConnectionOutgoingInitOpts {
     type Error = P2pConnectionOutgoingInitOptsParseError;
 
     fn try_from(value: multiaddr::Multiaddr) -> Result<Self, Self::Error> {
-        Ok(Self::LibP2P((&value).try_into()?))
+        (&value).try_into()
+    }
+}
+
+impl From<&P2pConnectionOutgoingInitOpts> for Multiaddr {
```
richardpringle, line 637:
```suggestion
impl From<P2pConnectionOutgoingInitOpts> for Multiaddr {
```
I think you take ownership in every single branch. In general, it's better to force the caller to call `Clone::clone` outside so that it's clear that there are potential allocations.

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -527,3 +811,304 @@ mod measurement {
...
+        let maddr_str = format!(
+            "/dns4/signal.example.com/tcp/8080/webrtc/http/p2p/{}",
+            TEST_PEER_ID_LIBP2P
+        );
+
+        // First decode: parse multiaddr string
+        let opts1: P2pConnectionOutgoingInitOpts = maddr_str.parse().unwrap();
+
+        // Encode: convert to Multiaddr, then to string
+        let maddr: Multiaddr = (&opts1).into();
```
richardpringle: Haha, this isn't worth changing since it's in a test, but generally, I only use `Into::into` if the type can be inferred, otherwise I find `let maddr = Multiaddr::from(opts1);` more concise.

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -527,3 +811,304 @@ mod measurement {
...
+        let encoded_str = maddr.to_string();
+
+        // Second decode: parse the encoded string
+        let opts2: P2pConnectionOutgoingInitOpts = encoded_str.parse().unwrap();
```
richardpringle: Same goes for `Type::from_str` instead of `let x: Type = some_str.parse().unwrap()`. 

Again, doesn't really matter for tests, not suggesting you change this now.

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -527,3 +811,304 @@ mod measurement {
...
+
+        // Test that legacy format /{peer_id}/{signaling} still parses
+        let legacy_str = format!("/{}/http/signal.example.com/8080", legacy_peer_id_str);
+        let opts: P2pConnectionOutgoingInitOpts = legacy_str.parse().unwrap();
+
+        match opts {
+            P2pConnectionOutgoingInitOpts::WebRTC { signaling, .. } => {
+                assert!(matches!(signaling, webrtc::SignalingMethod::Http(_)));
+            }
+            _ => panic!("Expected WebRTC variant"),
```
richardpringle: should print what you got instead.

crates/p2p/src/connection/outgoing/mod.rs
```diff
@@ -527,3 +811,304 @@ mod measurement {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
richardpringle, line 830: Couple comments about the tests. 
1. It seems like we're only testing valid string parse the way we expect them to (and the roundtrips 🥇), should we be testing invalid strings error out the way we expect them to? This can be done in a follow up unless there's a good argument that it doesn't matter (I wouldn't be surprised if this is the case). 
2. If we're keeping the panics in here, we should probably have tests that trigger them specifically? Haha... yes, that '?' is there on purpose. This seems like an attack vector. I'm not sure if we have defenses elsewhere, but whenever I see strings, I assume they came from the outside world and there seem to be specific strings that would cause a panic and take down the whole node. I could be wrong, but we should probably have some specific tests for this. 

1 and 2 are kind of the same thing, but I specifically wanted to call out the panicking code separately.

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -48,14 +49,70 @@
...
+/// URL path prefix for proxy signaling.
+///
+/// This newtype wraps a String path prefix (e.g., "/clusters/123") and provides
+/// BinProt serialization by encoding as a length-prefixed byte array.
+///
+/// Used by `Proxied` variant for flexible path-based proxy configurations.
+/// The legacy `HttpsProxy(u16, HttpSignalingInfo)` is preserved for BinProt
+/// backward compatibility.
+#[derive(Eq, PartialEq, Ord, PartialOrd, Debug, Clone, derive_more::Display)]
+#[display(fmt = "{_0}")]
```
richardpringle: Why the `_`?

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -48,14 +49,70 @@
...
+pub struct PathPrefix(pub String);
```
richardpringle: Does the string need to be pub? Should it be `pub(crate)`?

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -48,14 +49,70 @@
...
+    Http,
+    /// Secure HTTPS proxy connection.
+    #[display(fmt = "https")]
+    Https,
+}
+
+impl BinProtRead for PathPrefix {
+    fn binprot_read<R: std::io::Read + ?Sized>(r: &mut R) -> Result<Self, binprot::Error>
+    where
+        Self: Sized,
```
richardpringle, line 138: This is a nit (you can ignore it), but I really don't like looking at bounds in two different places:
```suggestion
    fn binprot_read<R>(r: &mut R) -> Result<Self, binprot::Error>
    where
        R: std::io::Read + ?Sized,
        Self: Sized,
```

richardpringle: ... but as I dive deeper, I'm guessing your IDE did this based on the trait definition

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -48,14 +49,70 @@
...
+    {
+        let bytes: Vec<u8> = BinProtRead::binprot_read(r)?;
+        let s = String::from_utf8(bytes).map_err(|e| binprot::Error::CustomError(Box::new(e)))?;
```
richardpringle: They have an error variant for this
```suggestion
        let s = String::from_utf8(bytes).map_err(|e| e.utf8_error().into())?;
```

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -116,27 +175,35 @@ pub enum SignalingMethod {
...
     /// * `false` for P2P relay methods
     ///
     /// This is useful for connection strategy decisions and determining whether
     /// bootstrap connections are needed before signaling can occur.
     pub fn can_connect_directly(&self) -> bool {
-        match self {
-            Self::Http(_) | Self::Https(_) | Self::HttpsProxy(_, _) => true,
-            Self::P2p { .. } => false,
-        }
+        !matches!(self, Self::P2p { .. })
```
richardpringle, line 242: 🥇

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -162,21 +230,38 @@ impl SignalingMethod {
...
+                        "{}{}{}",
+                        if needs_start_slash { "/" } else { "" },
+                        prefix,
+                        if needs_end_slash { "/" } else { "" }
+                    ))
+                };
                 return Some(format!(
-                    "https://{}:{}/clusters/{}/mina/webrtc/signal",
-                    info.host, info.port, cluster_id
+                    "{scheme}://{host}:{port}{prefix_cow}mina/webrtc/signal",
```
richardpringle, line 294: if `prefix_cow` is just `/`, isn't that not a valid proxy? I don't actually know what I'm talking about here, but it seems like this would parse into a non-proxied enum variant, no?

I feel like there's a way to do this that less complex, but I could be wrong.

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -352,6 +442,38 @@ impl FromStr for SignalingMethod {
                     .or(Err(SignalingMethodParseError::InvalidClusterId))?;
                 Ok(Self::HttpsProxy(cluster_id, rest.parse()?))
             }
+            "proxied" => {
+                // Format: /proxied/{scheme}/{encoded_prefix}/{host}/{port}
+                let mut iter = rest.splitn(4, '/').filter(|v| !v.trim().is_empty());
```
richardpringle, line 484: 👍

crates/p2p/src/webrtc/signaling_method/mod.rs
richardpringle: yeah, I don't think this should be pub based on the way you're using it. You can use encapsulation here to guarantee that (outside of this file) that the string isn't empty. You should also be able to normalize leading and trailing slashes such that they are either always or never present... I'm less sure about this though.

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -762,4 +884,394 @@ mod tests {
...
+        let method: SignalingMethod = "/proxied/https/%2Fclusters%2F123/proxy.example.com/443"
+            .parse()
+            .unwrap();
+        match method {
+            SignalingMethod::Proxied(ProxyScheme::Https, PathPrefix(prefix), info) => {
+                assert_eq!(prefix, "/clusters/123");
+                assert_eq!(info.host, Host::Domain("proxy.example.com".to_string()));
+                assert_eq!(info.port, 443);
+            }
+            _ => panic!("Expected Proxied variant"),
```
richardpringle: should print here

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -762,4 +884,394 @@ mod tests {
...
+            "/proxied/https/%2Fapi%2Fv2%2Fwebrtc/gateway.example.com/8443"
+                .parse()
+                .unwrap();
+        match method {
+            SignalingMethod::Proxied(ProxyScheme::Https, PathPrefix(prefix), info) => {
+                assert_eq!(prefix, "/api/v2/webrtc");
+                assert_eq!(info.host, Host::Domain("gateway.example.com".to_string()));
+                assert_eq!(info.port, 8443);
+            }
+            _ => panic!("Expected Proxied variant"),
```
richardpringle: should print here

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -762,4 +884,394 @@ mod tests {
...
+            HttpSignalingInfo {
+                host: Host::Domain("proxy.example.com".to_string()),
+                port: 443,
+            },
+        );
+
+        let serialized = original.to_string();
+        // Verify the serialized format contains scheme and URL-encoded path
+        assert!(serialized.contains("/proxied/https/"));
+        assert!(serialized.contains("%2F")); // URL-encoded slashes
```
richardpringle: I think we can assert on the entire string, can't we?

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -162,21 +230,38 @@ impl SignalingMethod {
     /// let url = method.http_url(); // Some("https://signal.example.com:443/mina/webrtc/signal")
     /// ```
     pub fn http_url(&self) -> Option<String> {
```
richardpringle, line 268: Can we not use the `url` crate for this? It should get rid of any redundant slashes, etc.

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -762,4 +884,394 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_proxied_missing_prefix() {
+        let result: Result<SignalingMethod, _> = "/proxied".parse();
+        assert!(result.is_err());
+        assert!(matches!(
+            result.unwrap_err(),
+            SignalingMethodParseError::NotEnoughArgs
+        ));
```
richardpringle: Should just add a PartialEq implementation on the error type

If we're worried about binary size (for wasm), can just use `#[cfg_attr(test, derive(PartialEq))]`

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -762,4 +884,394 @@ mod tests {
...
+        let method: SignalingMethod = "/proxied/http/%2Fdev%2Fproxy/localhost/3000"
+            .parse()
+            .unwrap();
+        match method {
+            SignalingMethod::Proxied(ProxyScheme::Http, PathPrefix(prefix), info) => {
+                assert_eq!(prefix, "/dev/proxy");
+                assert_eq!(info.host, Host::Domain("localhost".to_string()));
+                assert_eq!(info.port, 3000);
+            }
+            _ => panic!("Expected Proxied variant with Http scheme"),
```
richardpringle: same comment about printing

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -762,4 +884,394 @@ mod tests {
...
+
+    #[test]
+    fn test_proxied_invalid_scheme() {
+        let result: Result<SignalingMethod, _> = "/proxied/ftp/%2Fpath/example.com/21".parse();
+        assert!(result.is_err());
+        match result.unwrap_err() {
+            SignalingMethodParseError::UnknownSignalingMethod(method) => {
+                assert_eq!(method, "proxied/ftp");
+            }
+            _ => panic!("Expected UnknownSignalingMethod error"),
```
richardpringle: Same comment about printing

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -762,4 +884,394 @@ mod tests {
...
+        }
+    }
```
richardpringle, line 1246: 🥇 for all the tests in general

richardpringle (changes requested):
A lot of style comments. The only "requested changes" are really regarding the panics. Everything else looks pretty good. As for the style changes, I think there are enough that they do warrant adding but that can definitely be done in a follow up if it's easier.

crates/p2p/src/connection/outgoing/mod.rs
iostat: I tend to lean towards "if it ain't broke, and will be gone soon, don't risk breaking it further by trying to fix it"

crates/p2p/src/connection/outgoing/mod.rs
iostat: Fair and salient

crates/p2p/src/connection/outgoing/mod.rs
iostat: Yeah, didn't realize we had it in use already. Doing now

crates/p2p/src/connection/outgoing/mod.rs
iostat: I'd like to tackle this in a follow-up. We have, in `crates/p2p/src/identity/peer_id.rs`, an `impl TryFrom<mina_p2p::identity::PeerId> for libp2p_identity::PeerId`, which implies that you're able to construct a PeerId that's not a valid libp2p peer identity, and I suspect the proper fix for this has a larger surface area

iostat: It should really be a `From<mina_p2p::identity::PeerId> for libp2p_identity::PeerId`, with a massive comment in the inevitable `.expect()` of "how did you even manage to make an invalid PeerId??/1/1//1!!!". Changing `TryFrom` -> `From` /should/ force all of these expects to disappear

crates/p2p/src/connection/outgoing/mod.rs
iostat: same rationale as above

crates/p2p/src/connection/outgoing/mod.rs
iostat: applying

crates/p2p/src/connection/outgoing/mod.rs
iostat: Good one

crates/p2p/src/connection/outgoing/mod.rs
iostat: Except for Claude :P

crates/p2p/src/connection/outgoing/mod.rs
iostat: The whole "proxy" signaling thing gives me so much ick, but backwards-compat is a goal here. The original `SignalMethod::HttpsProxy(u16, ...)` was for prefixing the signal endpoint with `/clusters/{the_u16}`, and I think the only place that's used is the clustering feature in scenario tests. The same tests that don't test WebRTC afaik :).

Ironically, I don't think that server even does HTTPS. However, the equivalent in standard Multiaddr is `http-path`, and since that is separate from the `http/https` protocols in the multiaddr stack, `Proxied` is effectively a generalization that supports both HTTP and -S, and lets you set the path prefix.

crates/p2p/src/webrtc/signaling_method/mod.rs
iostat: or just assert!(matches!(result, Err(NotEnoughArgs))) ;)

crates/p2p/src/webrtc/signaling_method/mod.rs
iostat: So, `url` doesn't actually have a URL builder. You need to parse out some kind of URL to start, so I figured in for a penny, in for a pound. We had to pull in percent-encode as well since `url` doesn't reexport it in any way.

crates/p2p/src/connection/outgoing/mod.rs
iostat: Was a bit of a grind with all of these, but way better in the end

crates/p2p/src/connection/outgoing/mod.rs
iostat: There are a couple negative tests sprinkled in there, but generally speaking -- we assume:
  1. For multiaddrs, we assume invalid multiaddr will fail courtesy of the multiaddr library.
  2. For the legacy webrtc addrs -- at least there are now tests on that format. (Cheap enough to generate with AI, and useful for validating equivalence with proper multiaddr). Invalid strings *might* cause failures, but the only way an invalid WebRTC addr can make it is from misconfiguration (these don't go over the wire/p2p, you can only get them in there via CLI flags* or a rogue bootnodes.txt -- which itself is supplied by CLI flags*). 
  3. Legacy webrtc addr is a deprecated format, and the only reasons I didn't entirely remove it are because of potential surface area (possible usage in scenario tests), and so we can upgrade devnet-webrtc.txt bootnodes in o1-seeds independently of release, then remove this functionality completely in the release after.

(* CLI flags or frontend env vars -- same idea)

I assume by "panics" you mean the `expect`s earlier (non-test) in the module. since the panics here are to cause the test to fail :P. See comment above about surface area / separate issue. As far as attack vector goes, see assumption number 2.

crates/p2p/src/webrtc/signaling_method/mod.rs
iostat: That's how `derive_more::Display` does it!

crates/p2p/src/webrtc/signaling_method/mod.rs
iostat: I'm just doing pub/pub like every other structure in this module.

crates/p2p/src/webrtc/signaling_method/mod.rs
iostat: That + I vaguely remember there being a lint about trait impl definitions not matching the trait declaration. I know theres one for implementation ordering

crates/p2p/src/webrtc/signaling_method/mod.rs
iostat: `http_url()`s don't get parsed back into SignalingMethod. You are correct though, Proxied with empty or just %2F is equivalent to just `Http`/`Https`. The whole prefix_cow thing is a normalization to ensure there's */a/* slash so the final url is valid.

crates/p2p/src/connection/outgoing/mod.rs
iostat: All the "primitives" IP4, IP6, Port, and peer-id (<=32 bytes) do get owned. But `/dns*` and `/http-path` get Cow'd. Those are also the ones where the alloc could hurt!

crates/p2p/src/connection/outgoing/mod.rs
iostat: Validated that these expects won't get triggered over the wire, left a comment -- it's mostly quirks of how the p2p was originally written.

crates/p2p/src/connection/outgoing/mod.rs
richardpringle: #2074

crates/p2p/src/connection/outgoing/mod.rs
richardpringle: Ah, you're right! My bad

crates/p2p/src/connection/outgoing/mod.rs
richardpringle: Yeah, I mean `expect` as that's what causes the panic

crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -693,28 +801,25 @@ mod tests {
...
     #[test]
     fn test_whitespace_handling() {
         // The parser should filter empty components from split
         let result: Result<SignalingMethod, _> = "/http/ /8080".parse();
-        assert!(result.is_err());
         assert!(matches!(
-            result.unwrap_err(),
-            SignalingMethodParseError::NotEnoughArgs
+            result,
+            Err(SignalingMethodParseError::NotEnoughArgs)
```
richardpringle: Did I not leave a comment for all of these to be `assert_eq!`? I should have. Can address in a follow up

crates/p2p/src/webrtc/signaling_method/mod.rs
richardpringle: This shouldn't be pub. Can do in a follow up

richardpringle (dismissed):
@iostat Can you create an issue to address the two comments in this review? They can both be done in a follow up. I'm less concerned about the test comment but think we should really limit pub visibility where we can.

Limiting pub visibility makes deleting code sooooo much easier in the future

crates/p2p/src/webrtc/signaling_method/mod.rs
iostat: Missed the intent for clean error messages, just thought it was a compactness nit, fixed per slack clarification.

dannywillems (dismissed):

richardpringle (approved):

richardpringle (approved):

PR #2066, opened 2026-01-11
Cargo: share fields

dannywillems:


richardpringle (approved):
Cool, I didnt know you could do this

PR #2174, opened 2026-02-16
Updated mina node binary for interop testing framework

0xMimir:
### Description

Prepares mina node for interop testing framework

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to
      not work as expected)
- [ ] This change requires a documentation update
- [x] Build-related changes
- [ ] Performance improvement

### Changelog Entry

**Changelog Summary:**

### DIFF
--- a/crates/cli/src/commands/node/mod.rs
+++ b/crates/cli/src/commands/node/mod.rs
@@ -280,6 +280,17 @@ pub struct Node {
 
     #[arg(long, env)]
     pub rng_seed: Option<String>,
+
+    /// Override the chain ID used for P2P networking. (interop testing)
+    #[arg(long, env = "MINA_CHAIN_ID")]
+    pub chain_id: Option<String>,
+
+    /// Skip SNARK proof verification (interop testing).
+    /// WARNING: THIS DISABLES A CRITICAL SECURITY CHECK. ONLY USE
+    /// FOR INTEROP TESTING WITH OCAML NODES RUNNING --PROOF-LEVEL
+    /// NONE, WHICH PRODUCE BLOCKS WITH DUMMY PROOFS.
+    #[arg(long)]
+    pub skip_proof_verification: bool,
 }
 
 impl Node {
@@ -352,6 +363,14 @@ impl Node {
             ),
         };
 
+        // Parse chain ID override (applied to builder below).
+        let chain_id_override = self
+            .chain_id
+            .as_deref()
+            .map(mina_node::core::ChainId::from_hex)
+            .transpose()
+            .context("invalid --chain-id hex string")?;
+
         let custom_rng_seed = match self.rng_seed {
             None => None,
             Some(v) => match hex::decode(v)
@@ -373,6 +392,13 @@ impl Node {
 
         let mut node_builder: NodeBuilder =
             NodeBuilder::new(custom_rng_seed, daemon_conf, genesis_conf);
+
+        if let Some(chain_id) = chain_id_override {
+            node_builder.chain_id_override(chain_id);
+        }
+        if self.skip_proof_verification {
+            node_builder.skip_proof_verification(true);
+        }
 
         // Configure P2P identity
         if let Some(sec_key) = self.p2p_secret_key {
--- a/crates/core/src/constants.rs
+++ b/crates/core/src/constants.rs
@@ -1,3 +1,5 @@
+use std::sync::OnceLock;
+
 use binprot_derive::BinProtWrite;
 use mina_curves::pasta::Fp;
 use mina_p2p_messages::{bigint, number, v2};
@@ -11,16 +13,28 @@ pub const PROTOCOL_VERSION: v2::ProtocolVersionStableV2 = v2::ProtocolVersionSta
 };
 
 pub fn constraint_constants() -> &'static ConstraintConstants {
-    NetworkConfig::global().constraint_constants
+    CONSTRAINT_CONSTANTS_OVERRIDE
+        .get()
+        .unwrap_or(NetworkConfig::global().constraint_constants)
+}
+
+static CONSTRAINT_CONSTANTS_OVERRIDE: OnceLock<ConstraintConstants> = OnceLock::new();
+
+/// Override the fork constants used during genesis computation. (interop testing)
+pub fn set_fork_override(fork: ForkConstants) {
+    let mut cc = NetworkConfig::global().constraint_constants.clone();
+    cc.fork = Some(fork);
+    let _ = CONSTRAINT_CONSTANTS_OVERRIDE.set(cc);
 }
 
 /// Constants that define fork-specific blockchain state.
 ///
 /// Fork constants specify the blockchain state at which a protocol upgrade or fork occurred.
 /// These are used to handle protocol changes and ensure compatibility across network upgrades.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
 pub struct ForkConstants {
     /// Hash of the blockchain state at the fork point
+    #[serde(with = "fp_state_hash_serde")]
     pub state_hash: Fp,
 
     /// Blockchain length (number of blocks) at the fork point
@@ -28,6 +42,25 @@ pub struct ForkConstants {
 
     /// Global slot number since genesis at the fork point
     pub global_slot_since_genesis: u32,
+}
+
+mod fp_state_hash_serde {
+    use super::*;
+
+    pub fn serialize<S: serde::Serializer>(fp: &Fp, serializer: S) -> Result<S::Ok, S::Error> {
+        let state_hash = v2::StateHash::from_fp(*fp);
+        serializer.serialize_str(&state_hash.to_string())
+    }
+
+    pub fn deserialize<'de, D: serde::Deserializer<'de>>(deserializer: D) -> Result<Fp, D::Error> {
+        let s: String = serde::Deserialize::deserialize(deserializer)?;
+        let state_hash: v2::StateHash = s.parse().map_err(serde::de::Error::custom)?;
+        state_hash
+            .inner()
+            .0
+            .to_field()
+            .map_err(|e| serde::de::Error::custom(format!("invalid state hash: {e}")))
+    }
 }
 
 /// Protocol constraint constants that define core blockchain behavior.
--- a/crates/core/src/encrypted_key.rs
+++ b/crates/core/src/encrypted_key.rs
@@ -283,11 +283,21 @@ pub trait EncryptedSecretKey {
     /// - Default Argon2i parameters: 128MB memory cost, 6 iterations
     /// - Each encryption produces unique salt and nonce for security
     fn try_encrypt(key: &[u8], password: &str) -> Result<EncryptedSecretKeyFile, EncryptionError> {
-        let argon2 = setup_argon(Self::PW_DIFF)?;
-
         // add the prefix byte to the key
         let mut key_prefixed = vec![Self::SECRET_KEY_PREFIX_BYTE];
         key_prefixed.extend(key);
+
+        Self::try_encrypt_raw(&key_prefixed, password)
+    }
+
+    /// Encrypt raw bytes without prepending a version/prefix byte.
+    /// Used for account/BP keys, use [`EncryptedSecretKey::try_encrypt`] which adds the
+    /// standard 0x01 prefix byte.
+    fn try_encrypt_raw(
+        plaintext: &[u8],
+        password: &str,
+    ) -> Result<EncryptedSecretKeyFile, EncryptionError> {
+        let argon2 = setup_argon(Self::PW_DIFF)?;
 
         let salt = SaltString::generate(&mut OsRng);
         let password_hash = argon2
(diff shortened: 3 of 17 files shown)


Cargo.toml
```diff
@@ -157,7 +157,7 @@ mina-curves = { git = "https://github.com/o1-labs/proof-systems", tag = "0.3.0"
 mina-fuzzer = { path = "crates/fuzzer" }
 mina-hasher = { git = "https://github.com/o1-labs/proof-systems", tag = "0.3.0" }
 mina-macros = { path = "crates/macros" }
-mina-node = { path = "crates/node" }
+mina-node = { path = "crates/node", default-features = false }
```
richardpringle: Why do you have to turn off the default features, are the features additive in dependent crates? Can't you just add this flag in a dependent?

crates/cli/src/commands/node/mod.rs
```diff
@@ -280,6 +280,17 @@ pub struct Node {
...
+    /// Override the chain ID used for P2P networking. (interop testing)
+    #[arg(long, env = "MINA_CHAIN_ID")]
+    pub chain_id: Option<String>,
+
+    /// Skip SNARK proof verification (interop testing).
+    /// WARNING: THIS DISABLES A CRITICAL SECURITY CHECK. ONLY USE
+    /// FOR INTEROP TESTING WITH OCAML NODES RUNNING --PROOF-LEVEL
+    /// NONE, WHICH PRODUCE BLOCKS WITH DUMMY PROOFS.
+    #[arg(long)]
+    pub no_verify_proofs: bool,
```
richardpringle: Can you explain why we have to do this? Why can't we verify proofs in testing?

crates/cli/src/commands/node/mod.rs
```diff
@@ -374,6 +393,13 @@ impl Node {
         let mut node_builder: NodeBuilder =
             NodeBuilder::new(custom_rng_seed, daemon_conf, genesis_conf);
 
+        if let Some(chain_id) = chain_id_override {
+            node_builder.chain_id_override(chain_id);
```
richardpringle, line 397: Why is this an "override" as opposed to just setting the chain-id? Usually, you would get the chain-id from the environment and call unwrap_or_default() on the option.

crates/cli/src/commands/node/mod.rs
```diff
@@ -374,6 +393,13 @@ impl Node {
...
+        }
+        if self.no_verify_proofs {
```
richardpringle: I think `skip_proof_verification` is a better name. In any case, it's better to keep the naming consistent.

crates/core/src/constants.rs
```diff
@@ -1,3 +1,5 @@
+use std::sync::OnceLock;
+
```
richardpringle, line 2:
```suggestion
```

crates/core/src/constants.rs
```diff
@@ -11,7 +13,18 @@ pub const PROTOCOL_VERSION: v2::ProtocolVersionStableV2 = v2::ProtocolVersionSta
...
+        .unwrap_or(NetworkConfig::global().constraint_constants)
+}
+
+static CONSTRAINT_CONSTANTS_OVERRIDE: OnceLock<ConstraintConstants> = OnceLock::new();
+
+/// Override the fork constants used during genesis computation. (interop testing)
+pub fn set_fork_override(fork: ForkConstants) {
+    let mut cc = NetworkConfig::global().constraint_constants.clone();
+    cc.fork = Some(fork);
+    let _ = CONSTRAINT_CONSTANTS_OVERRIDE.set(cc);
```
richardpringle, line 27: 🤔 

While this code looks fine, what happens if we need several overrides? How do we choose the order? 

I know that part of this stems from the network-config being a global, but can't we just set all the values to what we want when the node is starting? 

Otherwise, it might be a better idea to refactor the code such that we don't have the network config as a global before applying those changes. 

I would argue that globals are an anti-pattern in all languages, but in Rust in particular they are very cumbersome to change as we can see with this code.

What do you think, @0xMimir?

crates/core/src/constants.rs
```diff
@@ -30,6 +43,41 @@ pub struct ForkConstants {
...
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        use serde::ser::SerializeStruct;
+        let state_hash = v2::StateHash::from_fp(self.state_hash);
+        let mut s = serializer.serialize_struct("ForkConstants", 3)?;
+        s.serialize_field("state_hash", &state_hash.to_string())?;
+        s.serialize_field("blockchain_length", &self.blockchain_length)?;
+        s.serialize_field("global_slot_since_genesis", &self.global_slot_since_genesis)?;
+        s.end()
+    }
+}
```
richardpringle: Checkout the serde.rs docs page to see how to write a custom function for a single field. Say we wanted to add a field to `ForkConstants`. With this custom Serialize implementation, it would be really easy to forget to serialize the new field here.

crates/core/src/constants.rs
```diff
@@ -30,6 +43,41 @@ pub struct ForkConstants {
...
+            blockchain_length: u32,
+            global_slot_since_genesis: u32,
+        }
+        let raw = Raw::deserialize(deserializer)?;
+        let state_hash: v2::StateHash = raw.state_hash.parse().map_err(serde::de::Error::custom)?;
+        let fp: Fp = state_hash
+            .inner()
+            .0
+            .to_field()
+            .map_err(|e| serde::de::Error::custom(format!("invalid state hash: {e}")))?;
```
richardpringle: You can use `expect` here. I've got a PR coming that converts `to_field`'s return type as it actually isn't fallible

richardpringle: But you'll want this in the custom deserialize function (that goes with my comment above)

crates/core/src/encrypted_key.rs
```diff
@@ -283,12 +283,22 @@ pub trait EncryptedSecretKey {
     /// - Default Argon2i parameters: 128MB memory cost, 6 iterations
     /// - Each encryption produces unique salt and nonce for security
     fn try_encrypt(key: &[u8], password: &str) -> Result<EncryptedSecretKeyFile, EncryptionError> {
-        let argon2 = setup_argon(Self::PW_DIFF)?;
-
         // add the prefix byte to the key
         let mut key_prefixed = vec![Self::SECRET_KEY_PREFIX_BYTE];
         key_prefixed.extend(key);
 
+        Self::try_encrypt_raw(&key_prefixed, password)
```
richardpringle, line 290: Why did you make this a new function when it only has a single call-site? Are you planning on using this elsewhere?

crates/node/common/src/service/snarks.rs
```diff
@@ -50,16 +51,17 @@ impl NodeService {
...
-                        if !ledger::proofs::verification::verify_block(
-                            header,
-                            &verifier_index,
-                            &verifier_srs,
-                        ) {
-                            Err(SnarkBlockVerifyError::VerificationFailed)
-                        } else {
-                            Ok(())
-                        }
+                    let result = if skip_proof_verification {
```
richardpringle: Can you refactor this, please? 

```rust
let verified = skip_proof_verification || verify_block(...);

let result = if verified { Ok(()) } else { Err(...) };
```

crates/node/common/src/service/snarks.rs
```diff
@@ -50,16 +51,17 @@ impl NodeService {
...
+                        eprintln!("verify({}) - skipping proof verification", block.hash_ref());
```
richardpringle, line 56: This should be a warning. Maybe we can add a proper logger to this crate before we finish this feature

crates/node/src/daemon_json/mod.rs
```diff
@@ -42,6 +44,11 @@ pub struct EpochData {
     pub seed: String,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Proof {
+    pub fork: Option<ForkConstants>,
+}
```
richardpringle, line 50: 🤔 

What's the purpose of this type. We should have some documentation, but maybe you can explain here first

crates/node/src/event_source/event_source_effects.rs
```diff
@@ -464,9 +464,34 @@ pub fn event_source_effects<S: Service>(store: &mut Store<S>, action: EventSourc
                     }
                 },
                 BlockProducerEvent::BlockProve(block_hash, res) => match res {
-                    Err(err) => todo!(
-                        "error while trying to produce block proof for block {block_hash} - {err}"
-                    ),
+                    Err(err) => {
+                        if store.state().config.skip_proof_verification {
```
richardpringle: Please move this to a guard on another match branch

crates/node/src/transition_frontier/genesis/transition_frontier_genesis_reducer.rs
```diff
@@ -104,16 +105,21 @@ impl TransitionFrontierGenesisState {
...
-                        constraint_system_digests,
-                        genesis_hash,
-                        &genesis.body.constants,
-                        constants::PROTOCOL_TRANSACTION_VERSION,
-                        constants::PROTOCOL_NETWORK_VERSION,
-                        &v2::UnsignedExtendedUInt32StableV1::from(constants::TX_POOL_MAX_SIZE),
-                    );
+                    let chain_id = if let Some(override_id) = &global_state.config.chain_id_override
+                    {
+                        override_id.clone()
```
richardpringle, line 110: Don't have to do this now, but maybe add a TODO? It would be better if we just made this a static with `Box::leak` as (if I'm not mistaken) it should live for the lifetime of the program, right?

crates/node/src/transition_frontier/genesis/transition_frontier_genesis_reducer.rs
```diff
@@ -130,6 +136,19 @@ impl TransitionFrontierGenesisState {
...
+                // OCaml --proof-level none), skip real genesis proving and
+                // use a dummy proof. The real prover can panic on custom
+                // genesis configurations due to sparse ledger mismatches.
+                if skip_proof_verification {
+                    let dispatcher = state_context.into_dispatcher();
+                    dispatcher.push(TransitionFrontierGenesisAction::ProvePending);
+                    dispatcher.push(TransitionFrontierGenesisAction::ProveSuccess {
+                        proof: dummy_blockchain_proof().clone(),
+                    });
+                    return;
```
richardpringle, line 149: Ah the early return, have I told you how much I love the early return? 

🥇

richardpringle (changes requested):
So most of what I see that needs changing is probably more because of the existing code than your new code. 

Can you reply to my comments and make the requested changes to the (de)serialization? 

After, we can discuss whether or not it's worth making the broader rafactor before landing this or not. My guess is that it will make more sense to land this first.

crates/cli/src/commands/node/mod.rs
0xMimir: It's needed so the Rust CLI binary can interop with OCaml nodes running --proof-level none. Those OCaml nodes produce blocks with dummy SNARK proofs that fail verification.

crates/cli/src/commands/node/mod.rs
0xMimir: Mainly to distinguish between computed one and one that will be set by testing framework

crates/core/src/constants.rs
0xMimir: Why and when would we need several  overrides?

crates/core/src/encrypted_key.rs
0xMimir: yes

crates/node/src/daemon_json/mod.rs
0xMimir: Proof maps the "proof" key in daemon.json, it only carries the optional fork field

crates/core/src/constants.rs
0xMimir: Does this mean to not use `OnceLock`

richardpringle: line 2 (empty space) not line 1

crates/core/src/constants.rs
richardpringle: I can't predict the future! 😅 haha. Usually we assume code will extended or changed

crates/node/common/src/service/snarks.rs
richardpringle: @0xMimir, do you know if we have an issue for logging? If not can we make one? I can't remember the crate name, but we should use something that's standardized with log-levels, etc

crates/core/src/constants.rs
richardpringle: Actually, ignore this, it's fine

crates/node/common/src/service/snarks.rs
0xMimir: Issue created #2177

richardpringle (approved):

PR #2175, opened 2026-02-17
ci: add workflow to suggest CHANGELOG entries via Claude

### LINKED ISSUE #2118: Dry run Claude suggesting changelog entries when they are missing
When a changelog entry is missing: Feed PR through Claude, see what it decides the affected surface area is, and leave a comment suggesting entries to add to the changelog.

iostat:
### Description

When a PR targets develop without modifying CHANGELOG.md and without the "no changelog" label, this workflow uses claude-code-action to generate suggested entries from commit messages and PR description, then posts them as a PR comment.

Previous suggestions are archived as base64-encoded HTML comments to preserve history across re-runs.

### Related Issue(s)

* Closes #2118 

### Type of Change

- [x] Build-related changes

### Testing

1. Open a test PR to `develop` without modifying CHANGELOG.md and without the `no changelog` label — the workflow should trigger and Claude should post a comment with suggested entries.
2. Push a new commit to the same PR — the previous comment should be deleted and a new one posted with hidden history
3. Add the `no changelog` label — the workflow should not run
4. Modify `CHANGELOG.md` in a PR — the workflow should skip the suggestion step

### Changelog Entry

Oh no, I forgot the entry!!11 >:)

### DIFF
--- a/.github/workflows/changelog-suggest.yaml
+++ b/.github/workflows/changelog-suggest.yaml
@@ -1,0 +1,195 @@
+name: Suggest CHANGELOG entries
+
+on:
+  pull_request:
+    types: [opened, synchronize, unlabeled]
+    branches:
+      - develop
+
+concurrency:
+  group: changelog-suggest-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  suggest-changelog:
+    name: Suggest changelog entries
+    if: >-
+      !contains(github.event.pull_request.labels.*.name, 'no changelog')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      issues: write
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Check changelog and manage previous suggestions
+        id: check
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { owner, repo } = context.repo;
+            const prNumber = context.issue.number;
+
+            // Check if CHANGELOG.md was modified in this PR
+            const { data: files } = await github.rest.pulls.listFiles({
+              owner,
+              repo,
+              pull_number: prNumber,
+              per_page: 100,
+            });
+
+            const changelogChanged = files.some(
+              (f) => f.filename === "CHANGELOG.md"
+            );
+            core.setOutput("changelog_changed", changelogChanged.toString());
+
+            if (changelogChanged) {
+              core.info("CHANGELOG.md was modified, skipping suggestion");
+              return;
+            }
+
+            // Look for previous suggestion comment
+            const marker = "<!-- changelog-suggestion -->";
+            let previousBody = "";
+            let page = 1;
+
+            while (true) {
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog-suggest.yaml
```diff
@@ -0,0 +1,195 @@
...
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Check changelog and manage previous suggestions
+        id: check
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { owner, repo } = context.repo;
```
dannywillems, line 32: I'm fine with refining over time, but I would prefer to have all of the scripts in actual files, and run typechecker, linter, formatter, etc on it. And we can iterate on the files.

.github/workflows/changelog-suggest.yaml
```diff
@@ -0,0 +1,195 @@
...
+              core.setOutput("had_previous", "false");
+            }
+
+      - name: Generate changelog suggestion
+        if: steps.check.outputs.changelog_changed == 'false'
+        uses: anthropics/claude-code-action@v1
+        with:
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          prompt: |
+            You are helping maintain a CHANGELOG for the mina-rust project.
```
dannywillems, line 104: Same for the prompt, if it is possible to have it in a txt file. It is more portable if we want to make some local tests for instance.

.github/workflows/changelog-suggest.yaml
```diff
@@ -0,0 +1,195 @@
...
+
+            Do not post a comment. Do not modify any repo files.
+            Write only to /tmp/changelog_suggestion.md.
+
+      - name: Post changelog suggestion
+        if: steps.check.outputs.changelog_changed == 'false'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const fs = require("fs");
```
dannywillems, line 153: See https://github.com/o1-labs/mina-rust/pull/2175/changes#r2818812824

.github/workflows/changelog-suggest.yaml
iostat: `actions/github-script` doesn't take a file arg -- I *could* try to hack in loading it into an env, but honestly, the way this action is meant to be used -- that logic isn't reusable or testable outside the script action anyway...

.github/workflows/changelog-suggest.yaml
iostat: Not possible, and actually, there is great wisdom in Anthropic doing it like this along with the "has to match main branch". Otherwise, anyone can modify the prompt, open a PR, and have Claude exfiltrate keys/tokens/whatever else the github runner has access to.

.github/workflows/changelog-suggest.yaml
iostat: Same response as above

dannywillems (approved):

PR #2200, opened 2026-03-26
fix(p2p): gracefully handle unknown protocols in Identify

iostat:
(stacked PRs: this is 1 of 3, merge this before #2201 and #2202)

### Description

Replace the strict `parse_protocol()` that rejected entire Identify messages
on unknown protocol strings with a `StreamProtocolId` enum (`Known`/`Unknown`)
that preserves unrecognized protocols for forward compatibility. This prevents
older Rust nodes from disconnecting peers that advertise new capability
strings, consistent with how go-libp2p handles unknown protocols.

### Related Issue(s)

Closes #2189

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

### Testing

- Existing test suite passes; `parse_protocol` no longer returns `Result`,
  so unknown strings are silently preserved rather than causing errors.

### Changelog Entry

**Fixed**: **P2P**: Identify protocol no longer rejects peers that advertise unknown
protocol strings, improving forward compatibility with newer nodes ([#2200](https://github.com/o1-labs/mina-rust/pull/2200))

PR #2201, opened 2026-03-26
p2p: remove deprecated legacy WebRTC address format

### LINKED ISSUE #2074: Fix fragile legacy webrtc-multiaddrish parsing code in p2p crate
This is... fragile. Think it's worth improving while we're editing this code? These `FromStr` implementations (at least this and `webrtc::SignalingMethod` implementations) rely on the leading `/` instead of using the `/` as a delimiter and parsing the actual stuff between the `/`. 

Personally, I would be in favour of changing this now so that it doesn't come to bite us at some other point. I realize that would scope creep on this PR though... so, you're call.

_Originally posted by @richardpringle in https://github.com/o1-labs/mina-rust/pull/2052#discussion_r2676855494_

@iostat EDIT: disposition is to remove this parsing with 0.20, along with the `Identify` fix since that's a "breaking" change to the P2P as well to enable webrtc-relay advertisement.

iostat:
(stacked PRs: this is 2 of 3, merge #2200 first and then #2202 after)

### Description

Remove the deprecated legacy WebRTC address format
(`/{peer_id}/{signaling_method}/...`) and simplify `FromStr` for
`P2pConnectionOutgoingInitOpts` to only accept standard multiaddr. The legacy
format was deprecated with warnings in place; this completes the removal.

Also cleans up the error enum: replaces `NotEnoughArgs`, `PeerIdParseError`,
`SignalingMethodParseError`, and the catch-all `Other` with three specific
variants: `InvalidMultiaddr`, `InvalidMultiaddrStructure`, `InvalidPeerId`.

### Related Issue(s)

Closes #2074

### Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality
      to not work as expected)
- [x] This change requires a documentation update

### Testing

- Legacy format test removed alongside the code path.
- All remaining multiaddr parsing tests pass.

### Changelog Entry
**Removed:** **P2P**: Deprecated legacy WebRTC address format (`/{peer_id}/{signaling}`); only standard multiaddr is now accepted ([#2201](https://github.com/o1-labs/mina-rust/pull/2201))


crates/p2p/src/webrtc/signaling_method/mod.rs
```diff
@@ -1078,113 +422,21 @@ mod tests {
...
-        assert_eq!(
-            m2.http_url().unwrap(),
-            "https://example.com:443/path/mina/webrtc/signal"
-        );
+        let cases = vec![
+            ("path", "https://example.com:443/path/mina/webrtc/signal"),
+            ("/path", "https://example.com:443/path/mina/webrtc/signal"),
+            ("path/", "https://example.com:443/path/mina/webrtc/signal"),
+            ("/path/", "https://example.com:443/path/mina/webrtc/signal"),
+        ];
```
richardpringle, line 429: Should use https://docs.rs/test-case/latest/test_case/

richardpringle (approved):

